PR #1, opened 2021-09-15
Add Missing AppVersions

fredericfran-gds:
Add missing appVersions to Publisher and Router.


nsabri1 (approved):
LGTM

sengi:
Why do we want appVersion?

PR #3, opened 2021-09-15
Add GitHub Actions linting

fredericfran-gds:
In addition, we bumped the versions of router and publisher because 
the helm chart releaser is failing since the chart was updated in #1 but 
the chart version was not updated. I.e the chart releaser cannot overwrite
an existing chart.

We decided to remove appVersion from the Chart.yaml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: linting
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1.1
+        with:
+          version: 3.*
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.*
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.1.0
+
+      - name: Run chart-testing (list-changed)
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1.1
+        with:
+          version: 3.*
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
sengi, line 19: I'd recommend using `3.x` here to stay up to date unless there's a known particular compatibility issue. 3.7 is only due another couple of years of security patches.

nsabri1 (approved):
LGTM

sengi (approved):
I don't understand the PR description but the code looks fine 👍 (what do you mean by "we bumped the versions of router and publisher as it has a previous versions already tarred" - not sure I can even parse that 😅)

.github/workflows/ci.yml
fredericfran-gds: ok

fredericfran-gds:
> I don't understand the PR description but the code looks fine 👍 (what do you mean by "we bumped the versions of router and publisher as it has a previous versions already tarred" - not sure I can even parse that 😅)

I improved the description, sorry for the lack of clarity.

In addition, we bumped the versions of router and publisher because
the helm chart releaser is failing since the chart was updated in #1 but
the chart version was not updated. I.e the chart releaser cannot overwrite
an existing chart.

PR #4, opened 2021-09-15
Remove redundant ssl-policy annotations from Ingresses.

sengi:
The TLS >= 1.2 policy is now configured as the default on the controller as of alphagov/govuk-infrastructure#451.

Tested: installed in namespace on the test cluster and ran an [audit of the TLS config](https://www.ssllabs.com/ssltest/analyze.html?d=www-origin-chris.eks.test.govuk.digital&s=34.240.19.213&latest) to verify that TLS 1.0 and 1.1 are still disabled.

Needs alphagov/govuk-infrastructure#451 to be merged first.

[Trello card](https://trello.com/c/kL8xAztY/637)

### DIFF
--- a/charts/publisher/Chart.yaml
+++ b/charts/publisher/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: publisher
 description: A Helm chart for GOV.UK Publisher
 type: application
-version: 0.2.1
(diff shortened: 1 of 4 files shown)


nsabri1 (approved):
LGTM

PR #5, opened 2021-09-21
Add integration tests

karlbaker02:
This PR primarily adds a new Helm charts for Smokey; as part of this, a new Helm chart for Signon has also been created, in order to get further through the integration tests run by Smokey.

Additionally, this PR also:

- fixes the naming convention we're using for application endpoints where we have those endpoints already defined within Router and Publisher
- updates Router to have a backend endpoint for Signon
- fixes the readiness healthcheck path for Publisher
- bumps the Helm chart versions for Router and Publisher

Trello: https://trello.com/c/ZBqP6fqp


charts/signon/templates/deployment.yaml
```diff
@@ -0,0 +1,117 @@
...
+              value: "{{ .Values.appAssetHost | default (printf "https://www-origin-%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+            - name: DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: signon
+                  key: database-url
+            - name: DEVISE_PEPPER
+              value: "secret"
+            - name: DEVISE_SECRET_KEY
+              value: "secret"
```
sengi, line 34: Please add TODOs for things like this. (like SECRET_KEY_BASE below)

charts/signon/templates/secrets.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: external-secrets.io/v1alpha1
+kind: ExternalSecret
+metadata:
+  name: signon
+spec:
+  refreshInterval: 2m
```
sengi, line 7: I think we should consider either omitting refreshInterval for now (default is 1h) or at least setting it to something a bit more conservative like say 30m. 2m sounds riskily frequent given than we're likely to have hundreds of secrets. (The risk isn't so much hammering Secrets Manager - which supports pretty high qps - but hammering our own etcd, which is highly undesirable.)

charts/signon/templates/deployment.yaml
```diff
@@ -0,0 +1,117 @@
...
+            - name: GOVUK_ASSET_ROOT
+              value: "{{ .Values.appGovukAssetRoot | default (printf "https://assets-%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+            - name: GOVUK_CONTENT_SCHEMAS_PATH
+              value: "{{ .Values.appGovukContentSchemasPath }}"
+            - name: GOVUK_ENVIRONMENT
+              value: "{{ .Values.govukEnvironment }}"
+            - name: PLEK_SERVICE_CONTENT_STORE_URI
+              value: "{{ .Values.appPlekServiceContentStoreUri | default (printf "http://content-store.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+            - name: PLEK_SERVICE_PUBLISHING_API_URI
+              value: "{{ .Values.appPlekServicePublishingApiUri | default (printf "http://publishing-api-web.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
```
sengi, line 56: Does Signon actually use this? I couldn't find any reference to /publishing.api/ in the tree.

charts/smokey/templates/cronjobs/all-smoke-tests.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: all-smoke-tests
+spec:
+  schedule: "*/10 * * * *"
```
sengi, line 6: Do the tests actually finish within 10 minutes? :) Might be worth giving it a bit more leeway.

charts/smokey/templates/jobs/frontend-smoke-tests-job.yaml
```diff
@@ -0,0 +1,24 @@
...
+        command: ["bundle", "exec", "cucumber", "--profile", "test", "--strict-undefined", "-t", "@app-frontend", "-t", "@replatforming", "-t", "not @notreplatforming and not @nottest"]
+        env:
+        - name: ENVIRONMENT
+          value: "{{ .Values.govukEnvironment }}"
+        - name: GOVUK_APP_DOMAIN
+          value: "{{ .Values.appGovukAppDomain | default (printf "%s.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.clusterDomain) }}"
+        - name: GOVUK_WEBSITE_ROOT
+          value: "{{ .Values.appGovukWebsiteRoot | default (printf "https://www-origin-%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+        # - name: PLEK_SERVICE_FRONTEND_URI
+        #   value: "{{ .Values.appPlekServiceFrontendUri | default (printf "http://frontend.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
```
sengi, line 20: Please either remove the commented-out code or add a TODO saying what needs to happen for it to be uncommented.

charts/smokey/values.yaml
```diff
@@ -0,0 +1,16 @@
+govukEnvironment: test
+govukDomainExternal: govuk.digital
+govukDomainInternal: govuk-internal.digital
+clusterDomain: svc.cluster.local
+
+appImage:
+  repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/smokey
+  pullPolicy: Always
```
sengi, line 8: nit: `pullPolicy: Always` is redundant when using `tag: latest`.

sengi (approved):
Cool. Thanks for fixing up the app URLs along the way!

Took me a minute to figure out that you meant "endpoints" in the generic sense 😅 Endpoints are also a type of Kubernetes object that represents (basically) the network address of a Pod that belongs to a Service.

charts/signon/templates/secrets.yaml
karlbaker02: Sounds reasonable, will change

charts/smokey/templates/cronjobs/all-smoke-tests.yaml
karlbaker02: Caveated that I couldn't actually get "all smoke tests" running because of missing apps. Checking what we currently have running via Jenkins, it seems to take less than three minutes for smoke tests to complete, so 10 minutes might be alright. Happy to give it some more time though!

charts/smokey/templates/jobs/frontend-smoke-tests-job.yaml
karlbaker02: Good spot

charts/signon/templates/deployment.yaml
karlbaker02: I will double-check, you may be right here

PR #7, opened 2021-09-28
Modify publisher chart values structure, normalise resource labels, and misc. tweaks

kerin:
This PR is currently a draft, as there are a few areas which warrant some further discussion across the team. Changes are:

### 1. Changes Helm values structure

While Helm's [best practices docs](https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values) recommend a preference for a flat values structure, they do also state that this is a preference and that "nested values may be used to improve readability". This is a common pattern when a single Helm chart contains multiple distinct components, in this case `web` and `worker` (see [`kube-prometheus-stack`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml) for a larger example).

`values.yaml` has been restructured to group all values under one of `common`, `web` or `worker`. In addition, `common` contains a docker `image` block that is copied into `web` and `worker` values via [YAML anchors](https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors), allowing the docker image properties to be specified either once for each component, or individually if required (ie. if the worker requires additional libraries beyond those present in the web container).

### 2. Normalised resource labels and adopted Kubernetes recommended labels

Helm generates charts with some ["top level" Kubernetes-recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) by default. These are now applied to all resources, and the `app.kubernetes.io/component` label has been added to resources, with a value of either `web` or `worker` to indicate the logical component.

This label structure may not be exactly what we want, and so likely requires some discussion across the team, particularly around how the `app.kubernetes.io/name`, `/component` and `/part-of` labels can be best used - for reference, other charts [use `/part-of` as the highest-order label rather than `/name`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/_helpers.tpl#L52).

Similarly we may want to consider:

- are there other labels we should introduce as standard? Team name, organisational area, etc?
- what labels should be used for `selector.matchLabels` in `Deployment` resources? k8s recommended labels could be used here, or a simple `app` label or similar could be used instead.

### 3. `prefix` value for the `ExternalSecret` object replaced with a full `externalSecretRef` value

The use of a per-install `prefix` value in conjunction with the Chart's release name as the AWS secret name would require that every instance of the Helm chart would need a distinct AWS Secret. This is likely not the case, as multiple non-prod instances of the Helm chart may use a single set of secrets (`dev` or similar) - `externalSecretRef` instead takes the full AWS Secret name, so any instance can use any set of secrets.

### 4. misc. tweaks

- `web.ingress.pathType` value added (with a `Prefix` default) to allow more control over load balancer path matching
- removed `extraEnv` values in favour of managing all env vars via the env `ConfigMap`
- renamed `ExternalSecret` and nginx.conf `ConfigMap` YAML files to be more descriptive
- added `externalSecretRefreshInterval` with a `5m` default
- `govukEnvironment` default changed to `test`, as `production` is possibly not a safe default
- `railsEnv` env var defaults to the same value as `govukEnvironment` unless otherwise specified


charts/publisher/templates/web/dbmigration-job.yaml
```diff
@@ -1,21 +1,23 @@
...
-        app.kubernetes.io/instance: {{ .Release.Name | quote }}
-        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+        {{- include "publisher.labels" . | nindent 8 }}
+        app.kubernetes.io/component: "web"
     spec:
       restartPolicy: Never
       containers:
       - name: {{ .Release.Name }}-dbmigrate
-        image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+        image: "{{ .Values.web.image.repository }}:{{ .Values.web.image.tag }}"
```
sengi, line 20: I'd recommend not allowing the image to differ between the "web" and "worker" deployments. By design they should always be running the same binary so it'd be best to model them that way in the chart, rather than potentially implying that the user could override them to be different from each other by passing in different values.

(In other words, if someone were to make use of these separate values by setting them to point at different images, that would always be a mistake - so it would be preferable to prevent the mistake by referring directly to common.image etc. in the Deployment specs. The idea being that it's preferable to make erroneous configurations impossible by design.)

charts/publisher/templates/common/configmap.yaml
```diff
@@ -2,29 +2,31 @@ apiVersion: v1
...
+  PLEK_SERVICE_CONTENT_STORE_URI: "{{ .Values.common.env.plekServiceContentStoreUri | default (printf "http://content-store.%s.%s" .Release.Namespace .Values.common.govukDomainInternal) }}"
+  PLEK_SERVICE_DRAFT_ORIGIN_URI: "{{ .Values.common.env.plekServiceContentStoreUri | default (printf "https://draft-origin-%s.eks.%s.%s" .Release.Namespace .Values.common.govukEnvironment .Values.common.govukDomainExternal) }}"
+  PLEK_SERVICE_PUBLISHING_API_URI: "{{ .Values.common.env.plekServicePublishingApiUri | default (printf "http://publishing-api-web.%s.%s" .Release.Namespace .Values.common.govukDomainInternal) }}"
+  PLEK_SERVICE_SIGNON_URI: "{{ .Values.common.env.plekServiceSignonUri | default (printf "https://signon-%s.eks.%s.%s" .Release.Namespace .Values.common.govukEnvironment .Values.common.env.govukDomainExternal) }}"
+  PLEK_SERVICE_STATIC_URI: "{{ .Values.common.env.plekServiceStaticUri | default (printf "http://static.%s.%s" .Release.Namespace .Values.common.govukDomainInternal) }}"
+  PORT: "{{ .Values.web.port }}"
+  RAILS_ENV: "{{ .Values.common.env.railsEnv | default .Values.common.govukEnvironment }}"
```
sengi, line 31: Somewhat confusingly, the right value for RAILS_ENV is `production` in basically all cases (except when doing local development where you want live-reloading on code changes and so on). Defaulting to the same value as govukEnvironment would cause all sorts of interesting and exciting issues 😀

sengi (commented):
Cool, thanks for the cleanup and especially for sorting out the defaults.

I think the labels usage is spot on 👍

charts/publisher/values.yaml
```diff
@@ -1,121 +1,118 @@
...
+  govukDomainInternal: svc.cluster.local
+
+  image: &commonImage
+    repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/publisher
+    pullPolicy: Always
+    tag: latest
+
+  secrets:
+    externalSecretRef:
+    externalSecretRefreshInterval: "5m"
```
sengi, line 16: Just spotted this - I'd recommend bumping this up a bit to minimise load on etcd. (Secrets Manager can take a hammering but we'll easily have hundreds of secrets once we have Signon and all the apps running, and rotation will be on the order of ~monthly so we should be totally fine with say half an hour or even the default 1h. It might also make it easier to spot transient issues e.g. if a buggy rotation script is revoking previous credentials before all the clients have switched to new ones.)

charts/publisher/templates/web/dbmigration-job.yaml
```diff
@@ -1,21 +1,23 @@
...
 metadata:
   name: {{ .Release.Name }}-dbmigrate-{{ now | unixEpoch }}
+  labels:
+    {{- include "publisher.labels" . | nindent 4 }}
+    app.kubernetes.io/component: "web"
 spec:
   template:
     metadata:
-      name: "{{ .Release.Name }}"
+      name: {{ .Release.Name }}-dbmigrate-{{ now | unixEpoch }}
```
sengi, line 12: Optional: might be marginally more user-friendly to use a human-readable timestamp, e.g. yyyymmddhhmmss (i.e. `now | date "20060102150405"`)

charts/publisher/templates/web/ingress.yaml
```diff
@@ -23,11 +24,11 @@ spec:
...
   {{- end }}
   rules:
-    - host: {{ .Values.ingress.host | default (printf "publisher-%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal ) }}
+    - host: {{ .Values.web.ingress.host | default (printf "publisher-%s.eks.%s.%s" .Release.Namespace .Values.common.govukEnvironment .Values.common.govukDomainExternal ) }}
       http:
         paths:
-          - path: {{ .Values.ingress.path }}
-            pathType: Prefix
+          - path: {{ .Values.web.ingress.path }}
+            pathType: {{ .Values.web.ingress.pathType }}
```
sengi, line 31: Would there be a use case for `Exact`, given the application? (I can't think of one.) Or is the idea to support potential future usage of `ImplementationSpecific` with a different ingress controller?

charts/publisher/templates/web/dbmigration-job.yaml
kerin: Fair enough - this was based on the principle that a worker process might not use exactly the same container as a web process, as the worker needs libraries that the web process doesn't (Sidekiq, resque, etc). That's obviously not the case just now though, so in the spirit of YAGNI I'll remove it - if that changes in the future the chart can obviously be updated.

charts/publisher/templates/web/ingress.yaml
kerin: I'm not sure if there'll ever be a case for a pathType other than `Prefix` as it depends on the specifics of the application's URL handling, and the structure/architecture of the app. It seems reasonable to me to have a sensible default with the ability to override it if necessary though, particularly as `Path` is a configurable value, and these two go hand-in-hand.

sengi (approved):
LGTM

I tweaked the PR description to reflect the 5m -> 1h RefreshInterval change.

charts/publisher/templates/web/dbmigration-job.yaml
sengi: Cool, thanks for fixing. With Rails I doubt we'd want a separate worker image because all the tests implicitly assume that the workers and serving processes are running the same binary; the extra complexity and possibility of version drift would likely outweigh the benefit of a slightly smaller image for workers (which itself might not even be a benefit anyway, because having two images would reduce cache efficacy).

charts/publisher/templates/web/ingress.yaml
sengi: I think it's YAGNI but if you'd prefer to leave it in then fine by me 👍

PR #9, opened 2021-09-29
Helm Chart for Test App

fredericfran-gds:
This is a Helm Chart for a Test App (Nginx) that we will use to explore different ways of deploying apps in GOV.UK EKS.

CD tools under investigation:
1. Argo
2. Flux

Co-Authored-By: Bill Franklin <bilbof@users.noreply.github.com>


bilbof:
Closing as we chose Argo and are no longer testing.

PR #10, opened 2021-10-06
Create pod security policies

karlbaker02:
This PR adds a new Helm chart, `cluster-security`. This chart adds a baseline `PodSecurityPolicy` to the cluster, [in line with the recommendations on the docs](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#recommended-practice). With this cluster-wide policy, containers running as root and with privilege escalation are blocked, whilst containers are also required to run with a read-only root filesystem. This policy is likely to be refined over time and does not represent the final state of the policy in its current form.

Frontend is also updated to comply with the new PSP, although at present due to the existence of `eks.privileged` within the cluster this is not necessary (and I may drop this from the PR).


charts/cluster-security/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,82 @@
...
+metadata:
+  name: psp-baseline
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: '*'
+spec:
+  privileged: false
+  # The moby default capability set, minus NET_RAW
+  allowedCapabilities:
+    - 'CHOWN'
+    - 'DAC_OVERRIDE'
```
sengi, line 14: Is there a particular reason for granting [`DAC_OVERRIDE`](https://manpages.debian.org/bullseye/manpages/capabilities.7.en.html#CAP_DAC_OVERRIDE)? That should normally ring alarm bells.

charts/frontend/templates/deployment.yaml
```diff
@@ -72,6 +76,10 @@ spec:
               port: http
             initialDelaySeconds: 30
             periodSeconds: 5
+          securityContext:
+            allowPrivilegeEscalation: false
+            runAsUser: 2000
+            runAsGroup: 3000
```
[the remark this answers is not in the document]
sengi: If by _real_ you mean do they need to have entries in `/etc/{passwd,group}` or whatever backs [getpwnam(3)](https://man7.org/linux/man-pages/man3/getpwuid.3.html) then (generally speaking) no, any non-zero UID should work.

charts/frontend/values.yaml
```diff
@@ -18,11 +18,11 @@ appImage:
   tag: latest
 
 nginxImage:
-  repository: nginx
+  repository: nginxinc/nginx-unprivileged
   pullPolicy: Always
-  tag: stable
+  tag: 1.21.3
```
sengi, line 23: We'd need a way to keep this up to date. nginx-unprivileged maintains a [`stable` tag](https://hub.docker.com/r/nginxinc/nginx-unprivileged/tags) like the other nginx image, so I'd recommend sticking with that for now. (If you do set an immutable tag (like 1.21.3) then you should remove `pullPolicy: Always` - but I wouldn't do that without an update mechanism in place because it'd be bad if this got forgotten and made it through to production with a pinned version and no automatic updates).

sengi (commented):
Cool - I think this will be really helpful in flushing out issues with our container images.

charts/cluster-security/templates/podsecuritypolicy.yaml
karlbaker02: No particular reason other than taking the boilerplate PodSecurityPolicy (as described in the comment at the top of the file) as the basis which we can iterate on.

charts/frontend/values.yaml
karlbaker02: Cool. The reason I chose the version above was support for `linux/amd64` which is what I _assume_ we require. Looking again now at _stable_ I must have missed this support as it's listed for that flag also.

PR #12, opened 2021-10-14
Use Puma as web server

sophiebrown01:
This PR overrides app `Dockerfile`s to use Puma as the default web server instead of Unicorn and will entirely sidestep the use of Foreman. Foreman isn't necessary when running on Kubernetes, as the controller managers within Kubernetes will take care of ensuring that our app is running without the need of a separate process manager.

Puma is started [as recommended in the docs](https://github.com/puma/puma#rails) by running `bundle exec puma` directly (rather than via `bundle exec rails` which would then use Puma as the default web server).

Co-authored-by: Chris Banks <chris.banks@digital.cabinet-office.gov.uk>

Trello: https://trello.com/c/irrbEFMV

### DIFF
--- a/charts/frontend/templates/deployment.yaml
+++ b/charts/frontend/templates/deployment.yaml
@@ -22,6 +22,8 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.appPort }}
+          command: ["bundle"]
+          args: ["exec", "puma"]
           env:
             - name: ASSET_HOST
               value: "{{ .Values.appAssetHost | default (printf "https://www-origin.%s.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
--- a/charts/publisher/templates/worker/deployment.yaml
+++ b/charts/publisher/templates/worker/deployment.yaml
@@ -25,8 +25,8 @@ spec:
       containers:
         - name: app
           image: "{{ .Values.common.image.repository }}:{{ .Values.common.image.tag }}"
-          command: ["foreman"]
-          args: ["run", "worker"]
+          command: ["bundle"]
+          args: ["exec", "sidekiq", "-C", "config/sidekiq.yml"]
           envFrom:
             - configMapRef:
                 name: {{ .Release.Name }}-env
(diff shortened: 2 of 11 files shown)


charts/frontend/templates/deployment.yaml
```diff
@@ -20,6 +20,8 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.appPort }}
+          command: ["bundle"]
+          args: ["exec", "rails", "server"]
```
kerin, line 24: Should we not specify the port here to match the `containerPort` above?

```
["exec", "rails", "server", "-p", "{{ .Values.appPort }}"]
```

or such?

sengi: Good point but we're already setting it (perhaps in a slightly less clear way) by the `PORT` env var below. In principle we could remove `PORT` and use `-p`, but in practice that'd mean we'd have to change the Dockerfile in a way that's not compatible with its existing usage (in alphagov/publishing-e2e-tests).

kerin: Ah, forgot about that env var - no need to change that to an arg 👍

bilbof: I think the preferred way to set the port would be to set the environment variable (https://github.com/alphagov/frontend/blob/main/config/puma.rb#L18) which is read by the config file.

charts/publisher/templates/worker/deployment.yaml
```diff
@@ -25,8 +25,8 @@ spec:
       containers:
         - name: app
           image: "{{ .Values.common.image.repository }}:{{ .Values.common.image.tag }}"
-          command: ["foreman"]
-          args: ["run", "worker"]
+          command: ["bundle"]
+          args: ["exec", "sidekiq", "-C", "./config/sidekiq.yml"]
```
sengi, line 29: nit: the `./` is superfluous here
```suggestion
          args: ["exec", "sidekiq", "-C", "config/sidekiq.yml"]
```

charts/static/templates/deployment.yaml
```diff
@@ -38,7 +40,7 @@ spec:
             - name: PORT
               value: "{{ .Values.appPort }}"
             - name: RAILS_ENV
-              value: "{{ .Values.govukEnvironment }}"
+              value: "{{ .Values.appRailsEnv }}"
```
sengi, line 43: Good catch!

sengi (approved):

PR #13, opened 2021-10-18
Configure ArgoCD to deploy GOV.UK Helm Charts

bilbof:
This adds configuration required for ArgoCD to install Helm charts in this repo into the `apps` namespace.

It demonstrates that we can use `PreSync` and `PostSync` hooks to run db migrations and post-deploy Slack notifications.

This PR also shows how we can use the [App of Apps pattern](https://argo-cd.readthedocs.io/en/stable/operator-manual/cluster-bootstrapping/) to bootstrap a cluster with applications. We can install all GOV.UK charts in a cluster in one step by installing the ArgoCD config chart, which then triggers the installation of all other apps.

The values file for Helm charts is partially overridden by the values specified in the environment-specific argocd-config Helm chart, e.g. `values-test.yaml`. This file is the only one we should need to modify to change environment-specific configuration for a chart.

See individual commits for clarity.

End result in ArgoCD UI shown below:

```
$ argocd app list

NAME         CLUSTER                         NAMESPACE  PROJECT  STATUS     HEALTH       SYNCPOLICY  CONDITIONS       REPO                                           PATH                  TARGET
argo-config  https://kubernetes.default.svc  apps       govuk    Unknown    Progressing  Auto-Prune  ComparisonError  git@github.com/alphagov/govuk-helm-charts      charts/argocd-config  bilbof/argo
publisher    https://kubernetes.default.svc  apps       govuk    OutOfSync  Missing      Auto-Prune  <none>           git@github.com/alphagov/govuk-helm-charts      charts/publisher      bilbof/argo
signon       https://kubernetes.default.svc  apps       govuk    Synced     Degraded     Auto-Prune  <none>           git@github.com/alphagov/govuk-helm-charts      charts/signon         HEAD
test         https://kubernetes.default.svc  default    default  Synced     Healthy      <none>      <none>           https://github.com/alphagov/govuk-helm-charts  charts/test           test-chart

```

ArgoCD apps

Trello: https://trello.com/c/knISYDSS/675-implement-argocd-pipeline-for-publisher
Design doc: https://docs.google.com/document/d/1EnWc-wMtHx7YNv9uFJL_v97s9Lf4PaSWwjfDF0b1HnI/edit


charts/publisher/templates/web/slack-notify-sync.yaml
```diff
@@ -0,0 +1,57 @@
...
+    spec:
+      containers:
+      - name: slack-notification
+        image: curlimages/curl
+        command:
+          - "curl"
+          - "-X"
+          - "POST"
+          - "--data-urlencode"
+          - "payload={\"channel\": \"#govuk-deploy-alerts\", \"username\": \"argocd\", \"text\": \"{{ .Release.Name }} sync succeeded (image_tag: {{ .Values.common.image.tag }}) :deployparrot:\", \"icon_emoji\": \":kubernetes:\"}"
```
sengi, line 19: Consider using the block scalar syntax here for readability, e.g. something like
```suggestion
          - >
              payload={
                "channel": "#govuk-deploy-alerts",
                "username": "argocd",
                "text": "{{ .Release.Name }} sync succeeded (image_tag: {{ .Values.common.image.tag }}) :deployparrot:",
                "icon_emoji": ":kubernetes:"
              }
```

charts/publisher/templates/web/deployment.yaml
```diff
@@ -26,7 +26,7 @@ spec:
         - name: app
           image: "{{ .Values.common.image.repository }}:{{ .Values.common.image.tag }}"
           command: ["bundle"]
-          args: ["exec", "puma"]
+          args: ["exec", "unicorn -c ./config/unicorn.rb -p ${PORT}"]
```
sengi, line 29: Oh fun, does it not work with puma? Or did something just go wonky during a rebase or something? (only asking cos we had puma working the other day 😅)

charts/publisher/templates/web/slack-notify-sync.yaml
bilbof: Thank _you_ I wasn't aware of this YAML syntax.

charts/publisher/templates/web/deployment.yaml
sengi: Oh yikes - sorry about the breakage! I'll go revert the puma commit.

sengi: Puma stuff reverted in #14.

sengi (approved):
Nice! If I've understood this correctly I think it neatly gives us GitOps-style config management for both the apps and the CD config itself (apart from the minimal initial config to create the top-level Argo "project" in the first place - and there's always going to be a chicken/egg situation there anyway).

I was initially unsure about the use of Helm templates for the app CD configs themselves, but having talked this through in some detail I'm pretty confident that it's a justified abstraction that will pull its weight. If we were to go with an untemplated file per app config per environment, we'd have close to 200 of those and they contain some settings that we might want to consider global and others that we'd want to manage at the app-level, plus a bunch of boilerplate that might change at some point (e.g. on a major Argo version upgrade perhaps). So I think I've talked myself down from "uh oh this looks like a lot of abstraction for what I thought would be a simple fairly flat config" 😅

The only potential challenge that came up in discussion was that (as @kerin pointed out) we may need to be prepared to refactor in the medium term if e.g. teams decide they want to split up their app configs into separate values files or similar - but I think that would be quite a straightforward refactor.

There was a lot of subtle stuff to consider here in terms of requirements and complexity/maintenance tradeoffs here (in an unfamiliar system and a new paradigm) and I think you've navigated this expertly 👍

charts/argocd-apps/templates/application.yaml
```diff
@@ -0,0 +1,55 @@
...
+  destination:
+    server: 'https://kubernetes.default.svc'
+    namespace: apps
+  syncPolicy:
+    automated: # automated sync by default retries failed attempts 5 times with following delays between attempts ( 5s, 10s, 20s, 40s, 80s ); retry controlled using `retry` field.
+      prune: true # Specifies if resources should be pruned during auto-syncing ( false by default ).
+      selfHeal: true # Specifies if partial app sync should be executed when resources are changed only in target Kubernetes cluster and no git change detected ( false by default ).
+      allowEmpty: false # Allows deleting all application resources during automatic syncing ( false by default ).
+    syncOptions:     # Sync options which modifies sync behavior
+    - Validate=false # disables resource validation (equivalent to 'kubectl apply --validate=false') ( true by default ).
```
kerin, line 36: What's the rationale for disabling resource validation? Basic pre-apply validation seems like a good thing with little to no downsides, unless there's some issue caused elsewhere with Argo or Helm?

charts/argocd-apps/values-test.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+applications:
+- name: cluster-secrets
+- name: publisher
+  helm:
+    parameters:
+    - name: common.govukEnvironment
+      value: "test"
+    - name: common.image.repository
+      value: "172025368201.dkr.ecr.eu-west-1.amazonaws.com/publisher"
```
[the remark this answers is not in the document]
kerin: Agree that it's useful to separate the registry and image name - that's primarily a change for the Helm charts themselves though, so we should change those first then update the argo values to match.

kerin (approved):

charts/argocd-apps/templates/application.yaml
bilbof: These options were included in the default application.yaml file provided in the docs (@sengi linked to the relevant section in the docs which explains why: https://github.com/alphagov/govuk-helm-charts/pull/13#discussion_r731677751). I've removed this config as it isn't currently required.

charts/argocd-apps/templates/application.yaml
```diff
@@ -0,0 +1,55 @@
...
+        duration: 5s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
+        factor: 2 # a factor to multiply the base duration after each failed retry
+        maxDuration: 3m # the maximum amount of time allowed for the backoff strategy
+
+  # Ignore differences at the specified json pointers
+  ignoreDifferences:
+  - group: apps
+    kind: Deployment
+    jsonPointers:
+    - /spec/replicas # for autoscaling
```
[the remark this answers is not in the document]
bilbof: Agreed, I've removed this.

PR #15, opened 2021-10-21
Add existing GOV.UK Apps to ArgoCD

fredericfran-gds:
Remaining existing GOV.UK Apps are added to ArgoCD:
1. frontend
2. static
3. content-store
4. router

In addition, we add the slack notification after the app is deployed by ArgoCD

Ref:
1. [trello card](https://trello.com/c/GqZuxOeR/699-create-argo-applications-for-remaining-apps)

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -36,3 +36,47 @@ applications:
       value: "1"
     - name: worker.replicas
       value: "1"
+- name: router
+  helm:
+    parameters:
+    - name: govukEnvironment
+      value: "test"
+    - name: image.repository
+      value: "172025368201.dkr.ecr.eu-west-1.amazonaws.com/router"
+    - name: image.tag
+      value: "latest" # Automatically updated
+    - name: replicaCount
+      value: "1"
+- name: frontend
+  helm:
+    parameters:
+    - name: govukEnvironment
+      value: "test"
+    - name: image.repository
+      value: "172025368201.dkr.ecr.eu-west-1.amazonaws.com/frontend"
+    - name: image.tag
+      value: "latest" # Automatically updated
+    - name: replicaCount
+      value: "1"
+- name: static
+  helm:
+    parameters:
+    - name: govukEnvironment
+      value: "test"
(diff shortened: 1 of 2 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -36,3 +36,47 @@ applications:
       value: "1"
     - name: worker.replicas
       value: "1"
+- name: router
+  helm:
+    parameters:
+    - name: govukEnvironment
+      value: "test"
```
sengi, line 43: It's a shame that this way we end up with the `{name: "foo", value: "bar"}` antipattern rather than `{foo: "bar"}`, times hundreds of config items times 3-4 environments. Maybe it's the lesser of evils for now?

I'm starting to think that once Argo starts supporting reading Helm values files from alternative repos we might want to switch back to using ordinary values files for these configs, which would be way more succinct and readable. It'd also neatly solve the problem that all the apps' configs are lumped together in one file rather than separated by app (which isn't a problem for now, but might potentially become one eventually depending on whether the restructured dev teams are happy with sharing config files).

charts/content-store/templates/slack-notify-sync.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: batch/v1
```
sengi, line 1: Why is this in the app Helm charts? Surely it's conceptually part of the CD config so would make more sense in argocd-apps - and then you'd probably be able to avoid copy-pasting is 65+ times 😅 (which I think probably wouldn't be acceptable from a maintenance perspective - future GOV.UK developers would curse us if they ever wanted to change the way the Slack notifications worked!)

charts/argocd-apps/values-test.yaml
sengi, line 43: Also, how do we feel about encapsulating Helm values files inside a Helm values file like this? Seeing it in practice, it seems kinda unnecessarily meta to me. I still think it the idea of making the values part of the Application resource sounds attractive in itself, but I hadn't anticipated this rather ugly consequence of it.

PR #19, opened 2021-11-08
Keep PrometheusRule resources in same namespace as apps.

sengi:
We don't want the app Helm charts to create resources in global/shared namespaces unnecessarily, otherwise we prevent copies of apps from being deployed in other namespaces.

Also some trailing whitespace cleanup.


karlbaker02 (approved):
Looks good and makes sense 👍

PR #20, opened 2021-11-08
Adopt Argo Events and Argo Workflows for post-deploy smoke tests

bilbof:
This adds configuration for Argo (events, workflows, cd, and cd-notifications) in order to run smoke tests for an application after it has been deployed by ArgoCD and is healthy in the cluster.

This adds a procedural step to the current GitOps model, following a successful sync the following process will be taken:

1. ArgoCD Notifications will send a webhook to Argo Events with context of the sync
2. Argo Events triggers an Argo Workflow named "post-sync"
3. Argo Workflows runs the post-sync workflow, which runs the smoke tests for the application and send the result to a Slack channel.

No additional Argo configuration is required when setting up other applications.

Other approaches implemented:

* Copy post-sync Job manifest to each application Helm chart
* Use Helm Library chart to share post-sync Job between Helm charts

Both of these alternatives worked, but we found that running 50+ applications would require more configuration, and we may be re-implementing some Argo Workflows functionality.

The main trade-off of this spiked approach is that we need to use the Argo Events and Workflows products, rather than using our existing toolset (Helm and Kubernetes). We think using Argo Workflows is justified here, as it enables us to run more complex workflows (e.g. easily run smoke tests for some apps, but not for others) without additional configuration. The resulting configuration should eventually be less complicated and more maintainable than other approaches. Using Argo Workflows also opens the door for us to implement complex pipelines unrelated to CD natively in-cluster (i.e. data science pipelines), a use case Kubernetes does not support well out-of-the-box.

We should be aware that Argo is in early development. Argo components generally lack detailed, up-to-date, and comprehensive documentation, and lack consistency between components, which made configuring Argo difficult. Argo components are distinct projects (e.g. there use different templating languages) and their inconsistencies can be confusing.

Co-Authored-By: @fredericfran-gds 

Trello: https://trello.com/c/h8F7nBbu/676-smoke-tests-after-each-deploy-by-app


charts/argo-services/templates/events/sync_notification_sensor.yaml
```diff
@@ -0,0 +1,59 @@
...
+              dataKey: body.commitSha
+          - dest: spec.arguments.parameters.2.value
+            src:
+              dependencyName: sync-notification
+              dataKey: body.commitUrl
+          - dest: spec.arguments.parameters.3.value
+            src:
+              dependencyName: sync-notification
+              dataKey: body.argoUrl
+          - dest: spec.arguments.parameters.4.value
```
sengi, line 38: Oof. It's not the most elegant CRD, this, is it? 😅 I couldn't see a better way either, though 🤷 I think it'll be ok since we're not forcing teams to work with this stuff day-to-day - should be pretty low-maintenance with any luck.

charts/argo-workflows/templates/notify-slack-workflow.yaml
```diff
@@ -0,0 +1,37 @@
...
+      container:
+        image: curlimages/curl
+        command:
+          - "curl"
+          - "-X"
+          - "POST"
+          - "--data-urlencode"
+          - >
+              payload={
+                "channel": "#{{"{{inputs.parameters.slackChannel}}"}}",
```
sengi, line 26: optional: Might be worth a comment here about the double-templating - probably quite useful for the reader to know that it's Jinja inside Go template or whatever it is. Similarly for the one in post-sync-workflow.yaml.

Or could pull this out into a file too, but that's probably overkill here since having to look through a zillion files also detracts from readability 🤷

charts/argo-workflows/templates/post-sync-workflow.yaml
```diff
@@ -0,0 +1,45 @@
...
+            template: notify-slack
+          arguments:
+            parameters:
+              - name: slackChannel
+                # NOTE: Change to {{"{{workflow.parameters.slackChannel}}"}} to
+                # send to team slack channel.
+                value: "govuk-deploy-alerts"
+              - name: text
+                value: "post-sync job for {{"{{workflow.parameters.application}}"}} has status {{"{{workflow.status}}"}} :deployparrot:\n
+                <{{"{{workflow.parameters.commitUrl}}"}}|{{"{{workflow.parameters.commitSha}}"}}> • <{{"{{workflow.parameters.argoUrl}}"}}|{{"{{workflow.parameters.application}} Application"}}>."
```
sengi, line 45: Ooh, utf-8! I'm guessing this all works cleanly end-to-end? 😅 (How things have moved on!)

charts/smokey/templates/workflows/smokey-loop.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: smokey-loop
+spec:
+  schedule: "*/3 9-17 * * 1-5"
+  timezone: Europe/London
+  concurrencyPolicy: "Replace"
+  startingDeadlineSeconds: 0
```
sengi, line 9: I don't think startingDeadlineSeconds is worth specifying here - the default is already 0 and it doesn't seem like something worth calling the reader's attention to.
```suggestion
```

charts/smokey/templates/workflows/smokey-loop.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: CronWorkflow
+metadata:
+  name: smokey-loop
+spec:
+  schedule: "*/3 9-17 * * 1-5"
+  timezone: Europe/London
+  concurrencyPolicy: "Replace"
```
sengi, line 8: nit: unnecessary quotes
```suggestion
  concurrencyPolicy: Replace
```

sengi (approved):
Thanks for the thoughtful PR description - the context, tradeoffs, and alternatives considered were very helpful in reviewing.

LGTM - let's go with this and see how it pans out 🚀

sengi:
Would be cool to link to the [design doc](https://docs.google.com/document/d/1EnWc-wMtHx7YNv9uFJL_v97s9Lf4PaSWwjfDF0b1HnI/edit) in the description, if you want.

charts/argo-services/templates/events/service-account.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright: The ArgoProj maintainers
+# https://github.com/argoproj/argo-events/tree/master/examples
```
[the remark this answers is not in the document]
fredericfran-gds: cool, will do

charts/argo-services/templates/events/sync_notification_event_source.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: argoproj.io/v1alpha1
+kind: EventSource
+metadata:
+  name: sync-notification
+spec:
+  # TODO: service should be replaced before production
+  # https://argoproj.github.io/argo-events/eventsources/services/
```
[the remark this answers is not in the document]
fredericfran-gds: Will do, thanks.

charts/smokey/templates/workflows/smokey-loop.yaml
fredericfran-gds: will do, thanks

charts/smokey/templates/workflows/smokey-loop.yaml
fredericfran-gds: will do, thanks

charts/argo-workflows/templates/notify-slack-workflow.yaml
fredericfran-gds: done thanks

PR #21, opened 2021-11-09
Fix name clashes on ALBs created by Ingresses.

sengi:
By default, aws-lb-controller creates load balancers with random strings as names. We want to be able to look up the LB for a given Ingress easily and vice versa. We previously acheived this using fixed names, but these would clash when deploying the same apps into more than one namespace for testing.

Include the namespace in the ALB name so that the names are unique while also being human-friendly.

This doesn't cater for overriding the ALB name via Helm values, but I can't think of any valid reason for wanting to do that at this stage. We can easily add a value for it later if we need it.

Also substitute ingress.name in the hostname instead of hardcoding. This doesn't change the output for any of our current use cases; just makes things more flexible.

Tested: charts produce valid kube yaml and apps come up successfully on install.

### DIFF
--- a/charts/publisher/Chart.yaml
+++ b/charts/publisher/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: publisher
 description: A Helm chart for GOV.UK Publisher
 type: application
-version: 0.8.0
+version: 0.8.1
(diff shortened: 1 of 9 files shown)


karlbaker02 (approved):
All looks good 👍

PR #22, opened 2021-11-10
Fixup ArgoCD Notifications ConfigMap

bilbof:
This was broken in commit 4f9b363154fe6cc24fe903b82ff4b5c46159278f
when moving the configmap. Example of what it should have looked
like: https://github.com/argoproj-labs/argocd-notifications/blob/master/docs/argocd-notifications-cm.yaml

### DIFF
--- a/charts/argo-services/cd-notifications-config.yaml
+++ b/charts/argo-services/cd-notifications-config.yaml
@@ -1,17 +1,17 @@
 # NOTE: this file is pulled into templates/argocd-notifications/cm.yaml
 # to avoid needing to nesting conflicting template tags (Helm uses Go, Argo
 # uses Jinja2 which both use the `{{}}` syntax).
-service.webhook.argo_events:
+service.webhook.argo_events: |
   headers:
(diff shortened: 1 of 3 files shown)


charts/argo-services/cd-notifications-config.yaml
```diff
@@ -1,33 +1,33 @@
...
+          "argoUrl": "{{.context.argocdUrl}}/applications/{{.app.metadata.name}}",
           "slackChannel": "{{.app.metadata.annotations.slackChannel}}"
         }
       method: "POST"
       path: "/argocd"
-template.send-slack:
-  webhook":
-    slack_webhook":
+template.send-slack: |
+  webhook": |
```
sengi, line 29: What's up with these unbalanced quotes? (similarly below)

bilbof: Oops, sorry about that -- fixed!

sengi (approved):
👍

PR #23, opened 2021-11-12
Move common env vars to a shared ConfigMap.

sengi:
We introduce a new `govuk-apps-env` ConfigMap to represent the set of environment variables which are common to all the GOV.UK apps (plus some which will have a couple of overrides in the draft apps - but that's ok). These env vars are included in the app Deployment/Pod specs via `envFrom`.

The CM is installed by a Helm chart `govuk-apps-conf` with relatively few values passed in, since the idea is that this chart represents only config which is common to most of the apps in an environment. There might end up being other shared ConfigMaps and/or ExternalSecrets that logically belong in here, hence not calling it `-env`. Open to better naming ideas 🙃

This PR covers `frontend` and `content-store`. In the interests of keeping this PR manageable and not having a long-lived branch, I'll give the other charts the same treatment in followup PRs (assuming we're happy with the general pattern here).

Also:
- Remove some unused/unnecessary/inappropriate env vars.
- Rename `configmap.yaml` to `nginx-configmap.yaml` for clarity.
- Sort out the probes to avoid the unnecessary ~30 s delay on startup. Kinda unrelated but it made testing a lot less annoying 😛 

The following are the same for all apps in a given environment, so they're in the shared ConfigMap:

- `ASSET_HOST`
- `GOVUK_APP_DOMAIN`: (not actually used anyway because we're setting all the `PLEK_SERVICE_*_URI` explicitly, but Plek requires it for now)
- `GOVUK_APP_DOMAIN_EXTERNAL`
- `GOVUK_ASSET_ROOT`
- `GOVUK_ENVIRONMENT`
- `GOVUK_WEBSITE_ROOT`
- `PLEK_SERVICE_*_URI`: (except for draft apps, which will override one or
  two of these)

The following are unnecessary or otherwise don't belong in the app chart:

- `GOVUK_APP_NAME`: this is an inherent property of the app and doesn't need to be overridden (except for draft apps). It should probably be removed entirely at some point, but in the meantime we'll just set it to `.Release.Name`. That covers the usage for draft apps, makes it easy to refactor later (because it's not part of the public interface to the chart) and covers us in case a library gem like govuk_app_config were to start using it in some new way.
- `GOVUK_APP_TYPE`: used only in govuk-puppet, not in the app
- `DEFAULT_TTL`: only relevant to Content Store.
- `GOVUK_CONTENT_SCHEMAS_PATH`: only relevant to Publishing API and test
runners. See [design doc](https://drive.google.com/open?id=1nxJzdbTK__jWDFjUj9C5sH4yP0caIHhXp6fx1yKGHw).
- `RAILS_ENV`: this is a property of the container image; it would never make sense to run a production image in dev mode - you'd need a different image that includes the dev dependencies.
- `RAILS_LOG_TO_STDOUT`: this always has to be set when running in a container, so it belongs in the Dockerfile.

Normally a change like this would merit a major version bump, but I've made it a minor because the major is still 0 and we're the only user. It seemed wrong to go to 1.0 at this point as that would be (more) misleading.

#### Tips for reviewing

I'd recommend reviewing commit-by-commit.

#### Tested

Deployed `frontend` and `content-store` using `helm install` and they both still work. Logs look ok and frontend is able to serve content from content-store.

[Trello card](https://trello.com/c/h6lGL53t/706)


charts/frontend/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v2
 name: frontend
 description: A Helm chart for GOV.UK Frontend
 type: application
-version: 0.6.2
+version: 0.7.0
```
sengi, line 5: Normally this would be a major version bump because I'm removing stuff from the public interface, but I've made it a minor because going to 1.0 for this would be even more misleading IMO.

charts/govuk-apps-conf/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
sengi, line 1: This file is just the standard boilerplate .helmignore from `helm create`.

charts/govuk-apps-conf/templates/_helpers.tpl
```diff
@@ -0,0 +1,61 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "govuk-apps-conf.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "govuk-apps-conf.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
sengi, line 51: Lines 1-51 are `helm create` boilerplate. I removed one template that was definitely unused (the service account one).

kerin (approved):
🙌 LGTM

PR #24, opened 2021-11-17
Add values file for the integration environment.

sengi:
This is just a temporary bodge to get the integration environment up and running for user testing this week. We'll need a much cleaner way to represent the common config items which are being duplicated many times within and between these per-environment values files before we go to prod.

I know the Redis that this refers to won't exist; I'll have another PR to address that shortly.


kerin (approved):
LGTM 👍

PR #25, opened 2021-11-17
Use Redis cluster from the EKS project; the ECS one is going away.

sengi:
This reflects the new DNS names for these Redis clusters from alphagov/govuk-infrastructure#483.

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -19,7 +19,7 @@ applications:
     - name: web.dbMigrationEnabled
       value: "true"
     - name: common.env.redisUrl
-      value: "redis://redis.ecs.integration.govuk-internal.digital:6379"
+      value: "redis://shared-redis.eks.integration.govuk-internal.digital"
     - name: web.replicas
       value: "1"
     - name: worker.replicas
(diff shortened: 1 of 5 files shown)


charts/argocd-apps/values-integration.yaml
```diff
@@ -19,7 +19,7 @@ applications:
     - name: web.dbMigrationEnabled
       value: "true"
     - name: common.env.redisUrl
-      value: "redis://redis.ecs.integration.govuk-internal.digital:6379"
+      value: "redis://shared-redis.eks.integration.govuk-internal.digital"
```
sengi, line 22: FYI the port number is definitely optional according to [the redis URI scheme spec](https://www.iana.org/assignments/uri-schemes/prov/redis).

karlbaker02 (approved):
All looks good to me 👍

charts/publisher/values.yaml
```diff
@@ -53,7 +53,7 @@ common:
     plekServicePublisherUri:
     plekServiceStaticUri:
     railsEnv: "production"
-    redisUrl: "redis://redis.ecs.test.govuk-internal.digital:6379"  # TODO: decide on Redis for EKS
+    redisUrl: "redis://redis"  # TODO: decide on Redis for EKS
```
sengi, line 56: This is supposed to be just a generic example rather than the values specific to our AWS stack, which are currently in the argocd-apps chart. Hence just setting it to a plausible value that illustrates how to use it, rather than the specific hostname of one of our environments' Redises.

PR #28, opened 2021-11-22
Set govukStack correctly for the integration environment.

sengi:
Also remove `govukStack` from static's values.yaml because it's unused there.

`govukStack` is a terrible wart, but we can't quite get rid of it just yet. We need to change the hostnames of MongoDB cluster nodes in the test account from `pink` to `blue` first, so until we get around to doing that it'll stick around, complicating up the Helm charts.

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.2.3
+version: 0.2.4
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -46,6 +46,8 @@ applications:
     parameters:
     - name: govukEnvironment
       value: "integration"
+    - name: govukStack
+      value: "blue"  # TODO: eliminate the need for govukStack
     - name: image.repository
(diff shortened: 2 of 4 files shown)


fredericfran-gds (approved):
Maybe we should have added `blue` as the default stack in the helm charts as 3 gov.uk environment uses that and overridden it to `pink` in the test account.

I think why `pink` was used to balance perceived gender bias linked with the colour `blue`.

sengi:
> Maybe we should have added blue as the default stack in the helm charts as 3 gov.uk environment uses that and overridden it to pink in the test account.

I had the same thought, but when I went to try it I found that the other stuff in those chart values.yamls were set for the test account, so I felt like I would have left it a bit incoherent if I'd just changed the `pink` to `blue` without rejigging the whole lot. I'm hoping we can just rename those Mongo nodes in the test account at some point and then we can delete the whole stack thing 😅 If not then I think it's a good idea to change the default.

> I think why `pink` was used to balance perceived gender bias linked with the colour `blue`.

🤯😅 [Ah yes, the three genders](https://knowyourmeme.com/memes/ah-yes-the-three-genders). `pink`, `blue` and [`govuk`](https://docs.publishing.service.gov.uk/manual/deploying-terraform.html#1-gds-cli) 😂

sengi:
I think `beige` and `brown` would have been more fitting imo 😂

PR #39, opened 2021-12-01
Avoid duplicate lint checks on first-party PRs.

sengi:
#### Alternatives considered

* We could just remove `pull_requests` from the `on` predicate. That'd mean we don't run checks automatically on third-party PRs, which has pros and cons: better for security, bit inconvenient for accepting third-party contributions (but this happens so rarely that we're probably talking about a handful of minutes a year).
* There's also https://github.com/marketplace/actions/skip-duplicate-actions, but that looks like it could become complex to configure.

#### Tested

Screenshot 2021-12-01 at 11 39 50

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,11 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bilbof (approved):
🚀 Awesome

PR #41, opened 2021-12-03
Factor out common env vars from `publisher` chart.

sengi:
See #23 for context, which is basically the same changes for `frontend` and `content-store`. Also see individual commit messages for more details.

- Remove common env vars from the Publisher chart and use the shared `govuk-apps-env` ConfigMap.
- Revert to the conventional Helm template for Ingress.
- Restructure the Helm values to be somewhat flatter / reduce excessive nesting.
- Use the external domain suffix for the Signon URL. This [trombones](https://en.wikipedia.org/wiki/Hairpinning) the internal Signon traffic but we have to live with it until we fix [Plek](https://github.com/alphagov/plek).
- Allow the chart to be used without ExternalSecrets, because we shouldn't be tying any of this stuff to our AWS environment (for general flexibility and testability).
- A bunch more general cleanup and refactoring.

Tested: installed in namespace `chris` in the test cluster, apps serves ok, all seems to be running without errors. The Signon login doesn't fully work because I haven't done the pointy-clicky Signon config/secrets dance for my test namespace.


charts/publisher/templates/common/external-secret.yaml
```diff
@@ -1,15 +1,19 @@
...
     name: aws-secretsmanager
     kind: ClusterSecretStore
   target:
-    name: {{ .Release.Name }}
+    name: {{ .Release.Name }}-env
+  # TODO: Use individual AWS secrets instead of lumping them all together,
+  # so that we can have automated rotation, proper audit trail, etc.
   dataFrom:
-    - key: "{{ .Values.common.secrets.externalSecretRef }}"
+    - key: govuk/publisher
```
kerin, line 18: Maybe minor, but should we not retain this as an explicit Helm value with this value as default, so that it's not tied to a hardcoded key name, to allow multiple deployments with different secrets values in the same AWS account?

kerin (approved):
One query, otherwise LGTM

charts/publisher/templates/common/external-secret.yaml
sengi: I completely agree we should be passing these in — I'm only hardcoding this temporarily because I don't want to bake the temporary everything-in-one-AWS-secret approach into the user interface of the chart. Once we've split the secrets up, I expect we'll want to pass a block of YAML in for the dataFrom field, or something like that. I've added a TODO to call that out specifically so I don't forget.

sengi: (yeah I know literally passing in YAML for dataFrom won't work — probably end up being a list or a map that we expand to a bunch of `data` fields or something like that, possibly even passing the whole spec, though that might be a bit much)

sengi:
Realised I forgot to update the values that Argo is passing in. One more commit incoming - sorry.

PR #46, opened 2021-12-06
Lint the per-environment values files for argocd-apps.

sengi:
`chart-testing` will test against all files matching the glob `ci/*-values.yaml` under the chart directory. Bizarrely, this doesn't appear to be configurable.

Sadly this still won't pick up errors like the one fixed in #45. This is because `helm lint` doesn't have a way to provide it the CRDs against which to validate the output. In order to catch issues like that, we'd need to do `ct lint-and-install`, which needs a cluster to install into. We might be able to use the test cluster for this, but that sounds complicated (not least from a security perspective) so I'm not going to look into that any time soon.

See https://github.com/helm/charts/blob/master/test/README.md#providing-custom-test-values

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.3.4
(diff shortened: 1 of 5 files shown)


sengi:
Rebased to resolve merge conflict on chart version number.

PR #51, opened 2021-12-09
GOVUK default Rails application Helm chart.

nsabri1:
Adding govuk-rails-app default chart, to be used to deploy new rails app which are somewhat similar.

Ingress, worker and db-migrations can be enabled via true/false value.

TODO as part future work:
- External secrets strategy is yet to be finalised. 

Documentation:
- Brief guide to be added once testing is completed.

Manual Testing:
helm install frontend . --namespace nadeem --set ingress.enabled=true --set ingress.name=frontend --set appImage.repository=172025368201.dkr.ecr.eu-west-1.amazonaws.com/frontend --set 'appExtraEnv[0].name=TEST_ENV' --set 'appExtraEnv[0].value=testvalue'

Trello card: https://trello.com/c/8hYNGIjG/659-write-up-a-starter-helm-chart-for-our-template

### DIFF
--- a/charts/govuk-rails-app/templates/_helpers.tpl
+++ b/charts/govuk-rails-app/templates/_helpers.tpl
@@ -1,0 +1,62 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "govuk-rails-app.name" -}}
+{{- default .Release.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "govuk-rails-app.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Release.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "govuk-rails-app.chart" -}}
+{{- printf "%s-%s" .Release.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "govuk-rails-app.labels" -}}
+helm.sh/chart: {{ include "govuk-rails-app.chart" . }}
+{{ include "govuk-rails-app.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "govuk-rails-app.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "govuk-rails-app.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "govuk-rails-app.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "govuk-rails-app.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
--- a/charts/govuk-rails-app/templates/deployment.yaml
+++ b/charts/govuk-rails-app/templates/deployment.yaml
@@ -1,0 +1,86 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    app: {{ .Release.Name }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}
+    spec:
+      securityContext:
+        runAsNonRoot: true
+      containers:
+        - name: {{ .Release.Name }}
+          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          ports:
+            - name: http
+              containerPort: {{ .Values.appPort }}
+          envFrom:
+          - configMapRef:
+              name: govuk-apps-env
+          env:
+            - name: PORT
+              value: "{{ .Values.appPort }}"
+            - name: SECRET_KEY_BASE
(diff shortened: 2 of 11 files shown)


charts/govuk-rails-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "govuk-rails-app.name" -}}
+{{- default .Release.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
```
nsabri1, line 5: `Chart.Name` has been replaced with `Release.Name`

charts/govuk-rails-app/templates/ingress.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
+  rules:
+  # TODO: Import GOVUK_APP_DOMAIN_EXTERNAL from global config map and apend to host name.
+    - host: {{ .Values.ingress.host | default (printf "%s-%s" .Values.ingress.name .Release.Namespace ) }}
```
nsabri1, line 27: This is where GOVUK_APP_DOMAIN_EXTERNAL will be added to end of string

charts/govuk-rails-app/values.yaml
```diff
@@ -0,0 +1,46 @@
+# Default/example values for a Rails application.
+
+service:
+  type: NodePort
+  annotations: {}
+  port: 80
+
+ingress:
+  enabled: false
+  name:
```
nsabri1, line 10: This can be set by ArgoCd if ingress has been enabled

charts/govuk-rails-app/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+    alb.ingress.kubernetes.io/target-type: ip
+    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
+    alb.ingress.kubernetes.io/ssl-redirect: "443"
+    alb.ingress.kubernetes.io/healthcheck-path: /healthcheck/ready
+
+replicaCount: 1
+
+appImage:
+  # TODO: point this at a public repo and override to the private repo in our CD config.
+  repository:
```
nsabri1, line 25: ArgoCD will set this value

charts/govuk-rails-app/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# resources:
+#   limits:
+#     cpu: 500m
+#     memory: 512Mi
+#   requests:
+#     cpu: 500m
+#     memory: 512Mi
+
+# appExtraEnv:
```
nsabri1, line 44: App specific extra env values can e set by ArgoCD ```--set 'appExtraEnv[0].name=TEST_ENV' --set 'appExtraEnv[0].value=testvalue'```

charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}
+    spec:
+      securityContext:
+        runAsNonRoot: true
+      containers:
+        - name: {{ .Release.Name }}
+          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
```
kerin, line 21: It'd be good to mark these two values as [`required`](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-required-function), given that the defaults are (rightly) null, so that a `helm install` without these values fails as fast as possible (render time rather than apply time).

charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+            <<: *nginx-probe
+            failureThreshold: 2
+          volumeMounts:
+          - name: {{ .Release.Name }}-nginx-conf
+            mountPath: /etc/nginx/nginx.conf
+            subPath: nginx.conf
+          - name: tmp
+            mountPath: /tmp
+      volumes:
+      - name: tmp
```
kerin, line 82: Is the `/tmp` volume and mount required?

charts/govuk-rails-app/templates/nginx-configmap.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-nginx-conf
+data:
+  nginx.conf: |-
```
kerin, line 6: Something for later, but we should validate the contents of nginx.conf at some stage - do we need to change the defaults for `client_max_body_size` and similar, configure log levels or output format, etc.

charts/govuk-rails-app/templates/deployment.yaml
nsabri1: Good shout on this one.

charts/govuk-rails-app/templates/deployment.yaml
nsabri1: I will look into this.

nsabri1: removed

charts/govuk-rails-app/values.yaml
sengi: I think we should call this `extraEnv` (which is conventional) and have it apply to the worker as well. We don't actually want the environment to be able to vary between the worker and "web" deployments. (see Publisher for example)

charts/govuk-rails-app/templates/nginx-configmap.yaml
nsabri1: I agree - i believe there is a Nginx related Trello card - i will make a note: https://trello.com/c/Lkb8N313/381-epic-nginx-and-varnish-changes-needed

charts/govuk-rails-app/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - path: /
+      pathType: Prefix
+
+replicaCount: 1
+
+# Enable to true if application requires a worker container.
+workerEnabled: false
+workerReplicaCount: 1
+
+# Enable true for Database migration
```
sengi, line 37:
```suggestion
# dbMigrationEnabled controls whether Rails database migrations are run on install/upgrade.
```

charts/govuk-rails-app/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+    alb.ingress.kubernetes.io/healthcheck-path: /healthcheck/ready
+  hosts:
+  - host:
+    paths:
+    - path: /
+      pathType: Prefix
+
+replicaCount: 1
+
+# Enable to true if application requires a worker container.
```
sengi, line 33:
```suggestion
# workerEnabled should be true if the app uses Sidekiq, false otherwise.
```

charts/govuk-rails-app/values.yaml
```diff
@@ -0,0 +1,70 @@
...
+service:
+  type: NodePort
+  annotations: {}
+  port: 80
+
+externalSecrets:
+  enabled: true
+  refreshInterval: 60m
+
+# enable true if Application requires public facing endpoint
```
sengi, line 12: This is just the standard Helm ingress template from the default starter chart, so doesn't really need a comment. (The default starter chart doesn't have a comment here either.)
```suggestion
```

sengi (approved):
Cool, looks good - let's see if this crazy idea works out 😅 🚀 

Just added a few trivial suggestions, mostly for making the values comments match the conventional style.

PR #54, opened 2021-12-10
Increase number of frontend replicasets

beccapearce:
This has been done as part of user testing.

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.3.10
+version: 0.3.11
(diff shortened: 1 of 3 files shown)


charts/argocd-apps/Chart.yaml
```diff
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.3.10
+version: 0.3.11
```
sengi, line 5: No need to bump this version but don't worry about it :)

sengi (approved):

charts/argocd-apps/Chart.yaml
sengi: Oh wait actually it's kinda the other way around — feel free to merge this and try it out though!

beccapearce: 🤞🏻

PR #56, opened 2021-12-14
Use YAML for Helm values instead of setting key/value params.

sengi:
This means we have the same syntax when passing Helm values via Argo as when deploying using Helm directly, which makes testing easier as well as making the config more readable.

Also:
* Pass `govukEnvironment` into every app chart instead of repeating it for every app.
* Fix the confusing comments about `tag: latest`. The idea is that those lines will be edited by release automation (an Argo post-sync hook), but we haven't implemented that yet. This PR should make implementing that automation a bit easier since everything is now plain YAML instead of the clunky `{name: x, value: y}` format.

Tested: tried it on the test cluster by editing app/argocd-apps to point at the branch (`kubectl edit app argocd-apps`). The Argo apps seem to work as expected.

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.3.13
+version: 0.4.0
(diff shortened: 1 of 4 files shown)


sengi:
No probs. Wasn't sure if there wasn't some subtle difference in semantics between `helm.parameters` and `helm.values` (other than precedence), but afaict it does the same thing.

PR #57, opened 2021-12-14
Use ArgoCD finalizers in test and staging envs only.

sengi:
In prod (and probably staging, just for prod-fidelity) we don't want Argo to delete an application's resources on deletion of its Argo app resource. In production it's more likely that Argo apps would be deleted temporarily or accidentally than with the intent of removing all trace of the application.


bilbof (approved):
🦸

PR #60, opened 2021-12-22
Create Signon Resources on cluster turnup

bilbof:
Signon Resources (Applications, ApiUsers, Authorisations) will be created automatically during cluster turnup.

Changes:

- Adds a Signon Resources Helm chart (contains a job that creates the Signon Resources by calling the Rake task implemented in https://github.com/alphagov/govuk-infrastructure/pull/499 which in turn calls the Signon API and the Kubernetes API to create the Secrets required by other charts)
- Updates other charts to use the Secrets generated by the Signon Resources chart

Since the Helm chart design is being considered elsewhere, please consider the (inconsistent) way we set secrets/env vars across charts as only temporary. 

Trello: https://trello.com/c/fUXJ6X1U/725-create-signon-resources-when-signon-is-deployed

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.5.0
+version: 0.5.1
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -57,3 +57,4 @@ applications:
     replicaCount: 1
     appMongodbUri: mongodb://mongo-1.integration.govuk-internal.digital,mongo-2.integration.govuk-internal.digital,mongo-3.integration.govuk-internal.digital/content_store_production
 - name: argo-workflows
+- name: signon-resources
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -58,3 +58,4 @@ applications:
     replicaCount: 1
     appMongodbUri: mongodb://mongo-1.test.govuk-internal.digital,mongo-2.test.govuk-internal.digital,mongo-3.test.govuk-internal.digital/content_store_production
 - name: argo-workflows
+- name: signon-resources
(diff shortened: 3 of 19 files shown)


charts/publisher/templates/worker/deployment.yaml
```diff
@@ -35,7 +35,13 @@ spec:
             - secretRef:
                 name: {{ .Release.Name }}-env
           env:
-            {{- .Values.extraEnv | toYaml | nindent 12 }}
+          {{- range $var_name, $secret := .Values.service.secrets }}
```
sengi, line 38: I see where you're coming from in reformatting the valueFrom/secretKeyRef stuff like this, but I'm not convinced it's worth the complexity of introducing a new format and transforming it in template code, vs. just passing in the spec YAML directly in its native form.

The extra layer of string-template stuff just means we're passing stuff like this in the values files:
```
  secrets:
    GDS_SSO_OAUTH_ID:
      name: signon-app-publisher-eks
      key: oauth_id
    GDS_SSO_OAUTH_SECRET:
      name: signon-app-publisher-eks
      key: oauth_secret
```

versus this, without the templating:

```
          env:
            - name: GDS_SSO_OAUTH_ID
              valueFrom:
                secretKeyRef:
                  name: signon-app-publisher-eks
                  key: oauth_id
            - name: GDS_SSO_OAUTH_SECRET
              valueFrom:
                secretKeyRef:
                  name: signon-app-publisher-eks
                  key: oauth_secret
```

So might as well just pass the second format verbatim and avoid the string-template-programming.

Moreover, I think we can avoid the hassle of passing this stuff in via values in the first place. The names of the secrets are really just part of the Deployment's interface, so it's reasonable to just define this stuff directly in the Deployment and not let the user change the names of the secrets. Don't worry about that for now though unless it makes things easier for this PR — I get that you didn't intend to fix up all the secrets stuff as part of this.

sengi (approved):
LGTM. Feel free to ignore my comments about the secrets stuff — just thought I'd give some thoughts/context on that. I'll be working on fixing up the secrets stuff shortly.

charts/publisher/templates/worker/deployment.yaml
sengi: Consider this all optional btw - I'm happy to fix up the secrets stuff after as it's the next thing I need to work on anyway.

sengi:
Oh and huge thanks for getting this up and running btw — this is gonna make working with this stuff soooo much easier!

PR #62, opened 2022-01-06
Suffix migration job names with chart version and image tag.

sengi:
This avoids the problem where `helm upgrade` or Argo sync fails because helm/Argo tries to patch the annotations/metadata on the existing (immutable) Job resource.

As a bonus, this makes it easier to see which version of an app a given migration job corresponds to.

It'll result in a few no-op job runs, but that's ok. We'll only be running migration jobs when the chart or the app image changes.

We don't need to worry about the possibility of concurrent migration jobs because Rails/ActiveRecord takes care of locking.

Considered moving the DB migration responsibility into Argo Workflow, but that'd be less portable and more complex so this approach is probably better.

Also considered whether we even want to be running these "migrations" (i.e. schema changes) automatically in the first place, but we really don't want to impose new, manual process on established developer workflow without good reason. So we're keeping the automatic migrations for the time being. It's [normal for Rails], apparently 😅

[normal for Rails]: https://www.bbc.co.uk/news/uk-england-norfolk-36082307

[Trello card](https://trello.com/c/h6lGL53t/706)

### DIFF
--- a/charts/publisher/Chart.yaml
+++ b/charts/publisher/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: publisher
 description: A Helm chart for GOV.UK Publisher
 type: application
-version: 0.9.0
+version: 0.9.1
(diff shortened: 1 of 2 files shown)


bilbof (approved):
Great job, thanks for adding in a comment.

sengi:
Cheers for the super quick review!

PR #66, opened 2022-01-20
Increase Frontend replicaCount in integration

theseanything:
Increasing the replicaCount as part of training and user research


fredericfran-gds (approved):
looks good

PR #67, opened 2022-01-20
Add Helm chart for info-frontend

lwilts:


bilbof (commented):
Nice one! Thanks for putting this up. 🎉 

The chart looks good to me. As a final change could you add `- name: info-frontend` under `applications` in `charts/argocd-apps/values-test.yaml`? This will get ArgoCD to deploy your chart to the test cluster. (It would be nice to have a way for ArgoCD to automatically detect new charts, maybe if we had a parent chart structure, or a directory for charts deployed by ArgoCD).

Also, this commit is pretty self explanatory so not necessary for this one - commiters to alphagov repos tend to write verbose commit messages so it's easy to see the context for a change from just the git log (can't find our docs for this, but there's an old styleguide at https://github.com/alphagov/styleguides/blob/master/git.md).

PR #68, opened 2022-01-21
Add default Argo Workflow TTL config

fredericfran-gds:
We need to change the default Argo Workflow TTL so
that resources are not wasted storing Argo workflow
artifacts "indefinitely".

Ref:
1. [trello card](https://trello.com/c/N9ER3ukt/755-cleanup-completed-jobs)


fredericfran-gds:
Moved to https://github.com/alphagov/govuk-infrastructure/pull/517 since configuring Argo Workflow directly and its helm chart supports that.

PR #72, opened 2022-01-31
Adding govuk-rails-app to ArgoCD as the defaut chart.

nsabri1:
govuk-rails-app chart is now been set as the default chart in ArgoCD - This can be overwritten in:
values-{env}.yaml with 
`- chartPath {chart location}`

Updated govuk-rails-app to fix Ingress.

Things to do:
- Secrets 

https://trello.com/c/OL5uMIxV/786-update-helm-chart-repository-to-include-the-new-govuk-rails-app-in-argocd

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -4,32 +4,173 @@ globalHelmValues:
   externalDomainSuffix: eks.test.govuk.digital
 
 applications:
+- name: argo-workflows
+  chartPath: charts/argo-workflows
 - name: cluster-secrets
+  chartPath: charts/cluster-secrets
 - name: govuk-apps-conf
+  chartPath: charts/govuk-apps-conf
 - name: smokey
+  chartPath: charts/smokey
 - name: publisher
-  ingressHostname: publisher.{{ .Values.globalHelmValues.externalDomainSuffix }}
   helmValues:
-    image:
+    appImage:
       repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/publisher
       tag: latest  # To be edited by automation (not yet implemented).
-    redisUrl: redis://shared-redis-govuk.eks.test.govuk-internal.digital
-    replicas: 1
-    worker:
-      replicas: 1
+    dbMigrationEnabled: true
+    workerEnabled: true
+    ingress:
+      enabled: true
+      annotations:
+        alb.ingress.kubernetes.io/load-balancer-name: publisher
+      hosts:
+      - name: publisher.eks.test.govuk.digital
+    replicaCount: 1
+    workerReplicaCount: 1
+    extraEnv:
+      - name: GDS_SSO_OAUTH_ID
+        valueFrom:
+          secretKeyRef:
+            name: signon-app-publisher-eks
+            key: oauth_id
+      - name: GDS_SSO_OAUTH_SECRET
+        valueFrom:
+          secretKeyRef:
+            name: signon-app-publisher-eks
+            key: oauth_secret
+      - name: ASSET_MANAGER_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: "signon-token-publisher-asset-manager"
+            key: bearer_token
+      - name: PUBLISHING_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: "signon-token-publisher-publishing-api"
+            key: bearer_token
+      - name: FACT_CHECK_PASSWORD
+        valueFrom:
+          secretKeyRef:
+            name: publisher
+            key: FACT_CHECK_PASSWORD
+      - name: FACT_CHECK_REPLY_TO_ADDRESS
+        valueFrom:
+          secretKeyRef:
+            name: publisher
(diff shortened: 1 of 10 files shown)


charts/govuk-rails-app/values.yaml
```diff
@@ -11,21 +11,17 @@ externalSecrets:
...
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/target-type: ip
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: "443"
     alb.ingress.kubernetes.io/healthcheck-path: /healthcheck/ready
+    alb.ingress.kubernetes.io/load-balancer-name:
   hosts:
-  - host:
-    paths:
-    - path: /
```
nsabri1, line 27: Will need to add Multiple path handling in a further iteration of the chart

charts/argocd-apps/values.yaml
```diff
@@ -9,6 +9,6 @@ globalHelmValues:
   govukStack: blue  # TODO: eliminate the need for govukStack.
   externalDomainSuffix: ""
   # TODO: delete internalDomainSuffix once Plek handles empty GOVUK_APP_DOMAIN.
-  internalDomainSuffix: apps.svc.cluster.local.
+  internalDomainSuffix: apps.svc.cluster.local
```
nsabri1, line 12: Removed trailing `.`

sengi: The trailing dot serves a useful purpose here - it prevents unnecessary lookups. See e.g. https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html

Ultimately I'm hoping we can remove these suffixes entirely but I need to fix Plek first.

nsabri1: Was looking at some of the logs for apps and they where generating paths using internal suffix with trailing . such as content-store.apps.svc.cluster.local./somepath.
It seemed it was causing problem in finding the path - but will need to do more testing - i can re-add the trailing dot for now and test locally with and without it?

sengi: Oh, if that's appearing in markup (like links in HTML and stuff like that) then that sounds like we might be using the internal suffix someplace where we should be using the external one 🤔

If it's breaking stuff then of course feel free to drop the dot and we'll figure out what's going on later.

charts/argocd-apps/values-test.yaml
```diff
@@ -38,24 +179,78 @@ applications:
...
-    image:
+    appImage:
       repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/frontend
       tag: latest  # To be edited by automation (not yet implemented).
     replicaCount: 1
-    appMemcacheServers:
-    - frontend-memcached.eks.test.govuk-internal.digital
+    extraEnv:
+      - name: SECRET_KEY_BASE
+        value: secret
```
sengi, line 188: I'd stick a TODO on this (and any other occurrences of it) just to make it a bit easier to find.
```suggestion
        value: secret  # TODO: set SECRET_KEY_BASE and make it a secret
```

charts/govuk-rails-app/values.yaml
```diff
@@ -11,21 +11,17 @@ externalSecrets:
...
-  host: ""
-  path: "/"
-  pathType: Prefix
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/target-type: ip
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: "443"
     alb.ingress.kubernetes.io/healthcheck-path: /healthcheck/ready
+    alb.ingress.kubernetes.io/load-balancer-name:
```
sengi, line 20: Probably better to leave load-balancer-name out than set it to empty string, since it'd likely cause problems if anyone were to actually use the example as-is. If you really wanted to keep it for illustrative purposes then you could leave it commented out with an example name set.

(Also a small nit in case it comes up elsewhere: if you need to set one of these values to empty string, it's better to write `foo: ""` rather than just `foo: ` so that it's strictly valid YAML. Helm seems to put up with it but it's a good idea to stick to the spec. Not sure why lint isn't catching it 🤷)
```suggestion
```

charts/govuk-rails-app/values.yaml
sengi: Ah that's my bad for templating the Ingress values 😅 Do we have any cases where we need multiple paths though? I haven't come across any yet.

sengi (approved):
Cool, let's go for it 🚀 

I can see a bunch of env vars that can probably be cleaned up but probably better to get this in first. Thanks for getting all this working!

charts/govuk-rails-app/values.yaml
nsabri1: I have not seen the need as of yet - but if we do need it it will be easily to implement without breaking anything - i will test this when i add Whitehall.

PR #73, opened 2022-02-01
Add individual cluster secrets to correct namespaces

fredericfran-gds:
By default, the cluster-secrets helm chart is installed in the
`apps` namespace and the secrets are created there.

Some secrets, such has the govuk-dex-* ones, are actually destined to
`cluster-services` namespaces.

This PR fixes that defining in which namespace each individual secret
should be located if not in the default `apps` namespace.


bilbof (approved):

sengi (approved):
Good catch, thanks!

PR #74, opened 2022-02-01
Add label to git-dex-argocd external secret

fredericfran-gds:
There is no need to use a template to add a label to git-dex-argocd,
labels and annotations are copied from external-secret to k8s-secret,
see [doc](https://external-secrets.io/api-externalsecret/)


sengi (approved):
Nice!

PR #75, opened 2022-02-01
Add info-frontend to the test cluster

lwilts:
The info-frontend app deployed to my own namespace successfully using a standalone Helm chart. I am testing its deployment using the govuk-rails-app shared chart and ArgoCD.

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -254,3 +254,18 @@ applications:
         value: content-store
 - name: signon-resources
   chartPath: charts/signon-resources
+- name: info-frontend
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/info-frontend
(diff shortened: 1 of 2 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -254,3 +254,20 @@ applications:
...
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/info-frontend
+      tag: latest  # To be edited by automation (not yet implemented).
+    replicaCount: 1
+    extraEnv:
+      - name: SECRET_KEY_BASE
+        value: secret  # TODO: set SECRET_KEY_BASE and make it a secret
+      - name: PORT
+        value: "3000"
```
sengi, line 267: I think you no longer need this since you've already cleaned up the Dockerfile and removed the non-standard port from there.

```suggestion
```

sengi (approved):

PR #76, opened 2022-02-02
Replace DB migrate Jobs during Sync rather than upgrading in-place

bilbof:
Jobs are immutable so our sync policy should use `kubectl replace` rather than `kubectl apply`. This is configured at the individual resource level only on DB migration jobs.

The motivation for this change is that DB migration jobs can fail or need to be re-run, and we'd like to ensure that we can re-sync those objects successfully.

Docs: https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#replace-resource-instead-of-applying-changes


fredericfran-gds (approved):
ok

PR #77, opened 2022-02-02
Fixup: make ArgoCD labels annotations

bilbof:
These were incorrectly placed under labels rather than
annotations.
See docs: https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/


fredericfran-gds (approved):
ok

PR #79, opened 2022-02-03
Adding whitehall-admin, whitehall-frontend and draft-whitehall-frontend

nsabri1:
Will be testing the application deployment and adjusting where required.

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -269,3 +269,57 @@ applications:
         value: https://www.gov.uk
       - name: GOVUK_APP_NAME
         value: info-frontend
+- name: whitehall-admin
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/whitehall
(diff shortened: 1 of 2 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -263,9 +263,59 @@ applications:
...
+    workerEnabled: true
+    ingress:
+      enabled: true
+      annotations:
+        alb.ingress.kubernetes.io/load-balancer-name: whitehall-admin
+      hosts:
+      - name: whitehall-admin.eks.test.govuk.digital
+    replicaCount: 1
+    workerReplicaCount: 1
+    extraEnv:
```
sengi, line 283: Consider using YAML anchors here if there's a block of environment variables which are supposed to be the same for all three instances.

sengi (approved):
Cool - I see the secrets and maybe some details of the env vars are a work in progress but that's totally fine at this point. Better to merge and fix forward since this doesn't exist yet.

The patterns for the ExternalSecrets stuff are pretty much there now (Publisher is the example). I think I'm gonna change to just having one ExternalSecret for Rails SECRET_KEY_BASE though cos it's not gonna be worth having separate ones any time soon and it's easy to change anyway (so let's do the quicker but still ok thing).

PR #82, opened 2022-02-03
Add OAUTH credentials for Grafana/Dex authentication

fredericfran-gds:
OAUTH credentials for Grafana to authenticate with Dex, a federated
OpenID Connect provider.

Since Dex is in `cluster-services` and Grafana in `monitoring`
namespace, there is a need to create 2 external secrets.
They will remain in sync as the source of truth is in AWS Secret
Manager.

Ref:
1. [trello card](https://trello.com/c/Snqwc4Ac/784-deploy-dex-idp-and-configure-github-backend)

### DIFF
--- a/charts/cluster-secrets/templates/dex-grafana.yaml
+++ b/charts/cluster-secrets/templates/dex-grafana.yaml
@@ -1,0 +1,21 @@
+# This secret contains the OAUTH secret which allows Grafana to
+# authenticate with Dex (https://dexidp.io/), a federated
+# OpenID connect provider.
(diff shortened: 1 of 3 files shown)


charts/cluster-secrets/templates/dex-grafana.yaml
```diff
@@ -0,0 +1,18 @@
+# Note: This secret contains the OAUTH secret which allows Argo-Workflows to
```
sengi, line 1: style nit: the "Note:" prefix is redundant.

sengi (approved):
Cool

PR #84, opened 2022-02-03
Use DocumentDB for Content Store db

bilbof:
The self-hosted mongo cluster in EC2 in the test
account seems to be in a bad state. A quick fix here
is to try and use the existing SharedDocumentDb cluster
for content store.


nsabri1 (commented):
LGTM

nsabri1 (approved):

PR #85, opened 2022-02-03
Use DocumentDB for Content Store db

bilbof:
The self-hosted mongo cluster in EC2 in the test account seems to be in a bad state. A quick fix here is to try and use the existing SharedDocumentDb cluster for content store.

I manually created the `govuk/common/content-store-documentdb` secret in the test AWS account's SM - we don't intend to migrate Content Store to DocumentDB as part of this migration - this is merely a quick fix for the test environment.

### DIFF
--- a/charts/argocd-apps/Chart.yaml
+++ b/charts/argocd-apps/Chart.yaml
@@ -2,4 +2,4 @@
 apiVersion: v2
 name: argocd-apps
 description: Installs ArgoCD applications into the cluster-services namespace
-version: 0.5.7
+version: 0.5.8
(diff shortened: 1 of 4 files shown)


sengi (approved):
Go for it. I'm about to put a whole bunch of similar ExternalSecrets into the govuk-apps-conf chart (since these things aren't really cluster secrets - they really belong in the apps' own charts but the shared chart makes that not very feasible, so the next best thing is to put them alongside the shared ConfigMap since they're a very similar kind of thing). So might be worth putting a note on the ExternalSecret here to say that it's very temporary.

bilbof:
Thanks @sengi let's get your PR merged first then I'll move mine over.

PR #86, opened 2022-02-03
Split up AWS Secrets Manager secrets for Publisher.

sengi:
Publisher secrets were being stored in a single AWS Secrets Manager secret. This was a temporary hack to get the app running quickly, but would prevent us from being able to rotate and manage the credentials contained in that all-in-one Secret.

Split the logical secrets up into separate Secrets Manager secrets so that we don't break auditability and automatability.

For now, we're putting the ExternalSecrets definitions in the govuk-apps-conf Helm chart, for want of a better place. This isn't ideal; they really belong in the apps' own Helm charts (behind an `if .Values.enableExternalSecrets`) but this won't be feasible until we refactor our shared `govuk-rails-app` Helm chart into per-app charts with library chart(s) - i.e. shift from inheritance to composition (as @bilbof eruditely put it).

Some of the ExternalSecrets definitions require the use of Go templates in order to generate (for example) database connection strings (see https://external-secrets.io/guides-templating/). The MongoDB connection string here is one example but there will be others.  These templates (which are Go templates but not Helm templates and are processed by the ExternalSecrets operator) are kept in separate files and included verbatim in the Helm templates in order to avoid horrible double-escaping like this:

```dont-try-this-at-home-kids
{{ "{{" }} .username | toString {{ "}}" }}:{{ "{{" }} .password | toString {{ "}}" }}@{{ "{{" }} .hostname | toString {{ "}}" }}
```
🤮🤮🤮

Also a bunch of minor fixes and cleanup; see commits for details.

#### Alternatives considered

Thought about passing the ExternalSecrets definitions into the shared chart as Helm values, but given that ExternalSecrets can contain (non-Helm) Go templates of their own, the number of levels of string templating and YAML embedding involved would be most unwise (see above - it'd be even worse than that 😱).

Generating the ExternalSecrets definitions in a Helm template in `govuk-rails-app` was another option. This would have avoided the very worst of the double-escaping but would have involved implementing a fair bit of logic inside a Helm template and defining a nontrivial structure for passing in the necessary information from values.yaml about how to construct the ExternalSecrets. Still a lot of unwarranted layers of templating and abstraction, i.e. complexity that doesn't seem worth it compared to the chosen approach above.

Also considered using a separate ConfigMap for templating the MongoDB connection string instead of using [ExternalSecrets Operator's templating](https://external-secrets.io/guides-templating/), but concluded that this only moves the templating problem rather than simplifying it, merely resulting in an extra level of indirection.

#### Rollout

I've created all the required Secrets Manager secrets in the test and integration accounts.

I haven't added the Publisher link-checker-api Signon bearer token to the Signon bootstrap config yet.

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -48,6 +48,11 @@ applications:
           secretKeyRef:
             name: signon-token-publisher-publishing-api
             key: bearer_token
+      - name: FACT_CHECK_USERNAME
+        valueFrom:
+          secretKeyRef:
+            name: publisher
+            key: FACT_CHECK_USERNAME
       - name: FACT_CHECK_PASSWORD
         valueFrom:
           secretKeyRef:
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -41,12 +41,12 @@ applications:
       - name: ASSET_MANAGER_BEARER_TOKEN
         valueFrom:
           secretKeyRef:
-            name: "signon-token-publisher-asset-manager"
+            name: signon-token-publisher-asset-manager
             key: bearer_token
       - name: PUBLISHING_API_BEARER_TOKEN
         valueFrom:
           secretKeyRef:
-            name: "signon-token-publisher-publishing-api"
+            name: signon-token-publisher-publishing-api
             key: bearer_token
       - name: FACT_CHECK_PASSWORD
         valueFrom:
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -51,42 +51,42 @@ applications:
       - name: FACT_CHECK_USERNAME
         valueFrom:
           secretKeyRef:
-            name: publisher
+            name: publisher-fact-check-email-account
             key: FACT_CHECK_USERNAME
       - name: FACT_CHECK_PASSWORD
         valueFrom:
           secretKeyRef:
-            name: publisher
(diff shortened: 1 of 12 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -53,31 +53,11 @@ applications:
           secretKeyRef:
             name: publisher
             key: FACT_CHECK_PASSWORD
-      - name: FACT_CHECK_REPLY_TO_ADDRESS
```
bilbof, line 56: Thanks, this will fix those `CreateContainerConfigError` errors we're seeing for publisher:
```
couldn't find key FACT_CHECK_REPLY_TO_ADDRESS in Secret apps/publisher
```

charts/govuk-apps-conf/templates/external-secrets/publisher/fact-check-email-account.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: external-secrets.io/v1alpha1
+kind: ExternalSecret
+metadata:
+  name: publisher-fact-check-email-account
+  labels:
+    {{- include "govuk-apps-conf.labels" . | nindent 4 }}
+  annotations:
+    a8r.io/description: >
+      Username and password for the factcheck Gmail account. See
+      https://github.com/alphagov/publisher/blob/main/docs/fact-checking.md
```
bilbof, line 10: These annotations are great! Thank you

bilbof (approved):
🚀 Thanks for this one

PR #88, opened 2022-02-04
Stop validating chart version number has incremented

bilbof:
This deactivates a check for chart version increments.

We find the chart versioning isn't particularly useful
given we don't install Helm charts using Helm (ArgoCD
uses kubectl apply/replace and only uses Helm for templating).
We also had many git conflicts on the chart version,
which is the reason for this change.

Docs: https://github.com/helm/chart-testing/blob/main/doc/ct_lint.md

### DIFF
--- a/ct.yml
+++ b/ct.yml
@@ -1,5 +1,6 @@
 remote: origin
 target-branch: main
 validate-maintainers: false
+check-version-increment: false
 chart-dirs:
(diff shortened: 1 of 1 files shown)


sengi (approved):
🚀🚀🚀 (🤞that it works! I think it should though.)

PR #97, opened 2022-02-07
Use one ExternalSecret for the Rails secret instead of O(n).

sengi:
Automating generation and rotation of Rails `SECRET_KEY_BASE` isn't a must-have for phase 1 launch, so let's not preemptively create 55-odd boilerplate ExternalSecrets files just because we might someday split up the single AWS secret to which they refer.

This avoids adding a boilerplate file per app, makes no functional change (since they'd all be pointing to the same AWS secret anyway) and doesn't really make it any harder to split up Rails `SECRET_KEY_BASE` later if we decide we really want to.

See #86 for background.

https://trello.com/c/uFDQulJk/796

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -86,7 +86,7 @@ applications:
       - name: SECRET_KEY_BASE
         valueFrom:
           secretKeyRef:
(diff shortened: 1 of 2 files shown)


nsabri1 (approved):
LGTM

PR #99, opened 2022-02-08
Have chart-releaser skip duplicate chart versions.

sengi:
With `CR_SKIP_EXISTING=true`, chart-releaser will skip uploading a chart
tarball where one already exists with the same filename (i.e. same chart name
and chart version). This means the chart-releaser GitHub Action will no longer
fail when we make changes to a chart without updating the version number.

This is essentially a quick bodge to avoid having to automate incrementing the
chart version on merge-to-main for now.

### But why?

We really don't want to have to update chart versions for charts which we're
only deploying via ArgoCD (which doesn't care about version numbers). Having to
change the version number on every change causes merge conflicts and has proved
a frequent source of annoyance even in the absense of merge conflicts, even for
experienced users. This was a recurring theme in the feedback from the 2021-11
user testing.

There's no technical reason why we need to bother with version numbers for the
charts which we're installing with ArgoCD, since Argo just evaluates the chart
using `helm template` and compares the output; it doesn't care about chart
version numbers. We still need to care about version numbers for charts which
we're deploying using the Helm Terraform provider, though.

Unfortunately the chart-releaser GH Action doesn't support filtering on paths;
it just releases every charts that's changed under `charts_dir` and it uses the
Git history rather than the chart version to tell whether a chart has changed.

We could configure separate chart-releaser actions for the Terraform-deployed
charts, but that sounds like a maintenance pain.

Eventually we might want to have all of our charts published in the chart repo,
at which point we could automate bumping the version number.  We'd have to
forego Helm's conventional use of semver and just increment (say) the minor
version number. Doesn't seem worth it for now though.

https://trello.com/c/wCTPia2K/812

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,4 +29,5 @@ jobs:
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.3.0
         env:
+          CR_SKIP_EXISTING: "true"
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"


bilbof (approved):
Nice, good solution.

sengi:
> Nice, good solution.

That's very kind. I'm not sure I'd describe it as "good" myself 😂 (hopefully good enough for now though)

sengi:
1-line bodge, 37 lines of excuse in the description 🤣🙈

sengi:
The thing I really dislike about this (hopefully temporary) solution is that if anyone does try to use `helm install` from the chart repo, they'll find that the chart tarballs in there are outdated and effectively frozen in time as of now. I wanted to find a way to delete them and have chart-releaser not put them back, but I couldn't figure out a way to do that other than forking the chart-releaser action or duplicating the action config many times.

I expect we'll end up automating the minor version bump eventually.

sengi:
Seems to have done the trick: https://github.com/alphagov/govuk-helm-charts/runs/5108997395?check_suite_focus=true

bodger and badger

PR #101, opened 2022-02-08
Adding whitehall-admin whitehall-frontend and draft-whitehall-frontend

nsabri1:
This Change is a work in progress.

From the initial looks on EC2 hosting Whitehall, all three contain the same ENV values,
therefore i am using YAML anchor,

Ultimately ENV values may differ, i will then update the chart.

Env values which i have skipped for now:
- GOVUK_CSP_REPORT_ONLY
- GOVUK_CSP_REPORT_URI
- GOVUK_DATA_SYNC_PERIOD

I have excluded a couple of secrets which i will add later.
secrets skipped:
- EMAIL_ALERT_API_BEARER_TOKEN
- RUMMAGER_BEARER_TOKEN

https://trello.com/c/lIyh6zBn/753-work-out-which-apps-can-be-merged-using-govuk-rails-app-helm-chart

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -295,3 +295,100 @@ applications:
         value: secret  # TODO: set SECRET_KEY_BASE and make it a secret
       - name: GOVUK_APP_NAME
         value: collections
+- name: whitehall-admin
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/whitehall
+      tag: f01a5161aa0d93b2efdd0775c61081695f12447a  # TODO: To be edited by automation (not yet implemented).
+    dbMigrationEnabled: true
+    workerEnabled: true
+    ingress:
+      enabled: true
+      annotations:
+        alb.ingress.kubernetes.io/load-balancer-name: whitehall-admin
+      hosts:
+      - name: whitehall-admin.eks.test.govuk.digital
+    replicaCount: 1
+    workerReplicaCount: 1
+    extraEnv: &whitehall-envs
+      - name: GOVUK_APP_NAME
+        value: whitehall
+      - name: GOVUK_APP_TYPE
+        value: rack
+      - name: NODE_ENV
+        value: production
+      - name: RACK_ENV
+        value: production
+      - name: RAILS_ENV
+        value: production
+      - name: GDS_SSO_OAUTH_ID
+        value: eiGhei3oop2seGh4Phee8eikuphai9coh5moJ2ohTh8soghohdohquashaor0joh
+      - name: GDS_SSO_OAUTH_SECRET
+        valueFrom:
+          secretKeyRef:
+            name: signon-app-whitehall-eks
+            key: oauth_secret
+      - name: ASSET_MANAGER_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-whitehall-asset-manager
+            key: bearer_token
+      - name: GOVUK_NOTIFY_API_KEY
+        valueFrom:
+          secretKeyRef:
+            name: publisher-notify
+            key: GOVUK_NOTIFY_API_KEY
+      - name: JWT_AUTH_SECRET
+        valueFrom:
+          secretKeyRef:
+            name: authenticating-proxy-jwt-auth-secret
+            key: JWT_AUTH_SECRET
+      - name: LINK_CHECKER_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-whitehall-link-checker-api
+            key: bearer_token
+      - name: LINK_CHECKER_API_SECRET_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: publisher-link-checker-api-callback-token
+            key: LINK_CHECKER_API_SECRET_TOKEN
+      - name: PUBLISHING_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-whitehall-publishing-api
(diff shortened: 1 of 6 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -295,3 +295,100 @@ applications:
         value: secret  # TODO: set SECRET_KEY_BASE and make it a secret
       - name: GOVUK_APP_NAME
         value: collections
+- name: whitehall-admin
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/whitehall
+      tag: f01a5161aa0d93b2efdd0775c61081695f12447a  # To be edited by automation (not yet implemented).
```
fredericfran-gds, line 302: I think the format is `# TODO: <message>` for automation/search to find it easier

charts/argocd-apps/values-test.yaml
```diff
@@ -295,3 +295,100 @@ applications:
...
+        value: whitehall
+      - name: GOVUK_APP_TYPE
+        value: rack
+      - name: NODE_ENV
+        value: production
+      - name: RACK_ENV
+        value: production
+      - name: RAILS_ENV
+        value: production
+      - name: GDS_SSO_OAUTH_ID
```
fredericfran-gds, line 324: we should create a single secret for GDS_SSO_OAUTH with format:
```json
{
clientID: <secret>,
clientSecret: <secret>
}
```

nsabri1: i can potentially look at this as  separate task

charts/argocd-apps/values-test.yaml
nsabri1: That right, thanks 
i will update this with an amend

charts/argocd-apps/values-test.yaml
```diff
@@ -295,3 +295,100 @@ applications:
...
+            key: bearer_token
+      - name: GOVUK_NOTIFY_TEMPLATE_ID
+        value: 759acac6-da53-4a19-b591-b7538c7c39de
+      - name: GOVUK_STATSD_PREFIX
+        value: whitehall
+      - name: REDIS_URL
+        value: redis://shared-redis-govuk.eks.test.govuk-internal.digital
+      - name: MEMCACHE_SERVERS
+        value: frontend-memcached.eks.test.govuk-internal.digital
+      - name: DATABASE_URL
```
fredericfran-gds, line 369: how was this secret store? was it in AWS Secrets Manager and done in the format that AWS SM expects to be able to rotate it automatically? see: https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_available-rotation-templates.html#sar-template-mysql-singleuser

charts/argocd-apps/values-test.yaml
```diff
@@ -295,3 +295,100 @@ applications:
...
+      - name: SECRET_KEY_BASE
+        valueFrom:
+          secretKeyRef:
+            name: rails-secret-key-base
+            key: SECRET_KEY_BASE
+- name: whitehall-frontend
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/whitehall
+      tag: f01a5161aa0d93b2efdd0775c61081695f12447a  # To be edited by automation (not yet implemented).
```
fredericfran-gds, line 383: Same as above.

charts/argocd-apps/values-test.yaml
```diff
@@ -295,3 +295,100 @@ applications:
...
+    replicaCount: 1
+    extraEnv:
+      *whitehall-envs
+- name: draft-whitehall-frontend
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/whitehall
+      tag: f01a5161aa0d93b2efdd0775c61081695f12447a  # To be edited by automation (not yet implemented).
```
fredericfran-gds, line 391: same as above.

charts/argocd-apps/values-test.yaml
fredericfran-gds: It seems so from below, how external secrets are created

nsabri1: I will be adding it in SM manually - there will be no rotation for now - i will be adding the current password and user used in old world.

nsabri1: can look into rotation as work that stems of from this initial test of Whitehall

fredericfran-gds (approved):
for me, it is fine to give it a try...

charts/argocd-apps/values-test.yaml
sengi: There should already be such a secret because these are generated by Bill's Signon bootstrap rake task.

You should be able to just refer to it like this:

```
      - name: GDS_SSO_OAUTH_ID
        valueFrom:
          secretKeyRef:
            name: signon-app-whitehall-eks
            key: oauth_id
```

If you've created any of these GDS_SSO_* secrets manually, you'll want to delete them and then run Bill's Signon bootstrap job. It looks like it's already [configured](https://github.com/alphagov/govuk-helm-charts/blob/main/charts/signon-resources/templates/bootstrap-job.yaml) to create some Whitehall stuff.

charts/argocd-apps/values-test.yaml
sengi: No need to worry about actually implementing rotation just yet, but when you create the AWS secret please try to use the built-in RDS secret format if possible (when it asks you what kind of secret you're creating, in the web UI).

sengi: Oh, same as what Fred already said 😅 (sorry Fred!)

PR #102, opened 2022-02-08
Add secret description to annotation of k8s object

fredericfran-gds:
The secret description are added as an annotation of the k8s externalsecrets object.

see pr #86 for further details.


sengi (approved):
Sweet - thanks for doing this cleanup!

PR #103, opened 2022-02-09
Adding government-frontend to helm chart

alexbakervr:
Adding government-frontend to the default helm chart.

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -303,7 +303,10 @@ applications:
     replicaCount: 1
     extraEnv:
       - name: SECRET_KEY_BASE
-        value: secret  # TODO: set SECRET_KEY_BASE and make it a secret
+        valueFrom:
+          secretKeyRef:
+            name: rails-secret-key-base
+            key: SECRET_KEY_BASE
       - name: GOVUK_APP_DOMAIN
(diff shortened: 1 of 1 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -295,3 +295,18 @@ applications:
...
         value: collections
+- name: government-frontend
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/government-frontend
+      tag: latest  # To be edited by automation (not yet implemented).
+    replicaCount: 1
+    extraEnv:
+      - name: SECRET_KEY_BASE
+        value: secret  # TODO: set SECRET_KEY_BASE and make it a secret
```
sengi, line 306: I've set up an external secret for SECRET_KEY_BASE now so you can fix this TODO like so:
```suggestion
      - name: SECRET_KEY_BASE
        valueFrom:
          secretKeyRef:
            name: rails-secret-key-base
            key: SECRET_KEY_BASE
```

sengi (approved):
👍

charts/argocd-apps/values-test.yaml
alexbakervr: Thanks Chris, will take a look at this now.

PR #106, opened 2022-02-09
Removing External secret file from Rails Helm chart

nsabri1:
External secrets does not need to be included in the default Rails Helm chart.

External secrets for charts will need to be added into govuk-apps-conf Helm chart.


sengi (approved):
Thanks for cleaning this up!

PR #107, opened 2022-02-10
Just a small tidy of the values within test:

nsabri1:
removed ENV values that are already added via govuk-apps-conf
        added secret_key_base
        fixed memcache dns name


sengi (approved):
Thanks!

PR #115, opened 2022-02-11
Set resource limits on pub-api

bilbof:
We're seeing Publishing API containers exit with code 137 so want to set a limit so we can provide that it truly is OOMing.


fredericfran-gds (approved):
ok

PR #119, opened 2022-02-11
Allow overriding container probes

bilbof:
When you provide two values.yaml files to helm template
the blocks are *merged* so that two blocks with the same
keys have the union of both sets of keys/values.

This meant that it wasn't possible to set a non-httpGet livenessProbe
as the httpGet block in the default liveness probe would always
be included in any livenessProbe.


sengi (approved):

sengi:
Thanks! Fixes my screwup in #114

PR #126, opened 2022-02-11
Add steps to bootstrapping docs

bilbof:
We're turning up the integration environment and finding manual steps that have crept in - documenting them for staging turnup and the future automation of these steps.


sengi (approved):
Thanks!

PR #127, opened 2022-02-11
Add PORT envvar to all app pods

bilbof:
We set the exposed containerPort to `.Values.appImage.appPort` so this ensures that we set the PORT envvar (used by our Rails apps pretty consistently) to the same value.

Currently both government-frontend and finder-frontend are failing their healthchecks as their [Dockerfiles] specify PORT=3090 so the apps are not running on the exposed port.

[Dockerfiles]: https://github.com/alphagov/government-frontend/blob/main/Dockerfile#L32

### DIFF
--- a/charts/govuk-rails-app/templates/deployment.yaml
+++ b/charts/govuk-rails-app/templates/deployment.yaml
@@ -25,6 +25,8 @@ spec:
           - configMapRef:
               name: govuk-apps-env
           env:
+          - name: PORT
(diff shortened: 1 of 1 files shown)


sengi (approved):
I think this is a good idea if only as a temporary measure to make it easier to get things going for now.

The dockerfiles that have these weird port numbers hardcoded are all ones which will need fixing up anyway soon enough though.

PR #128, opened 2022-02-11
Add ExternalSecret for Publishing API RDS DB

bilbof:
Integration uses the RDS hosted database rather than the EC2 hosted postgres instance. I've manually added a Secret `govuk/publishing-api/postgres` to AWS SecretsManager for this ExternalSecret.

This should get the ArgoCD apps `govuk-apps-conf` and `publishing-api` (I hope) green.


sengi (approved):
Thanks!

PR #129, opened 2022-02-11
Fixup govuk-rails-app env var YAML

bilbof:
This fixes an indentation issue which was caught at sync-time.


fredericfran-gds (approved):
ok

PR #142, opened 2022-02-18
Fix healthchecks for draft-router and configure all backends.

sengi:
Healthchecks for draft-router were failing because the control port was set inconsistently (an oversight in #114).

Also configure all the remaining backends for router and draft-router. I've checked these against the contents of the `router.backends` and `draft_router.backends` collections from the production `router_backend` MongoDB cluster. Services which no longer resolve or have been decommissioned have been omitted.

Licensify isn't being moved at the same time as the other frontends, hence it points at the Licensify service in EC2.

Signon isn't a backend for Router, so remove that.

### DIFF
--- a/charts/argocd-apps/values-test.yaml
+++ b/charts/argocd-apps/values-test.yaml
@@ -289,14 +289,42 @@ applications:
           router-backend-1.pink.test.govuk-internal.digital,\
           router-backend-2.pink.test.govuk-internal.digital,\
           router-backend-3.pink.test.govuk-internal.digital"
+      - name: BACKEND_URL_collections
+        value: "http://collections"
       - name: BACKEND_URL_content-store
         value: "http://content-store"
+      - name: BACKEND_URL_email-alert-frontend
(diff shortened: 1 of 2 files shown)


charts/argocd-apps/values-test.yaml
```diff
@@ -289,14 +289,42 @@ applications:
...
+      - name: BACKEND_URL_feedback
+        value: "http://feedback"
+      - name: BACKEND_URL_finder-frontend
+        value: "http://finder-frontend"
+      - name: BACKEND_URL_info-frontend
+        value: "http://info-frontend"
+      - name: BACKEND_URL_licencefinder
+        value: "http://licencefinder"
+      - name: BACKEND_URL_licensify  # There is no Licensify in test.
+        value: "https://licensify.integration.publishing.service.gov.uk"
```
bilbof, line 323:
```suggestion
        value: "https://licensify.test.publishing.service.gov.uk"
```

PR #144, opened 2022-02-18
Exclude test-only ec2-database-url ExternalSecret in other envs.

sengi:
This was making govuk-apps-conf look unhealthy in the ArgoCD UI.

Tested: `helm template .` and `helm template . --set govukEnvironment=integration`


nsabri1 (approved):
LGTM

PR #146, opened 2022-02-21
EKS Signon app registration.

nsabri1:
Regsiter our EKS apps.

  - Updated name off EKS app.
  - Updated URI's.


fredericfran-gds (changes requested):
Thanks for doing that. can we add eks in the key of the map too to keep consistency with publisher?

fredericfran-gds (approved):
ok cool

sengi:
Cool, thanks for cleaning these up!

PR #148, opened 2022-02-21
Fixup send-slack template format

bilbof:
To map to the format in https://raw.githubusercontent.com/argoproj-labs/argocd-notifications/release-1.0/catalog/install.yaml and the other template in this file.

This should resolve the error we've been seeing in the logs for `deploy/argocd-notifications-controller`:

```
Failed to parse new settings: failed to unmarshal template send-slack: error unmarshaling JSON: json: cannot unmarshal string into Go struct field Notification.webhook of type services.WebhookNotifications
```

https://trello.com/c/iw709yOo/820-set-up-slack-notifications-for-integration-argocd


fredericfran-gds (approved):
cool, thanks

PR #150, opened 2022-02-21
Don't nest appPort and nginxPort values inside appImage and nginxImage.

sengi:
The port numbers are unrelated to the image spec.

There are no other references to `appImage` or `nginxImage` to update.

Tested: `helm template . --set appImage.repository=foo/bar`


fredericfran-gds (approved):
looks good

PR #154, opened 2022-02-23
Remove obsolete Helm chart

fredericfran-gds:
With the introduction of the `govuk-rails-app` Helm chart, many apps
use that Helm chart rather than their own. These PR cleans the obsolete
Helm charts to avoid confusion.


sengi (approved):
Thanks!

PR #155, opened 2022-02-23
Fix typo in gov-frontend backend url for router

bilbof:
The backend in the Routes table is government-frontend rather than government_frontend, so this will get smoke tests for government-frontend passing.

A full list of backends pulled from the Integration routes table is available at https://trello.com/c/qtAZfkjg/267-make-router-work-with-multiple-domains-and-a-single-database#comment-6033d3120e7785455f222f99


sengi (approved):
Good catch, thanks! Looks like I messed up in #142 and botched one of them instead of copy/pasting 🤦

PR #156, opened 2022-02-23
Use EC2-hosted APIs

bilbof:
The frontend applications will send requests to Content Store, Search API, Router API (etc.) hosted in EC2, rather than instances of these services in Kubernetes.

This will make it easier for us to get frontend apps serving traffic sooner as we'll need a smaller subset of prod-ready helm charts.

This is needed to get smoke tests passing for applications that rely on Search API, since this service is not yet available at `http://search.apps` in-cluster.


fredericfran-gds (approved):
can we use `ec2` instead of `legacy`? ;-)

bilbof:
@fredericfran-gds have changed - sorry, EC2 is more appropriate.

PR #159, opened 2022-02-23
Fixup search plek env var

bilbof:
It's search_uri not search_api_uri


marcpomfret (approved):
LGTM

sengi:
So confusing 😅 I guess this is the legacy of the `rummager` -> `search` -> `search-api` renaming journey 🙈

Even more confusingly, there is a Router backend called `search` but it's [obsolete](git push --set-upstream origin sengi/defunct-backend), whereas the real Search backend in Router is called `search-api` 🙃

PR #160, opened 2022-02-23
Remove long-obsolete `search` backend from Router.

sengi:
I didn't spot in 77d4dc7 that the `search` backend is disused. It points to a Carrenza VIP which no longer exists (double-checked in staging and in prod).

The `search-api` backend is still in use and remains in the config.


fredericfran-gds (approved):
looks good

PR #163, opened 2022-02-23
Added label to post sync for app identification

marcpomfret:


bilbof (approved):
LGTM let's see if it works!

PR #164, opened 2022-02-23
Remove onceper option from on-deployed trigger

marcpomfret:


bilbof (approved):
Thanks. Want to note that this is just a temporary change to enable to us move a bit quicker with testing our workflows.

PR #165, opened 2022-02-23
Fixup signon app slugs

bilbof:
This should get the signon-resources job passing. It looks like the slugs were erroneously renamed.

We're currently still deciding how we'll migrate Signon Applications with minimal disruption to end users. For now we've decided to create a separate Application for the publishing apps (Whitehall Admin, Publisher) while we consider how we'll resolve this problem (probably we'll use env var overrides for the REDIRECT_URIs for Applications).

We'll come back to this after we've hit the frontend serving milestone.


fredericfran-gds (approved):
looks good

PR #167, opened 2022-02-23
Retry failed post-sync workflows on transient error

bilbof:
We've seen smokey is a little flakey and can fail when the Java proxy it uses for Selenium tests (BrowserMob) fails to start. It's unclear why this proxy is needed.

We can go further and specify a transient error pattern, but I think we should avoid this since we want to use the argo workflows component for more than just smoke testing. E.g. https://github.com/argoproj/argo-workflows/blob/master/util/errors/errors.go#L37

Not 100% sure that the retryStrategy will work on the WorkflowTemplate rather than the whole Workflow (docs aren't super useful https://argoproj.github.io/argo-workflows/retries/).

### DIFF
--- a/charts/argo-workflows/templates/post-sync-workflow.yaml
+++ b/charts/argo-workflows/templates/post-sync-workflow.yaml
@@ -5,6 +5,9 @@ metadata:
 spec:
   entrypoint: handle-sync-event
   onExit: exit-handler
+  retryStrategy:
+    limit: "3"
+    expression: "asInt(lastRetry.exitCode) > 1"
   arguments:
     parameters:
       - name: application


charts/argo-workflows/templates/post-sync-workflow.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
 spec:
   entrypoint: handle-sync-event
   onExit: exit-handler
+  retryStrategy:
+    limit: "3"
+    expression: "asInt(lastRetry.exitCode) > 1"
```
sengi, line 10: The `> 1` is a little surprising. Does the job return 2 or greater when it fails?

sengi (approved):

charts/argo-workflows/templates/post-sync-workflow.yaml
bilbof: Yes it's surprising, not sure why but it used exit code 2:

```
STEP                       TEMPLATE                   PODNAME                     DURATION  MESSAGE
 ✖ post-sync-nb66d         handle-sync-event                                                                     
 └─✖ smoke-test            smoke-test/smoke-test      post-sync-nb66d-402463690   38m       Error (exit code 2)  
```

bilbof: Output of `kubectl -napps describe pod post-sync-nb66d-402463690` below.

```
    State:          Terminated
      Reason:       Error
      Exit Code:    2
```

richardTowers:
> It's unclear why this proxy is needed.

The selenium webdriver thing we use doesn't let you fiddle with HTTP headers - that's what the proxy is for. I'm not sure what our exact use case is for fiddling with HTTP headers - maybe some Cache-Control shenanigans? Or HTTP Basic Auth?

Annoying that it's flakey. Retrying seems like a good pragmatic solution.

bilbof:
Thanks @richardTowers we should probably look at improving it in future.

PR #169, opened 2022-02-24
Test configmap causes update of pods

bilbof:
The intent of this change is to check whether pods which use env vars provided by this ConfigMap are updated when the configmap is modified.


sengi (approved):
I'm assuming this is a manual test of the [checksum/config annotation trick](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments)?

PR #173, opened 2022-02-24
Increase Memory limit and request for ClamAV

nsabri1:
This change is required to test scanning - as currently OOM is triggered on file scan.

I will be fine tuning MEM request and limits further as i test.


fredericfran-gds (approved):
ok

PR #177, opened 2022-02-24
Use $http_host for Nginx Host header

bilbof:
host and http_host are subtly different
see https://stackoverflow.com/a/67226842

We use http_host in EC2 so this is an attempt
to preserve some prior behaviour.

Rails apps generate urls using the HOST header
so we want to ensure that the HOST is set to
the original host provided at the edge, not
the internal backend host used by Router.


sengi (approved):
LGTM - though it looks like you've omitted `http_host` from the commit title and PR description 😅

bilbof:
@sengi haha! Thanks, fixed, that's my use of `git commit -m` catching me out 😮‍💨

PR #179, opened 2022-02-25
Add specific nginx configuration for router

samsimpson1:
Trello cards:
https://trello.com/c/Lkb8N313/381-epic-nginx-and-varnish-changes-needed
https://trello.com/c/jRmN4yoh/805-robotstxt-page

This adds a separate nginx config specifically for router. This is needed to implement a lot of functionality that nginx and varnish currently provide in the old EC2-based environments. This also allows for specifying individual nginx config files for other applications in the future.

### DIFF
--- a/charts/govuk-apps-conf/templates/router-configmap.yaml
+++ b/charts/govuk-apps-conf/templates/router-configmap.yaml
@@ -1,0 +1,104 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: router-nginx-conf
+  labels:
+    {{- include "govuk-apps-conf.labels" . | nindent 4 }}
+    app.kubernetes.io/component: web
+data:
+  nginx.conf: |-
+    user nginx;
+
+    load_module /usr/lib/nginx/modules/ngx_http_perl_module.so;
+
+    error_log  /var/log/nginx/error.log warn;
+    pid        /var/run/nginx.pid;
+
+    events {
+      worker_connections  1024;
+    }
+
+    http {
+      include       /etc/nginx/mime.types;
+      default_type  application/octet-stream;
+
+      server_tokens off;
+
+      sendfile        on;
+      keepalive_timeout  65;
+
+      perl_set $uri_lowercase 'sub {
+        my $r = shift;
+        return lc($r->uri);
+      }';
+
+      upstream router {
+        server 127.0.0.1:3000;
+      }
+
+      server {
+        listen 8080;
+
+        location / {
+          proxy_set_header   Host $http_host;
+          proxy_set_header   X-Real-IP $remote_addr;
+          proxy_pass         http://router;
+          proxy_redirect     off;
+        }
+        location /assets {
+          proxy_set_header   Authorization "";
+          proxy_set_header   Connection "";
+          proxy_set_header   X-Real-IP $remote_addr;  # TODO: pass the actual end-client address
+          proxy_hide_header  x-amz-id-2;
+          proxy_hide_header  x-amz-meta-server-side-encryption;
+          proxy_hide_header  x-amz-request-id;
+          proxy_hide_header  x-amz-server-side-encryption;
+          proxy_hide_header  x-amz-version-id;
+          add_header         Cache-Control max-age=31536000;
+          proxy_intercept_errors on;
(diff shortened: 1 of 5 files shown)


fredericfran-gds (commented):
looks promising, do you want to add some description in the PR/commit as recommended here: https://gds-way.cloudapps.digital/standards/pull-requests.html#opening-a-pull-request

charts/govuk-rails-app/templates/nginx-configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.nginxConfigMap.create }}
```
bilbof, line 1: Currently when an app needs an Nginx config it can use a custom one by specifying `.Values.nginxConfigMap.name: my-custom-configmap`, or it can use the one built into the `govuk-rails-app` chart by setting `.Values.nginxConfigMap.create: true` (the default).

Perhaps we should move this configmap to the `govuk-apps-conf` Helm chart, so we have a consistent pattern for providing Nginx configs to apps. What do you think?

charts/govuk-apps-conf/templates/router-configmap.yaml
```diff
@@ -0,0 +1,104 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: router-nginx-conf
+  labels:
+    {{- include "govuk-apps-conf.labels" . | nindent 4 }}
+    app.kubernetes.io/component: web
+data:
+  nginx.conf: |-
```
bilbof, line 9: It would be good to document (perhaps in your commit message) why this config was chosen. With config like this it's often hard to figure out if there was an important reason particular config was added, or if it is safe to remove. 

Documenting the intent behind Nginx config will make it much easier to avoid 'teleological' config where we explain it by how it behaves rather than how it is designed to behave.

For instance, were these config options (like lowercasing urls) taken from the Puppet config, if so maybe we can port over the reasons given there?

charts/govuk-apps-conf/templates/router-configmap.yaml
```diff
@@ -0,0 +1,104 @@
...
+        location ~ ^\/[A-Z]+[A-Z\W\d]+$ {
+          rewrite ^(.*)$ $scheme://$host$uri_lowercase permanent;
+        }
+        location = /robots.txt {
+          root /usr/share/nginx/html;
+        }
+      }
+    }
+  robots.txt: |-
+    {{- if eq .Values.govukEnvironment "staging" }}
```
bilbof, line 73: I would think we could have Production as the base case and all other cases (non-live environments like staging, integration, test, etc) disallow all robots.

bilbof (approved):
Looks great! Thanks for doing this. I have a few comments in line which are not blockers to merge, but would be good to fix at some point.

There is more Nginx config to port across from [EC2 Router Puppet config](https://github.com/alphagov/govuk-puppet/blob/main/modules/router/templates/router_include.conf.erb#L6) which we can do later if it's not needed to complete this bit of work.

charts/govuk-rails-app/templates/nginx-configmap.yaml
samsimpson1: We might run in to issues if we move the config map into govuk-apps-conf, since we lose the ability to template in values for that specific application (things like app port would have to be static). If we don't need that, it would probably make sense to move it

bilbof: You're right, let's leave it here - thanks!

sengi: I don't think it's possible to move this configmap (the one that's being created for most apps by govuk-rails-conf chart at the moment) into govuk-apps-conf chart. Each app needs its own nginx configmap. If you meant giving every app its own separate configmap in govuk-apps-conf chart, then I don't think that buys us anything and just adds complexity.

sengi (approved):
LGTM - thanks for fixing this up.

sengi:
Also +1 to all Bill's comments

charts/govuk-rails-app/templates/nginx-configmap.yaml
bilbof: Yep makes sense @sengi, I didn't realise that each app needed a unique nginx configmap.

PR #175, opened 2022-02-24
Add Email-alert-frontend

fredericfran-gds:
Various commits to add email-alert-frontend to the EKS platform.

Ref:
1. [trello card](https://trello.com/c/0gdHdyUe/833-package-email-alert-frontend-for-migration)

PR #181, opened 2022-02-25
Fix env var value for email-alert-frontend

fredericfran-gds:
In #175, `yes` was used without quotes so was interpreted by yaml as
bool.


sengi (approved):
TIL 😅

sengi:
Turns out there are a whole bunch of these 🙃 https://yaml.org/type/bool.html (why, YAML, whyyy? 😅)

PR #182, opened 2022-02-25
Add a PreSync job to the govuk-rails-app chart

lwilts:
This uploads Rails assets to S3, so that they can be served from there.

### DIFF
--- a/charts/govuk-rails-app/templates/assets-upload-job.yaml
+++ b/charts/govuk-rails-app/templates/assets-upload-job.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.uploadAssets.enabled -}}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -21,3 +22,4 @@ spec:
           aws s3 sync /app/public/assets/{{ .Release.Name }} \
             s3://govuk-app-assets-{{ .Values.govukEnvironment }}/assets/{{ .Release.Name }}/
       restartPolicy: Never
+{{- end }}
--- a/charts/govuk-rails-app/templates/assets-upload-job.yaml
+++ b/charts/govuk-rails-app/templates/assets-upload-job.yaml
@@ -1,0 +1,23 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-uploadRailsAssets
+  annotations:
+    argocd.argoproj.io/hook: PreSync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
+spec:
+  template:
+    spec:
+      containers:
+      - name: {{ .Release.Name }}-uploadRailsAssets
+        image: "{{ required "Valid .Values.appImage.repository required!" .Values.appImage.repository }}:{{ required "Valid .Values.appImage.tag required!" .Values.appImage.tag }}"
+        command:
+        - "bash"
+        - "-c"
+        - |
+          [ -d "/app/public/assets/{{ .Release.Name }}" ] && \
+          apt-get update && \
+          apt-get install -y awscli && \
+          aws s3 sync /app/public/assets/{{ .Release.Name }} \
+            s3://govuk-app-assets-{{ .Values.govukEnvironment }}/assets/{{ .Release.Name }}/
+      restartPolicy: Never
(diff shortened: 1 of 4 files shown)


charts/govuk-rails-app/templates/assets-upload-job.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: batch/v1
```
bilbof, line 1: Would be good to wrap this in a `if` statement and disable it for apps that don't have assets at `/app/public/assets`.

Though this helm chart is called `govuk-rails-app` it's used by non-Rails Ruby apps and Router (Golang).

charts/govuk-rails-app/templates/assets-upload-job.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: {{ .Release.Name }}-uploadRailsAssets
+  annotations:
+    argocd.argoproj.io/hook: PreSync
+    argocd.argoproj.io/hook-delete-policy: HookSucceeded
+spec:
+  template:
+    spec:
+      containers:
+      - name: {{ .Release.Name }}-uploadRailsAssets
+        image: "{{ required "A valid .Values.appImage.repository entry required!" .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
```
bilbof, line 13: We have a warning for the repo but not the image, is it intentional?

charts/govuk-rails-app/templates/assets-upload-job.yaml
lwilts: I wasn't aware of that - talking it through with Chris now to identify things like that, so thanks!

charts/govuk-rails-app/templates/assets-upload-job.yaml
```diff
@@ -0,0 +1,22 @@
...
+        command:
+        - "bash"
+        - "-c"
+        - |
+          apt-get update && \
+          apt-get install -y awscli && \
```
bilbof, line 19: Would there be a way to use a pre-built image (perhaps the [official awscli image](https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-docker.html)) rather than using `apt-get install` during the deploy?

lwilts: This is the tricky part of doing the upload on deploy - it needs to get the assets from the built app image. Open to better suggestions

charts/govuk-rails-app/templates/assets-upload-job.yaml
lwilts: Not particularly, I lifted this from the dbmigrate job

charts/govuk-rails-app/templates/assets-upload-job.yaml
bilbof: Yes tricky one! I'm not entirely against adding the AWS CLI to the images, but if we need to do it, we should do it in the Dockerfile.

Did you already consider doing the upload at build time instead? We push artefacts to ECR from GitHub Actions, we could add a build step to a GitHub Action to push the assets up. I don't think this necessarily means we'd need to install `awscli` into the image, there might be a way to add a second step to the GitHub Action which reads from the output of the first step 🤔

Alternatively, we could do something with an init container in Kubernetes 🤔 

We built this previously into our deploy pipeline using Concourse:
https://github.com/alphagov/govuk-infrastructure/blob/ecs/concourse/pipelines/deploy.yml#L829-L837

The way this worked was we'd pass the assets directory as an output from the app image, and then use a `cli` image to push them up to S3.

bilbof: Sounds like a good plan. Thanks @sengi, I'm happy for us to merge it as-is to get it working 😃

charts/govuk-rails-app/templates/assets-upload-job.yaml
sengi: done!

sengi (approved):

bilbof (approved):

charts/govuk-rails-app/templates/assets-upload-job.yaml
sengi: We also made the script (hopefully) a harmless no-op when `/app/public/assets/$name_of_app` doesn't exist, so it shouldn't break stuff when someone adds an app that doesn't have an assets directory but doesn't disable the assets upload.

PR #186, opened 2022-02-25
Disallow all robots except in production.

sengi:
Invert the conditional so that we serve the disallow-all robots.txt in all non-production environments. We wouldn't want crawlers to index a non-production site if it were to be accidentally made public.

Also move the prod robots.txt into a separate file to make it easier to read and maintain. Since the file doesn't vary by environment, include it verbatim using `$.Files.Get` to further simplify matters. 

Tested: inspected output of `helm template .` and `helm template . --set govukEnvironment=production`.


bilbof (approved):
LGTM

PR #188, opened 2022-02-28
Smartanswers

fredericfran-gds:
This PR adds the `smart-answers` frontend application to the EKS platform.

Note that for integration, the Zendesk password in EC2 is empty and hence we keep it this way for EKS Integration.

Ref:
1. [trello card](https://trello.com/c/vR41bV93/840-package-smart-answers-for-migration)

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -780,3 +780,37 @@ applications:
           secretKeyRef:
             name: email-alert-auth
             key: token
+- name: smart-answers
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/smart-answers
+      tag: latest  # To be edited by automation (not yet implemented).
+    replicaCount: 1
+    extraEnv:
+      - name: EXPOSE_GOVSPEAK
+        value: 1
+      - name: COMPANIES_HOUSE_API_KEY
+        valueFrom:
+          secretKeyRef:
+            name: smart-answers-companies-house
+            key: api_key
+      - name: LINK_CHECKER_API_BEARER_TOKEN
+        valueFrom:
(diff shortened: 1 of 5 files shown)


charts/argocd-apps/values-integration.yaml
```diff
@@ -780,3 +780,37 @@ applications:
...
+      - name: ZENDESK_CLIENT_USERNAME
+        valueFrom:
+          secretKeyRef:
+            name: smart-answers-zendesk-client
+            key: username
+      - name: ZENDESK_CLIENT_PASSWORD
+        valueFrom:
+          secretKeyRef:
+            name: smart-answers-zendesk-client
+            key: password
```
sengi, line 816: Good use of Secrets Manager here - thanks!

sengi (approved):
Cool!

charts/argocd-apps/values-test.yaml
```diff
@@ -745,3 +745,37 @@ applications:
...
+- name: smart-answers
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/smart-answers
+      tag: latest  # To be edited by automation (not yet implemented).
+    replicaCount: 1
+    extraEnv:
+      - name: EXPOSE_GOVSPEAK
+        value: 1
+      - name: COMPANIES_HOUSE_API_KEY
```
theseanything, line 757: Just double checked and realised Smart Answers no longer uses this (we never implemented the functionality that used the API). I've raised a [PR to remove it from puppet](https://github.com/alphagov/govuk-puppet/pull/11564) - but it'd also be safe to remove from here too.

sengi: Oh haha, thanks! Fewer secrets is always good :D

PR #190, opened 2022-03-01
Require Basic auth on `www-origin` in test and integration environments.

sengi:
- Add an ExternalSecret to fetch the htpasswd contents from AWS Secrets Manager.
- Allow passing in extra volumes for the nginx pod in the `govuk-rails-app` chart.
- Mount the `htpasswd` file using Secret and VolumeMount.

Plus some minor cleanup while I'm there:

- Rename `nginxConfigMap.extraVolumeMounts` to `nginxExtraVolumeMounts` (because it's not specific to configmaps).
- Add missing nginx Helm values in the test environment.
- Remove an unnecessary `if not (empty .)` inside a `with` block. (A `with` block with an empty argument is already a no-op.)

I've added AWS Secrets Manager secrets in both accounts.

https://trello.com/c/iausYhX2/848

#### Testing

Tested by overriding `targetRevision` to the PR branch in `app/argocd-apps` via `kubectl edit`, with a temporary commit to `charts/argocd-apps/values-test.yaml` to override `targetRevision` for the `govuk-apps-conf` and `router` apps. nginx serves `401 Authorization Required` without basic auth creds (and with wrong creds) and proxies the request when valid creds are supplied.

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -246,10 +246,16 @@ applications:
     nginxConfigMap:
       create: false
       name: router-nginx-conf
-      extraVolumeMounts:
-        - name: router-nginx-conf
-          mountPath: /usr/share/nginx/html/robots.txt
-          subPath: robots.txt
+    nginxExtraVolumeMounts:
+      - name: router-nginx-conf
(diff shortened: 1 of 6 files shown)


fredericfran-gds (approved):
looks good

PR #194, opened 2022-03-01
Fix assets smart-answers

fredericfran-gds:
(34af6072d982d6a5c7fa619fd74ed639973d7d1b) Allow assets path to be specified in the k8s job

Not all apps store their assets in the path `/app/public/assets/<app_name>`, e.g. smart-answers. Therefore, we introduce an override to be able to explicitly pass in the full path where the assets reside.

(de3c2e8337a191ca80537bb0c639ec67b76263a7) Specify assets paths for smart-answers
smart-answers store its assets under `/app/public/assets/smartanswers` rather `/app/public/assets/smart-answers`, i.e. no dash between smart and answers. Therefore, we specify the full explicit path.


sengi (approved):
Cool, thanks for adding the override!

PR #195, opened 2022-03-01
Fix assets smartanswers 2

fredericfran-gds:
(65535f8fbfaec07b04ead118d365427a5e7fd400) Fix https://github.com/alphagov/govuk-helm-charts/pull/194 explicit/full assets_path
In https://github.com/alphagov/govuk-helm-charts/pull/194, the full assets_paths feature was introduced but the changes where not propagated in whole bash. This PR fixes that.

In addition, the `s3Directory` parameter is introduced to specify the directory under `s3://govuk-app-assets-{{ .Values.govukEnvironment }}/assets/` where assets are stored.

(0a73bc2b452a89cc962b9321232c0ae7e8ea0f3f) specify new assets s3 directory for smart-answers

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -788,6 +788,7 @@ applications:
     replicaCount: 1
     uploadAssets:
       path: "/app/public/assets/smartanswers"
+      s3Directory: "smartanswers"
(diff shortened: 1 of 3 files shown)


sengi (approved):
Oh, good shout :D

PR #202, opened 2022-03-01
Add permissions to Signon Applications

bilbof:


fredericfran-gds (approved):
looks good. As discussed on Slack, these permissions are the set of permissions that particular apps support.

PR #205, opened 2022-03-01
Set GOV.UK Website Root to Fastly domain

bilbof:
Fastly bypasses authentication for requests from our VPC
which should resolve a number of smoke test failures caused
by requests from apps to apis served on the public website root
failing due to a lack of basic auth.

I've fixed a couple of these problems already by going app-by-app:
* https://github.com/alphagov/finder-frontend/pull/2738
* https://github.com/alphagov/publisher/pull/1552
and I think we should continue to do this.

This change is intended to catch any other instances of this issue
that we don't spot (though this change may hide them from us!).


bilbof:
Discussed offline with @sengi - closing this as we want to flush out other cases where apps are using public apis when internal apis are available.

PR #207, opened 2022-03-01
Adding Asset-manager Helm chart

nsabri1:
This change gives Asset manager application its own Helm chart -
	Will allow testing separate ClamAV sidecar.

We may discover that using the shared chart is a better option, at which point we will make the change.


sengi (approved):
I've only skimmed most of this since it's mostly a fork of `govuk-rails-app` for now (right?)

LGTM

Worth considering giving the Clam service its own chart (and therefore its own Argo app, and also then asset-manager itself could probably use govuk-rails-app again) - probably a bit more microservicey. If there's anything else that needs virus-scan things then it'd definitely make sense to have one scanning service rather than it being attached to asset-manager. Not a big deal for right now though.

PR #208, opened 2022-03-01
Use `pullPolicy: Always` for the mutably-tagged nginx image.

sengi:
Once we have a mechanism for automatically updating pinned image tags, we'll be able to stop referring to mutable tags like `stable`. Until then, it's the lesser of evils.

The `pullPolicy: Always` got accidentally dropped in #179 when switching from `stable` to `stable-perl` to support the lowercasing-URLs config.

Also fix the `govuk-rails-app` chart so that it actually honours the `pullPolicy` values that are passed in 😅


bilbof (approved):
Thanks, useful change 👍

PR #211, opened 2022-03-02
Disable service links (`*_SERVICE_*` env vars) on app pods.

sengi:
Disable [service links] (the automatically-set environment variables with names like `$service_SERVICE_{HOST,PORT}_*`) in the `govuk-rails-app` chart. We're not using them and they create a lot of clutter in the environment which can be a pain when debugging.

Maybe someday we could use them in Router and Plek, but that's unlikely to happen soon (and we already have too many service discovery mechanisms!)

#### Testing

* Ran `helm template . --set appImage.repository=foo --set workerEnabled=true` and inspected the output.
* Installed the chart (`helm install frontend-test ../govuk-rails-app --values <(helm template . --values values-integration.yaml |yq e '.|select(.metadata.name=="frontend").spec.source.helm.values')`), verified that pods come up healthy and the output of `kubectl exec frontend-test-d56c666c4-t4jbb -c frontend-test -- env` no longer contains dozens of unused `*_SERVICE_{HOST,PORT}` vars.

[service links]: https://kubernetes.io/docs/concepts/services-networking/connect-applications-service/#environment-variables

### DIFF
--- a/charts/govuk-rails-app/templates/deployment.yaml
+++ b/charts/govuk-rails-app/templates/deployment.yaml
@@ -86,6 +86,7 @@ spec:
           {{- with .Values.nginxExtraVolumeMounts }}
           {{ . | toYaml | trim | nindent 10 }}
           {{- end }}
+      enableServiceLinks: false
       {{- with .Values.dnsConfig }}
       dnsConfig:
         {{- . | toYaml | trim | nindent 8 }}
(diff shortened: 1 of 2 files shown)


smford (approved):

smford:
> We're 100% not using these at the moment in the app charts so there's no reason not to turn them off. If we want to use them in future them we'd just enable them again for whichever apps we want to use them with.

I think what I'm meaning is:  We probably wouldn't want to enable this in the future, and if we spot someone trying to do it, persuade them otherwise.  My worry is if we start using the two ways it would be horrid.

Anyway, this is just all academic and worrying about a situation that might not ever occur; I think this change you propose is the right one, and I'll just shoosh about my unreasonable concerns that dont even and might not ever exist.  I was just thinking out loud really.

sengi:
No worries - that's a good point about the risk of us growing yet another service discovery mechanism in the GOV.UK apps. I think that's a stronger reason for `enableServiceLinks: false` than my original one 😀

PR #212, opened 2022-03-02
Add specific nginx config for static

samsimpson1:
This adds a new nginx config file for static. It is required to serve a google site verification page (and will probably be needed for future work too)

Trello cards:
* https://trello.com/c/gSKgSrLJ/844-google-site-verification
* https://trello.com/c/jPruwYgF/807-canary-page


charts/govuk-apps-conf/templates/static-nginx-configmap.yaml
```diff
@@ -0,0 +1,69 @@
...
+          proxy_hide_header  x-amz-server-side-encryption;
+          proxy_hide_header  x-amz-version-id;
+          add_header         Cache-Control max-age=31536000;
+          proxy_intercept_errors on;
+          proxy_pass         https://govuk-app-assets-{{ .Values.govukEnvironment }}.s3.eu-west-1.amazonaws.com;
+        }
+      }
+
+      # Google site verification
+      location /googlec908b3bc32386239.html {
```
sengi, line 57: This verification page is currently only served on assets.publishing.service.gov.uk (see [govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/32c1bbb/modules/router/templates/assets_origin.conf.erb#L87)). Where does the `static` app come into this?

charts/govuk-apps-conf/templates/static-nginx-configmap.yaml
```diff
@@ -0,0 +1,69 @@
...
+        return 200 'google-site-verification: googlec908b3bc32386239.html';
+      }
+
+      # Endpoint that isn't cached, which is used to assert that an external
+      # service can receive a response from GOV.UK origin on assets hostname. It
+      # is intended for pingdom monitoring
+      location = /__canary__ {
```
sengi, line 64: I don't think we need `/__canary__` here. It only needs to be on Router and assets-origin.

sengi (commented):
I don't think this is needed - at least not yet.

charts/govuk-apps-conf/templates/static-nginx-configmap.yaml
samsimpson1: At the moment the nginx config that does all this is on the cache nodes, and it sends the traffic by default to static: https://github.com/alphagov/govuk-puppet/blob/32c1bbb/modules/router/templates/assets_origin.conf.erb#L82-L85

I assumed this would be a good place to put it, since it's the default backend. From what I can see, k8s doesn't have an ingress for assets-origin yet, but I thought this would be where it would go to if it was to replicate existing configuration

charts/govuk-apps-conf/templates/static-nginx-configmap.yaml
samsimpson1: I added it because the current assets-origin nginx config also has it. I can get rid of it if it isn't required

charts/govuk-apps-conf/templates/static-nginx-configmap.yaml
sengi: Oh, I see. In that case I think it might be in our interests to wait until we have assets-origin set up before adding these bits and pieces, then. Otherwise we're liable to add complexity without knowing whether it's needed.

samsimpson1: That makes sense to me. I'll update the trello card so nobody else tries to do the same thing

sengi: Cool, thanks!

PR #216, opened 2022-03-03
Feedback-helmchart-creation # 842

BhupsAL1:
As part of the new GOV.UK EKS, A Docker file has been created previously for the feedback app in the integration environment, this is a HelmChart Creation for this feedback app to enable it in Argo -CD
This is part of the All apps migration for launch.

Link URL to Trello ticket
https://trello.com/c/kBkPNDie/842-package-feedback-for-migration

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -833,3 +833,50 @@ applications:
           secretKeyRef:
             name: smartanswers-zendesk-client
             key: password
+- name: feedback
+  helmValues:
+    appImage:
+      repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/feedback
+      tag: latest  # To be edited by automation (not yet implemented).
+    replicaCount: 1
+    extraEnv:
+      - name: SECRET_KEY_BASE
+        valueFrom:
+          secretKeyRef:
+            name: rails-secret-key-base
+            key: SECRET_KEY_BASE
+      - name: SUPPORT_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-support-bearer-feedback
+            key: bearer_token
+      - name: SUPPORT_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-support-api-bearer-token-feedback
+            key: bearer_token
+      - name: PUBLISHING_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-publishing-api-bearer-feedback
+            key: bearer_token
+      - name: ASSISTED_DIGITAL_GOOGLE_SPREADSHEET_KEY
+        valueFrom:
+          secretKeyRef:
+            name: feedback-assisted-digital-google-sheet
+            key: sheet_id
+      - name: GOOGLE_CLIENT_EMAIL
+        valueFrom:
+          secretKeyRef:
+            name: feedback-assisted-digital-google-sheet
+            key: client_email
+      - name: GOOGLE_PRIVATE_KEY
+        valueFrom:
+          secretKeyRef:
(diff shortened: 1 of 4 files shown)


charts/argocd-apps/values-integration.yaml
```diff
@@ -833,3 +833,70 @@ applications:
...
+        valueFrom: 
+          secretKeyRef:
+            name: google-private-key-feedback
+            key: bearer_token
+      - name: GOVUK_NOTIFY_API_KEY
+        valueFrom: 
+          secretKeyRef:
+            name: govuk-notify-api-key-feedback
+            key: api_key
+       - name: GOVUK_NOTIFY_SURVEY_SIGNUP_TEMPLATE_ID
```
sengi, line 883: Some of these don't need to be (and therefore shouldn't be) secrets:

* [GOVUK_NOTIFY_SURVEY_SIGNUP_TEMPLATE_ID](https://github.com/alphagov/govuk-puppet/search?q=GOVUK_NOTIFY_SURVEY_SIGNUP_TEMPLATE_ID)
* [GOVUK_NOTIFY_SURVEY_SIGNUP_REPLY_TO_ID](https://github.com/alphagov/govuk-puppet/search?q=GOVUK_NOTIFY_SURVEY_SIGNUP_REPLY_TO_ID)
* [GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_TEMPLATE_ID](https://github.com/alphagov/govuk-puppet/search?q=GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_TEMPLATE_ID)
* [GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID](https://github.com/alphagov/govuk-puppet/search?q=GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID)

charts/argocd-apps/values-integration.yaml
```diff
@@ -833,3 +833,70 @@ applications:
...
+      - name: GOOGLE_CLIENT_EMAIL
+        valueFrom: 
+          secretKeyRef:
+            name: google-client-email-feedback
+            key: bearer_token
+      - name: GOOGLE_PRIVATE_KEY
+        valueFrom: 
+          secretKeyRef:
+            name: google-private-key-feedback
+            key: bearer_token
```
sengi, line 877: GOOGLE_CLIENT_EMAIL and GOOGLE_PRIVATE_KEY are a username/password pair and should therefore be stored as two keys in the same AWS/Kubernetes secret, for example:

```json
{
  "username": "foo@bar.com",
  "password": "hunter2"
}
```

sengi: This is so that we're not making it unnecessarily difficult for someone else to implement automatic rotation of these creds in future.

charts/argocd-apps/values-integration.yaml
sengi: The reason for this is that secrets are costly (in terms of human effort) to maintain, so we definitely don't want to be storing things as secrets that we don't really need to.

BhupsAL1: Thanks Chris - I'll remove these now.

charts/argocd-apps/values-integration.yaml
BhupsAL1: Hi Chris, I'll amend that accordingly, many thanks.

BhupsAL1: Amended

charts/argocd-apps/values-integration.yaml
BhupsAL1: Identifier templates within GOV.UK Notify have been removed.

sengi (approved):
Awesome - LGTM!

PR #221, opened 2022-03-08
Split out image tags into dedicated files

bilbof:
This creates a new directory apps/ which contains files which
specify the image tag to use in a given environment.

For instance:

```
├── Chart.yaml
├── apps
│   ├── asset-manager
│   │   └── images
│   │       ├── integration
│   │       ├── production
│   │       ├── staging
│   │       └── test
```

To find the image that should be deployed to production for asset-manager,
one can read apps/asset-manager/images/production to find the appropriate
image tag.

The ArgoCD Application resources pull in the files during templating,
so any update of these files will trigger deployments to kubernetes.

This will make it easier to update the image tag for an Application.
In order to continuous deployment and release promotions automatically
we must only overwrite a single file.
It will also be easier to monitor releases using e.g. git log on apps/asset-manager.

This approach introduces many new files, but overall the config isn't
more complicated. Originally I wanted to move all Helm values for an
app to dedicated file, since eventually values-<env>.yaml files will
be gigantic, hard to test, have a massive blast radius, and make it
harder to avoid duplication. However, I've avoided that for now, since
we probably want to take a step back and reconsider whether we want a
layer on top of or instead of Helm that would be a better fit.


sengi (approved):
Cool, LGTM. I think having these separate like this gives us a clean, simple interface for the rollout automation and should avoid merge conflicts on the big YAML files.

PR #224, opened 2022-03-09
Add metrics port for prometheus exporter to govuk-rails-app

marcpomfret:


charts/govuk-rails-app/templates/service.yaml
```diff
@@ -13,7 +13,11 @@ metadata:
 spec:
   type: {{ .Values.service.type }}
   ports:
-    - port: {{ .Values.service.port }}
+    - name: app
+      port: {{ .Values.service.port }}
       targetPort: {{ .Values.nginxPort }}
+    - name: monitor
```
bilbof, line 19: Good to use the names consistently (metrics not monitor)

charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -22,6 +22,8 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.appPort }}
+            - name: metrics
+              containerPort: {{ .Values.metricsPort }}
```
bilbof, line 26: I wonder if we could define .Values.extraPorts` or something like that? I guess we want all apps to export metrics, so this is probably fine.

bilbof (approved):
Looking good. I think we'll want the requests to the metrics port to go through the Nginx sidecar, so we can monitor the `/metrics` requests (through Nginx metrics) and enforce policies like timeouts in a consistent way. This is fine for now however, since we're still working on this. Maybe worth adding a TODO in there.

Approving since we are actively working on this and will fix up issues once our spike is complete.

charts/govuk-rails-app/templates/service.yaml
marcpomfret: yup agreed, will get that changed

charts/govuk-rails-app/templates/deployment.yaml
marcpomfret: yeah could do, i'm not really sure of what other ports that there might be coming at the minute, could change it to this then if there are more to add 
extraPorts:
  metricsPort: 9394

marcpomfret: suppose it will be easy to do later as well

PR #231, opened 2022-03-09
fix search-api

fredericfran-gds:
fix signon-resources for signon-api and fix plek entry for signon


smford (approved):
Cool! search-api in eks!

PR #232, opened 2022-03-09
Use www instead of www-origin as GOVUK_WEBSITE_ROOT.

sengi:
This should make apps construct hrefs with the `www.eks.integration.govuk.digital` domain instead of `www-origin.eks.integration.govuk.digital`, which should fix several issues including surprising Basic Auth prompts on Signon and fonts not loading because of CORS issues.

[Trello](https://trello.com/c/9BPpFRBe/851)


fredericfran-gds (approved):
cool, thanks

PR #237, opened 2022-03-10
Add pre-push hook to prevent pushing to main

bilbof:
We are no longer enforcing branch protection constraints on administrators on this repository. This enables repo admins to
push to main. To prevent this, we have installed a git pre-push hook.

The reason we have disabled this branch protection is we would like the administrator user govuk-ci to bypass the PR and status check process and push directly to main (automatically) without needing to raise a PR.

### DIFF
--- a/git-hooks/pre-push
+++ b/git-hooks/pre-push
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# @link https://gist.github.com/vlucas/8009a5edadf8d0ff7430
+#
+# Called by "git push" after it has checked the remote status,
+# but before anything has been pushed.
+#
+# If this script exits with a non-zero status nothing will be pushed.
+#
+# Steps to install, from the root directory of your repo...
+# 1. Copy the file into your repo at `.git/hooks/pre-push`
+# 2. Set executable permissions, run `chmod +x .git/hooks/pre-push`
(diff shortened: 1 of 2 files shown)


git-hooks/pre-push
```diff
@@ -0,0 +1,64 @@
...
+# `git push --force origin main`
+# `git push --delete origin main`
+#
+# Set the environment variable $REALLY_PUSH_TO_MAIN to bypass this pre-push
+# hook, i.e. `REALLY_PUSH_TO_MAIN=1 git push origin main`. Note: this is only
+# intended for use in automation (i.e. CI/CD).
+
+protected_branch='main'
+
+policy="\n\n[Policy] You cannot push code directly to the "$protected_branch" branch. (Prevented with pre-push hook.)\n\t  Use --really-push-to-$protected_branch to override this policy.\n\n"
```
sengi, line 28: This says that `--really-push-to-main` is a command-line option but it's implemented only as an env var.

sengi (approved):

git-hooks/pre-push
bilbof: Ah great catch, thank you 👍 Yes couldn't use this flag as it gets passed to `git-push`.

PR #239, opened 2022-03-10
Fix assets paths for Whitehall and remove an unnecessary domain suffix.

sengi:
The assets upload job was defaulting to nonexistent paths for the three Whitehall deployments. Since whitehall-admin, whitehall-frontend and draft-whitehall-frontend each run the same image (though not necessarily the same version) with a different `.Release.Name` and none of those names match the assets directory inside the image, we need to override it.

We still run the assets upload job independently for each of the three deployments because it's possible for them to be running different versions and therefore need to upload different assets files - even though the vast majority of the time it'll just be fruitlessly uploading the same thing three times.

Also clean up an unnecessary domain suffix while we're there.

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -533,6 +533,9 @@ applications:
   helmValues:
     dbMigrationEnabled: true
     workerEnabled: true
+    uploadAssets: &whitehall-upload-assets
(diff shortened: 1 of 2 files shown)


fredericfran-gds (approved):
looks good, thanks.

sengi:
Forgot to update `values-test.yaml` (force-pushed just now to fix).

PR #240, opened 2022-03-10
Validate Kubernetes manifests against schemas

bilbof:
This adopts [kubeconform](https://github.com/yannh/kubeconform) for pre-merge checks.

This should suffice for now to eliminate typos and other minor Helm / Kubernetes manifest errors.

Schemas for the CRDs were generated from the corresponding CRD manifests [see kubeconform doc](https://github.com/yannh/kubeconform#converting-an-openapi-file-to-a-json-schema).

The tests can be run locally after installing `kubeconform`:

```shell
# generate manifests from Helm templates
for c in $(ls charts); do helm template $c "charts/${c}" --output-dir helm-dist; done
# validate manifests
kubeconform -summary -schema-location default -schema-location './schemas/{{ .ResourceKind }}_{{ .ResourceAPIVersion }}.json' helm-dist
```

Example failure:

```
helm-dist/cluster-secrets/templates/dex-github.yaml - ExternalSecret govuk-dex-github is invalid: For field spec: Additional property test is not allowed
```

In future it would be great to spin a test cluster, install the helm charts (e.g. using [`ct install`](https://github.com/helm/chart-testing/blob/main/doc/ct_install.md) since we are already making use of `ct lint`), and run potentially run e2e tests.

We could also test the charts using the values provided for each environment (i.e. generate manifests for test, integration, staging, and production, passing in the appropriate values). This would give more confidence that the config we specify for apps in argocd-apps is valid.

I would recommend reviewing by commit; the first commit includes the JSON schemas for the CRDs we reference in our Helm charts, which were autogenerated.

Trello: https://trello.com/c/pEXfFZBP/830-write-helm-validation-tests


.github/workflows/ci.yml
```diff
@@ -33,3 +33,22 @@ jobs:
 
       - name: Run chart-testing (lint)
         run: ct lint --config ct.yml
+
+      - name: Helm template
+        run: |
+          mkdir helm-dist
+          for c in $(ls charts)
```
sengi, line 40:
```suggestion
          for c in charts/*
```

charts/govuk-rails-app/values.yaml
```diff
@@ -29,7 +29,7 @@ workerReplicaCount: 1
 dbMigrationEnabled: false
 
 appImage:
-  repository: ""  # Dummy value, overridden in ArgoCD config.
+  repository: "172025368201.dkr.ecr.eu-west-1.amazonaws.com/dummy-value"  # Dummy value, overridden in ArgoCD config.
```
sengi, line 32: Does this defeat the `required` in the template?

.github/workflows/ci.yml
```diff
@@ -33,3 +33,22 @@ jobs:
...
+          do
+            helm template $c "charts/${c}" --output-dir helm-dist
+          done
+
+      - name: Login to Github Packages
+        run: echo "${{ github.token }}" | docker login https://ghcr.io -u ${GITHUB_ACTOR} --password-stdin
+
+      - name: kubeconform
+        uses: docker://ghcr.io/yannh/kubeconform:master
```
sengi, line 49: Do we need to pin this?

sengi (approved):
Awesome!

I think the `schemas` dir could use a README.md to say where the schemas come from (just like you've mentioned in the PR description).

.github/workflows/ci.yml
sengi: Ah sorry my suggestion isn't quite equivalent

.github/workflows/ci.yml
```diff
@@ -33,3 +33,22 @@ jobs:
 
       - name: Run chart-testing (lint)
         run: ct lint --config ct.yml
+
+      - name: Helm template
+        run: |
+          mkdir helm-dist
+          for c in $(ls charts)
+          do
+            helm template $c "charts/${c}" --output-dir helm-dist
```
sengi, line 42:
```suggestion
            helm template "${c}" "${c}" --output-dir helm-dist
```

charts/govuk-rails-app/values.yaml
bilbof: Yep good point. I'll remove the `required` constraint and set this to `""`. Hopefully we have a constraint in the schemas.

bilbof:
Thanks @sengi I've made those changes!

PR #241, opened 2022-03-11
Fix request routing to Search API and Account API.

sengi:
We intend to use Search API and Account API in EC2 for now, while we're focusing on frontend (www) functionality. There were various configuration issues preventing these requests from being routed correctly.

- Configure Router to forward to EC2 search-api and account-api.
- Configure apps (via Plek) to use EC2 search-api.
- Use the internal load balancer addresses (`govuk-internal.digital`) where we can.
- It's `PLEK_SERVICE_ACCOUNT_API_URI`, not `PLEK_SERVICE_ACCOUNT_URI`, because the client code calls [`Plek.find("account-api")`][1].

Not making these changes in the test environment because the EC2 stuff in the test account is too broken for this to be meaningful there.

[1]: https://github.com/alphagov/gds-api-adapters/blob/599e216/lib/gds_api.rb#L34

### DIFF
--- a/charts/govuk-apps-conf/templates/configmap.yaml
+++ b/charts/govuk-apps-conf/templates/configmap.yaml
@@ -14,14 +14,16 @@ data:
   GOVUK_ENVIRONMENT: {{ .Values.govukEnvironment }}
   GOVUK_ENVIRONMENT_NAME: {{ .Values.govukEnvironment }}
   GOVUK_WEBSITE_ROOT: https://www.{{ .Values.externalDomainSuffix }}
-  PLEK_SERVICE_ACCOUNT_URI: https://account-api.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
+  # TODO: switch back to in-cluster Account API once it's fully working.
+  PLEK_SERVICE_ACCOUNT_API_URI: https://account-api.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
   PLEK_SERVICE_CONTENT_STORE_URI: https://content-store.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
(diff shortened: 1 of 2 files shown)


charts/govuk-apps-conf/templates/configmap.yaml
```diff
@@ -14,14 +14,16 @@ data:
...
+  PLEK_SERVICE_ACCOUNT_API_URI: https://account-api.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
   PLEK_SERVICE_CONTENT_STORE_URI: https://content-store.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
   PLEK_SERVICE_DRAFT_ORIGIN_URI: https://draft-origin.{{ .Values.externalDomainSuffix }}
   PLEK_SERVICE_EMAIL_ALERT_API_URI: https://email-alert-api.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
   PLEK_SERVICE_FRONTEND_URI: http://frontend.{{ .Values.internalDomainSuffix }}
   PLEK_SERVICE_PUBLISHING_API_URI: http://publishing-api.{{ .Values.internalDomainSuffix }}
   PLEK_SERVICE_ROUTER_API_URI: https://router-api.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
-  PLEK_SERVICE_SEARCH_URI: http://search-api.{{ .Values.internalDomainSuffix }}
+  # TODO: switch back to in-cluster Search API once it's fully working.
+  PLEK_SERVICE_SEARCH_URI: http://search-api.{{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}
```
bilbof, line 26: Should this be HTTPS?

sengi: Ooh, good catch, thanks! Fixed.

bilbof (approved):

PR #252, opened 2022-03-17
Add job to upload error pages to S3 and serve 404 error pages from S3

samsimpson1:
This adds a job that uploads rendered error pages from static to S3 whenever static is deployed. It also gets nginx in front of router to start serving 404 error pages from S3.

Trello card: https://trello.com/c/SaGBEzbI/775-return-static-error-pages-from-origin


sengi (approved):
Couple of suggestions to ensure that error paths work as intended — for example not writing an error to the file instead of the actual content if something goes wrong. Otherwise LGTM. Thanks for sorting this out - I think it's a nice clean solution.

PR #256, opened 2022-03-18
Improve post-sync workflow name

fredericfran-gds:
Currently, all post-sync workflows are named
`post-sync-<random-string>` which makes it hard to
decipher which post-sync is associated to which
application.

Now, we prepend the application name to the post-sync name.


lwilts (approved):
Assuming this works, LGTM!

PR #260, opened 2022-03-18
Add promote release step to postsync workflow

theseanything:
This adds a workflowTemplate to update the image tag value in the helm chart, which is used by the promote release step of the post-sync workflow.

### DIFF
--- a/charts/argo-workflows/templates/post-sync-workflow.yaml
+++ b/charts/argo-workflows/templates/post-sync-workflow.yaml
@@ -27,11 +27,20 @@ spec:
               parameters:
                 - name: extra-args
                   value: "{{" and @app-{{workflow.parameters.application}}"}}"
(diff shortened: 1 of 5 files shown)


charts/argo-workflows/templates/post-sync-workflow.yaml
```diff
@@ -27,11 +27,19 @@ spec:
               parameters:
                 - name: extra-args
                   value: "{{" and @app-{{workflow.parameters.application}}"}}"
-          # TODO: Promote releases
-          # - name: promote-release
-          #   templateRef:
-          #     name: promote-release
-          #     template: promote-release
+          - name: promote-release
```
lwilts, line 30: You can add 
depends: smoke-test.Succeded
here

lwilts (approved):

PR #261, opened 2022-03-18
add annotation to enable switching-on/off postsync workflow

fredericfran-gds:
Some apps do not need a postsync workflow since currently
the workflow does a smokey run and slack notification.

if the postSyncWorkflowEnabled annotation of an argo app is set
to false, no trigger is sent to argo events & workflow to trigger
a postsync workflow.


nsabri1 (approved):
LGTM

PR #266, opened 2022-03-21
Fix location of the workflow scripts

theseanything:
The scripts folder should have been created as a subdirectory in the argo-workflows chart. This is will fix the update-image-tag workflowTemplate which references a script from this location.


nsabri1 (approved):
LGTM

PR #268, opened 2022-03-21
fix smokey

fredericfran-gds:
some env values from smokey were outdated with the implementation of fastly for GOV.UK EKS integration.

To avoid setting common variables in Smokey again, they are now taken from the common confimap `govuk-apps-env`.


nsabri1 (approved):
LGTM

PR #270, opened 2022-03-22
Allow any host header match for govuk-rails-app ingress

fredericfran-gds:
Allow wildcard host header specification for k8s ingress.
This is needed for the origin of GOV.UK where fastly will
use the `www.*` domain in the host header field rather than
the `www-origin.*` domain.

Some GOV.UK apps, such as whitehall, uses the host header to
construct a response.

the `ingress.hosts` is now a list which specifies the DNS
domain to be created as well as the tls cert to be attached.

Ref:
1. [trello card](https://trello.com/c/cQm9F4Bf/879-make-www-origin-ingress-ignore-the-host-header)
2. [tls cert](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/cert_discovery/#discover-via-ingress-tls)
3. [dns](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/alb-ingress.md#ingress-examples)

### DIFF
--- a/charts/govuk-rails-app/values.yaml
+++ b/charts/govuk-rails-app/values.yaml
@@ -14,9 +14,8 @@ ingress:
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: "443"
     alb.ingress.kubernetes.io/healthcheck-path: /healthcheck/ready
-  hosts:
-  - name:
-    path: /
+  rules:
+  - path: /
     pathType: Prefix
 
 replicaCount: 1
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -26,7 +26,7 @@ applications:
       annotations:
         alb.ingress.kubernetes.io/load-balancer-name: publisher
       hosts:
-      - name: publisher.eks.integration.govuk.digital
+      - publisher.eks.integration.govuk.digital
     replicaCount: 1
     workerReplicaCount: 1
     extraEnv:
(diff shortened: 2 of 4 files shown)


charts/govuk-rails-app/values.yaml
```diff
@@ -14,9 +14,8 @@ ingress:
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: "443"
     alb.ingress.kubernetes.io/healthcheck-path: /healthcheck/ready
-  hosts:
-  - name:
-    path: /
+  rules:
+  - path: /
```
sengi, line 18: I think you're missing the `http` and `paths` here. See `kubectl explain ing.spec.rules` - the next level down has to be either a `host` (which we don't want any more of course) or an `http`. Then `kubectl explain ing.spec.rules.http` tells us that `http` has to contains a `paths`, which is a list of `path`. Here's a [full example](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/cert_discovery/#discover-via-ingress-tls).

sengi (approved):
Cool, thanks for sorting this out! I think you're missing a couple of lines from the Ingress spec in the ~template~ values file but otherwise looks great 👍

charts/govuk-rails-app/values.yaml
sengi: Oops sorry, I initially misread this as the spec rather than the Helm value being passed in, then when I noticed it was the Helm value I didn't go back and check it against the template. D'oh. You're absolutely right - it's all good.

PR #271, opened 2022-03-22
Fix bootstrap-signon-resources config and `feedback` ApiUser.

sengi:
- Make it slightly clearer as to which Applications are EC2 vs. ECS.
- Generate OAuth bearer tokens correctly for the Feedback app to auth to Support and Support API.
- Add some documentation so hopefully others won't have to go on the same magical mystery tour.

See individual commits for details.

Tested: ran manually by extracting the config\* and starting the job with `kubectl apply -f job.yaml`.

https://trello.com/c/kBkPNDie/842
https://trello.com/c/qyKykw22/859

\* like this: `helm template . --values <(helm template ../argocd-apps --values ../argocd-apps/values-integration.yaml |yq e '.|select(.metadata.name=="signon-resources").spec.source.helm.values') |yq e '.|select(.kind=="Job")' >job.yaml`

### DIFF
--- a/charts/signon-resources/templates/bootstrap-job.yaml
+++ b/charts/signon-resources/templates/bootstrap-job.yaml
@@ -1,3 +1,47 @@
+# This job creates objects in GOV.UK Signon which are needed for end-user and
+# app-to-app authentication/authorisation. It creates:
+#
+# - Signon Doorkeeper::Application objects
+# - Signon ApiUser objects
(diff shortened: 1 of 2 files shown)


sengi:
Added another point to the docs comment (hence the force-push).

nsabri1 (approved):
LGTM - thanks for adding

PR #272, opened 2022-03-22
Allow multiple hostnames match for govuk-rails-app ingress

fredericfran-gds:
Add the ability to have multiple hostnames match for a given
app.

This is needed for the origin of GOV.UK where fastly will
use the `www.*` domain in the host header field rather than
the `www-origin.*` domain.

Some GOV.UK apps, such as whitehall, uses the host header to
construct a response.

The `ingress.hosts` is now a list which specifies the DNS
domain to be created as well as the tls cert to be attached.

Ref:
1. [trello card](https://trello.com/c/cQm9F4Bf/879-make-www-origin-ingress-ignore-the-host-header)
2. [tls cert](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/ingress/cert_discovery/#discover-via-ingress-tls)
3. [dns](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/alb-ingress.md#ingress-examples)

### DIFF
--- a/charts/govuk-rails-app/templates/ingress.yaml
+++ b/charts/govuk-rails-app/templates/ingress.yaml
@@ -10,28 +10,27 @@ metadata:
     app.kubernetes.io/component: web
   annotations:
     {{- toYaml .Values.ingress.annotations | nindent 4 }}
+    external-dns.alpha.kubernetes.io/ingress-hostname-source: annotation-only
+    external-dns.alpha.kubernetes.io/hostname: {{ join "," .Values.ingress.hosts }}
 spec:
-  {{- if .Values.ingress.tls }}
+  {{- if .Values.ingress.hosts }}
   tls:
-    {{- range .Values.ingress.tls }}
     - hosts:
-        {{- range .hosts }}
+        {{- range .Values.ingress.hosts }}
         - {{ . | quote }}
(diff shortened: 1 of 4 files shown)


nsabri1 (approved):
LGTM

charts/govuk-rails-app/templates/ingress.yaml
```diff
@@ -10,28 +10,27 @@ metadata:
...
   annotations:
     {{- toYaml .Values.ingress.annotations | nindent 4 }}
+    external-dns.alpha.kubernetes.io/ingress-hostname-source: annotation-only
+    external-dns.alpha.kubernetes.io/hostname: {{ join "," .Values.ingress.hosts }}
 spec:
-  {{- if .Values.ingress.tls }}
+  {{- if .Values.ingress.hosts }}
   tls:
-    {{- range .Values.ingress.tls }}
     - hosts:
```
sengi, line 18: I'd much prefer we just have it accept any hostname. It's easier to understand and easier to test, which will means fewer things to go wrong when we cut over in production.

fredericfran-gds:
close in favour of https://github.com/alphagov/govuk-helm-charts/pull/270

PR #274, opened 2022-03-22
Fix pre-push hook by setting its exec mode bit.

sengi:
Otherwise it fails with:

    hint: The 'git-hooks/pre-push' hook was ignored because it's not set as executable.
    hint: You can disable this warning with `git config advice.ignoredHook false`.

Had to fix shellcheck lints for checks to pass, since we added the shellcheck check after this script was added.


nsabri1 (approved):
LGTM

PR #275, opened 2022-03-22
Adding name to ingress fro Asset-manager as its using seperate ingress to gouvuk-rails-app Helm chart  

nsabri1:
As Asset-manager is using its own Helm chart and not govuk-rails-app, the changes made to ingress dont affect it for now. 
Therefore adding name back to ingress host.


fredericfran-gds (approved):
thanks for fixing this.

PR #276, opened 2022-03-23
Add service manual frontend

fredericfran-gds:
Add service manual frontend to EKS integration

We add the image tag files too.


alexbakervr (approved):
LGTM!

PR #278, opened 2022-03-23
Allow concurrent smokey cron job

fredericfran-gds:
It is ok to allow concurrent smokey loop as on average a
feature/step will be then tested every 10 minutes.

Currently, some steps are not run because Smokey take longer than
10 minutes to run now.


nsabri1 (approved):
LGTM

PR #279, opened 2022-03-23
Add resources to router app

fredericfran-gds:
Router app is suffering from cpu issues which result in some
buffering. This attempts to solve that.

Note that load testing will be done later to get more accurate
value for the resources and will include using more than
1 replica.

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -233,6 +233,13 @@ applications:
     replicaCount: 1
     uploadAssets:
       enabled: false
+    appResources:
+      limits:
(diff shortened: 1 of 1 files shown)


nsabri1 (commented):
lgtm

nsabri1 (approved):

sengi (approved):
It's being throttled a bit, but it doesn't look too bad. I would probably just increase the limit rather than the reservation but it's not a big deal - as you say, we'll adjust these when we load test anyway. https://grafana.eks.integration.govuk.digital/goto/0G4h4qP7z?orgId=1

PR #280, opened 2022-03-23
Point frontend (Milestone 1) apps at backends/APIs in EC2.

sengi:
For milestones 1 through 4, we've said we'll have frontend apps running in k8s and talking to the existing backends/APIs in EC2 - see [diagram] (link is internal only, sorry).

Ensure that we're defining EKS and non-EKS versions of Signon Applications as appropriate. Apps which are in k8s should have a corresponding `[EKS]` app in Signon. Apps which are used by frontend apps in k8s should have a non-EKS (`-ec2`) Signon app defined. Frontend apps should have tokens for the EC2 backends/APIs and use those until after [Milestone 4].

Add `-ec2` suffices to add slugs and secret names to make it clearer when we're pointing at the old stack.

Remove `-eks` suffixes from app slugs and secret names. When we've finished moving apps to k8s, those suffixes would become meaningless. The ` [EKS]` suffixes in the app names stay, because we don't want to rename the original apps in Signon just yet and we still need a way to differentiate them in the Signon UI.

Remove `-eks` suffixes from ApiUser slugs. These are internal to the bootstrap script and are not used in Signon itself. All ApiUsers created by the bootstrap script are for apps in EKS, so this suffix is redundant.

I audited the frontend apps' configs in `charts/argocd-apps/values-integration.yaml` to determine which EC2 backends are used by the frontends and therefore which backend apps need tokens in EC2, EKS or both:

- Account API isn't in k8s and email-alert-frontend has a token for it.
- Email Alert API isn't in k8s and Collections, Email Alert Frontend and Finder Frontend have tokens for it.
- Router API is in both. Content Store needs a token for it (though probably only for a rake task and not for serving). Use EC2 Router API for now. We're counting Content Store as a frontend app because Router forwards traffic to it.
- Asset Manager is in both (but still a work in progress in k8s). Whitehall has a token for it. (Probably only whitehall-admin really needs it, but all the Whitehall instances share the same config in k8s, same as in Puppet.) Safer to match the existing config than try to get away without the token on whitehall-frontend, since we have limited test coverage.
- Link Checker API isn't in k8s and Smart Answers and Whitehall have tokens for it.
- Publishing API is in both and several frontend apps have tokens for it.
- No frontend apps have tokens for Publisher (but it's already in EKS).
- No frontend apps have tokens for Whitehall (but it's already in EKS).
- Search API is in both. Whitehall has a token for it.
- Support isn't in k8s and Feedback has a token for it.
- Support API isn't in k8s and Feedback has a token for it.
- No frontend apps have tokens for HMRC Manuals API (but it's already in EKS).

Also:

- Alphabetise app permissions and consistently omit the built-in `signin` permission (which Signon always creates implicitly).

- Make the app description fields match what's actually in Signon, so that we don't inadvertently change them. In some cases our descriptions were less informative than the originals.

Testing: easiest to merge and fix forward with this one I think. I've checked for things like missing security group rules to allow traffic from the cluster to the apps in EC2.

https://trello.com/c/FmnogFnn/880

[diagram]: https://docs.google.com/presentation/d/1qz8c9A-2hIw0i5H67OtKTc7vF4QtoVp7G0pe-fC-WJw/edit#slide=id.g1152675d55e_0_515
[Milestone 4]: https://docs.google.com/presentation/d/1qz8c9A-2hIw0i5H67OtKTc7vF4QtoVp7G0pe-fC-WJw/edit#slide=id.g1152675d55e_0_632

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -33,12 +33,12 @@ applications:
       - name: GDS_SSO_OAUTH_ID
         valueFrom:
           secretKeyRef:
-            name: signon-app-publisher-eks
+            name: signon-app-publisher
             key: oauth_id
       - name: GDS_SSO_OAUTH_SECRET
         valueFrom:
           secretKeyRef:
-            name: signon-app-publisher-eks
+            name: signon-app-publisher
             key: oauth_secret
       - name: ASSET_MANAGER_BEARER_TOKEN
         valueFrom:
@@ -439,7 +439,7 @@ applications:
       - name: PUBLISHING_API_BEARER_TOKEN
         valueFrom:
           secretKeyRef:
-            name: signon-token-frontend-publishing-api
+            name: signon-token-frontend-publishing-api-ec2
             key: bearer_token
       - name: MEMCACHE_SERVERS
         value: frontend-memcached-govuk.eks.integration.govuk-internal.digital
@@ -465,7 +465,7 @@ applications:
       - name: ROUTER_API_BEARER_TOKEN
         valueFrom:
           secretKeyRef:
-            name: signon-token-content-store-router-api
+            name: signon-token-content-store-router-api-ec2
             key: bearer_token
       - name: GDS_SSO_OAUTH_ID
         valueFrom:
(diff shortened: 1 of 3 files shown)


fredericfran-gds (approved):
cool, I guess when ec2 apps are referred to in the `applications` section, it is just to get the reference to apply later in `api_users` section and no config modification of existing ec2 apps is being done in signon?

sengi:
> cool, I guess when ec2 apps are referred to in the `applications` section, it is just to get the reference to apply later in `api_users` section and no config modification of existing ec2 apps is being done in signon?

Yeah that's right. It would update the description field and add/remove permissions if there were differences, but I went through all of them and made the config here match exactly what's in integration Signon. (It doesn't currently update the URLs - those are currently ignored when it finds an existing application with the same name).

PR #285, opened 2022-03-23
Adding Armosec Kubescape as an ArgoCD application

lwilts:
This will perform in-cluster scanning for vulnerabilities.

### DIFF
--- a/charts/argocd-apps/templates/application.yaml
+++ b/charts/argocd-apps/templates/application.yaml
@@ -29,7 +29,7 @@ spec:
         {{- end }}
         appImage:
           repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/{{ $reponame }}
-          tag: {{ $.Files.Get (printf "image-tags/%s/%s" $.Values.globalHelmValues.govukEnvironment $reponame) }}
+          tag: {{ $.Files.Get (printf "image-tags/%s/%s" $.Values.globalHelmValues.govukEnvironment $reponame) | trim }}
         {{- with .helmValues }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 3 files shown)


charts/argocd-apps/templates/application.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         {{- end }}
         appImage:
           repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/{{ $reponame }}
-          tag: {{ $.Files.Get (printf "image-tags/%s/%s" $.Values.globalHelmValues.govukEnvironment $reponame) }}
+          tag: {{ $.Files.Get (printf "image-tags/%s/%s" $.Values.globalHelmValues.govukEnvironment $reponame) | trim }}
```
lwilts, line 32: This fixes an aesthetic bug - an extra newline in the resulting manifest.

charts/argocd-apps/templates/application.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: argoproj.io/v1alpha1
 kind: Application
 metadata:
   name: {{ .name }}
-  {{- if has $.Values.govukEnvironment (list "test" "integration") }}
+  {{- if has $.Values.globalHelmValues.govukEnvironment (list "test" "integration") }}
```
lwilts, line 7: Don't think this would have been working as intended before

smford (approved):
looks good!

charts/argocd-apps/templates/kubescape.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.kubescape }}
+{{- $kubescapesecret := (lookup "v1" "secret" "cluster-services" "govuk-kubescape") }}
```
sengi, line 2: This might lead to a bootstrap issue, but it's not a big deal for now.

The (potential) issue is that there's a race between this chart being deployed and the Secret being populated, and when the Secret doesn't exist, Helm doesn't retry or fail but returns an empty string. So there's a chance the chart may fail to be deployed on cluster turnup and need a manual retry. Who knows whether it'd actually happen in practice though. The solution is not to read secrets in Helm and only refer to them in the kube YAML itself (i.e. `secretKeyRef` or volumes).

sengi (commented):
Sorry for the late review. LGTM - though there's a potential (maybe theoretical, maybe real) race condition on cluster turnup. No need to worry about it for now though.

PR #287, opened 2022-03-24
Redirect if URL finishes with a trailing slash

fredericfran-gds:
redirect if URL finishes with a trailing slash to
URL without trailing slash.

Might need to add this to fastly too.

This was set in varnish before, [see](https://github.com/alphagov/govuk-puppet/blob/2383a96a9188d729340d97c7303c7409a18bdf6d/modules/varnish/templates/default.vcl.erb#L36)

### DIFF
--- a/charts/govuk-apps-conf/templates/router-configmap.yaml
+++ b/charts/govuk-apps-conf/templates/router-configmap.yaml
@@ -92,6 +92,11 @@ data:
           return 404;
         }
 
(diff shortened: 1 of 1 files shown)


smford (approved):

sengi:
Yeah if this was in Varnish before then we should definitely add it to Fastly, otherwise we're creating extra load on the cluster that we didn't have in the old system.

PR #288, opened 2022-03-24
Add custom response headers in nginx

samsimpson1:
Add headers to default nginx config:
* Strict-Transport-Security
* GOVUK-Request-Id
* X-Frame-Options
* X-Robots-Tag
* Cache-Control and Surrogate-Key for `/assets`

Add new nginx config for asset-manager that replicates functionality on EC2

Co-authored-by: samsimpson1 <sam.simpson@digital.cabinet-office.gov.uk>
Co-authored-by: jibolaolu <seun.adiomowo@digital.cabinet-office.gov.uk>

Trello: https://trello.com/c/uCGahIqu/862-custom-response-headers-at-origin

### DIFF
--- a/charts/govuk-apps-conf/templates/asset-manager-nginx-configmap.yaml
+++ b/charts/govuk-apps-conf/templates/asset-manager-nginx-configmap.yaml
@@ -1,0 +1,167 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: asset-manager-nginx-conf
+  labels:
+    {{- include "govuk-apps-conf.labels" . | nindent 4 }}
+    app.kubernetes.io/component: web
+data:
+  nginx.conf: |-
+    user nginx;
+
+    error_log /dev/stderr warn;
+    pid /var/run/nginx.pid;
+
+    events {
+      worker_connections 1024;
+    }
+
+    http {
+      upstream asset-manager-proxy {
+        server localhost:3000;
+      }
+
+      # Set GOVUK-Request-Id if not set
+      # See http://nginx.org/en/docs/http/ngx_http_perl_module.html
+      perl_modules perl/lib;
+      perl_set $govuk_request_id '
+        sub {
+          my $r = shift;
+          my $current_header = $r->header_in("GOVUK-Request-Id");
+
+          if (defined $current_header && $current_header ne "") {
+            return $current_header;
+          } else {
+            my $pid = $r->variable("pid");
+            my $msec = $r->variable("msec");
+            my $remote_addr = $r->variable("remote_addr");
+            my $request_length = $r->variable("request_length");
+            return "$pid-$msec-$remote_addr-$request_length";
+          }
+        }
+      ';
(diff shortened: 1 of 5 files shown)


charts/govuk-apps-conf/templates/asset-manager-nginx-configmap.yaml
```diff
@@ -0,0 +1,167 @@
...
+        proxy_set_header X-Forwarded-Server $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header GOVUK-Request-Id $govuk_request_id;
+        proxy_redirect off;
+        proxy_connect_timeout 1s;
+        proxy_read_timeout 60;
+
+        access_log /var/log/nginx/asset-manager-access.log timed_combined;
+        access_log /var/log/nginx/asset-manager-json.event.access.log json_event;
+        error_log /var/log/nginx/asset-manager-error.log;
```
sengi, line 65: As discussed, we'll need to ensure these aren't filling up ephemeral storage on the container (and are being sent to LogIt or wherever they're supposed to go). No probs if you wanna do that in a separate PR or in this one.

charts/govuk-apps-conf/templates/router-configmap.yaml
```diff
@@ -32,6 +32,36 @@ data:
...
+      ';
+
+      # This map creates a $sts_default variable for later use.
+      # If this header is already set by upstream, then $sts_default will
+      # be an empty string, which will later lead to:
+      #    add_header Strict-Transport-Security ''
+      # which will be ignored according to http://serverfault.com/a/598106
+      # If the header is not set by upstream, then $sts_default will be set
+      # and later uses in add_header will be effective.
+      map $upstream_http_strict_transport_security $sts_default {
```
sengi, line 61: So I understand this just comes verbatim from the old config, so let's assume for now that it works as intended 😅 I agree it's a good idea to keep it as-is for now and leave potential improvements for maybe later 👍

charts/govuk-rails-app/templates/nginx-configmap.yaml
```diff
@@ -35,12 +65,22 @@ data:
         proxy_connect_timeout {{ .Values.nginxProxyConnectTimeout }};
         proxy_read_timeout    {{ .Values.nginxProxyReadTimeout }};
 
+        add_header Strict-Transport-Security $sts_default;
+        add_header X-Frame-Options DENY;
```
sengi, line 69: As discussed, probably a good idea to omit this if it wasn't in the original config. The homepage seems to have `x-frame-options: ALLOWALL`, for example (presumably coming from the application itself?)

sengi (approved):
Read through this together on a hangout. Makes sense - I understand almost all of the gnarly bits of config are verbatim from the old Puppet rig, so that's all good for now. I've annotated the couple of potential changes that we talked about. Thanks for sorting through all this stuff!

PR #291, opened 2022-03-24
Added license-finder to values-integration and values-test

NH10:
https://trello.com/c/F7vSHnws/841-package-licence-finder-for-migration

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -991,10 +991,12 @@ applications:
           secretKeyRef:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
-- name: license-finder
+- name: licence-finder
   helmValues:
-    appImage: 
     replicaCount: 1
+    uploadAssets:
+      path: "/app/public/assets/licencefinder"
+      s3Directory: "licencefinder"
     extraEnv:
       - name: SECRET_KEY_BASE
         valueFrom:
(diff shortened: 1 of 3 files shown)


charts/argocd-apps/values-integration.yaml
```diff
@@ -991,6 +991,21 @@ applications:
           secretKeyRef:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
+- name: license-finder
+  helmValues:
+    appImage:
```
sengi, line 996: Looks like this line was accidentally left in, maybe? You don't need to specify `appImage` and the missing value here is causing a syntax error, I think. (Similarly in `values-test.yaml`.)
```suggestion
```

sengi (approved):

charts/argocd-apps/values-integration.yaml
```diff
@@ -991,6 +991,21 @@ applications:
...
+    extraEnv:
+      - name: SECRET_KEY_BASE
+        valueFrom:
+          secretKeyRef:
+            name: rails-secret-key-base
+            key: SECRET_KEY_BASE
+      - name: PUBLISHING_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-publisher-publishing-api
```
sengi, line 1007: Copy-pasto? Should be `signon-token-license-finder-publishing-api` presumably. (similarly in the other file)

NH10: Ahh right thanks will do

PR #294, opened 2022-03-24
Revert "Redirect if URL finishes with a trailing slash"

fredericfran-gds:
Reverts alphagov/govuk-helm-charts#287


fredericfran-gds:
it was a router issue actually.

PR #295, opened 2022-03-24
Actually give everything 2 replicas by default.

sengi:
D'oh. I missed this in 2ceab2f.

Thought about setting them all to 2 here, but decided it's just unnecessary boilerplate in integration. In production (and maybe staging to some extent) we'll want to specify these per app in most cases, but in integration the inverse is true - we generally don't care about horizontal capacity as long as the app is basically up most of the time.


fredericfran-gds (approved):
thanks. looks good

PR #296, opened 2022-03-24
Use HTTP probes for nginx and increase the timeouts a bit.

sengi:
The TCP probes were generating spurious "499" (client closed connection early) response codes in the nginx logs. Make the timeouts a bit more generous so that we don't unnecessarily kill containers when things get a bit slow. Chances are if one container is responding slowly, others are too, and we don't want to invite cascading failures.

Also increase the default timeouts on the app container.


fredericfran-gds (approved):
looks good thanks

samsimpson1 (approved):

PR #297, opened 2022-03-25
Fix redirect when there is a traling full stop in URL

fredericfran-gds:
This was taken from ec2 varnish: [ref](https://github.com/alphagov/govuk-puppet/blob/2383a96a9188d729340d97c7303c7409a18bdf6d/modules/varnish/templates/default.vcl.erb#L41)

Needs to be added to fastly too.


nsabri1 (approved):
LGTM assume been tested

PR #298, opened 2022-03-25
Add Mapit plek entry to redirect to mapit EC2

fredericfran-gds:
Mapit is used by the EKS platform but not yet deployed there.
Therefore, redirect mapit traffic to the EC2 instance.


nsabri1 (approved):
LGTM

PR #299, opened 2022-03-25
Add Plek entry for imminence & local-links-manager

fredericfran-gds:
These 2 apps are dependencies of frontend but are not spun
in EKS so redirecting to ec2 where they reside on the backend
machine.


nsabri1 (approved):
LGTM - wonder if we can create value which incorporates both {{ .Values.govukEnvironment }}.{{ .Values.ec2InternalDomainSuffix }}

PR #300, opened 2022-03-25
Add a `/readyz` to nginx for use by probes.

sengi:
Initially we'll use this instead of `/` (the homepage) for the Fastly healthcheck. (We'll probably want to terminate the Fastly healthcheck at the load balancer, ultimately, though.)

We'll also use it for the nginx container's liveness and readiness checks instead of misusing `/__canary__`.

`readyz` is probably the closest fit as regards a [conventional name](https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health) for this.

https://trello.com/c/RkVL0rMF/788


fredericfran-gds (approved):
nice

PR #301, opened 2022-03-25
Use `/readyz` for nginx container probes.

sengi:
This is preferable to misusing `/__canary__`, which is meant for use by a blackbox prober.

We'll want to return 404 for this from the load balancer, but there are a whole bunch of other paths that need the same treatment so we'll deal with those all at once later.

Deployment: needs #300 merged and rolled out first.


fredericfran-gds (approved):
thanks

sengi:
Rebased against main to pick up the changes to the pre-merge checks.

PR #302, opened 2022-03-25
Fix syntax error in nginx.conf.

sengi:
Introduced in 06ab177. Opportunity for a test here 😅

https://trello.com/c/RkVL0rMF/788


nsabri1 (approved):
LGTM

PR #308, opened 2022-03-28
Remove nonexistent permission from Licence Finder API user.

sengi:
This was causing the Signon API to crash when attempting to create the Authorisation.

Licence Finder doesn't currently have any additional permissions on its Authorisation for Publishing API. Publishing API doesn't have a permission called `internal_app` (or anything similar).

https://trello.com/c/F7vSHnws/841


fredericfran-gds (approved):
thanks

sengi:
Raised https://github.com/alphagov/signon/issues/1824 for the API bug. Not really a huge deal now that we know about it, but still would be nice if it gave a helpful error message (e.g. would have saved me maybe an hour of debugging).

PR #310, opened 2022-03-28
Create Smokey Asset-manager bearer token

nsabri1:
Adding Smokey bearer token which authenticates while testing GET of an asset on Asset-manger app. 

Link to code on Smokey which show the need of an bearer_token ENV to authenticate. https://github.com/alphagov/smokey/blob/db2af4e5803d167f006fabe7ca204f52a511620a/features/support/http_requests.rb#L88

Paired with @kengds


sengi (approved):
LGTM - the existing user (the one for EC2 Smokey) has an authorisation for Publishing API, but we checked that this was unused, hence only keeping the asset-manager one here 👍

PR #311, opened 2022-03-28
Prometheus switch for app monitoring only on EKS cluster

marcpomfret:
Added Environment variable to the apps config map to enable the monitoring on the EKS cluster. https://github.com/alphagov/govuk_app_config/pull/231
Cleaned up service ports from earlier testing that are not needed.

### DIFF
--- a/charts/govuk-apps-conf/templates/configmap.yaml
+++ b/charts/govuk-apps-conf/templates/configmap.yaml
@@ -13,6 +13,7 @@ data:
   GOVUK_ASSET_ROOT: https://assets.{{ .Values.publishingServiceDomainSuffix }}
   GOVUK_ENVIRONMENT: {{ .Values.govukEnvironment }}
   GOVUK_ENVIRONMENT_NAME: {{ .Values.govukEnvironment }}
+  GOVUK_PROMETHEUS_EXPORTER: "true"
   GOVUK_WEBSITE_ROOT: https://www.{{ .Values.externalDomainSuffix }}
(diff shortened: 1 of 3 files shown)


charts/govuk-apps-conf/templates/configmap.yaml
```diff
@@ -13,6 +13,7 @@ data:
   GOVUK_ASSET_ROOT: https://assets.{{ .Values.publishingServiceDomainSuffix }}
   GOVUK_ENVIRONMENT: {{ .Values.govukEnvironment }}
   GOVUK_ENVIRONMENT_NAME: {{ .Values.govukEnvironment }}
+  GOVUK_PROMETHEUS_EXPORTER: "{{ .Values.govukPrometheusExporter }}"
```
sengi, line 16: No need to put this in the configmap, because it's not really a global setting. It'd make more sense in the app charts (`govuk-rails-app`, `asset-manager` and any others that have been split off). I would just set it right in the Deployment template, in the `env` section.

sengi: Actually, sorry, scratch that - it's fine either way. I would just set it to true though rather than adding a value that we're never going to use.

sengi (approved):

charts/govuk-apps-conf/templates/configmap.yaml
marcpomfret: make sense, have changed to be directly in the configmap.

PR #313, opened 2022-03-28
Add 304 support to mixed case URLs

fredericfran-gds:
Mixed case URLs should return 404s with a cache of 30s.
When S3 sends the 404 response, it has the Etag and last-modified
headers set. This causes Fastly to issue a request with header
`if-none-match` or `if-last-modified` after the object expires
in the fastly node. This the 404 response has not changed,
a 304 should be sent rather than the 404 status code.

Refs:
1. [fastly doc](https://support.fastly.com/hc/en-us/community/posts/360041815632-304-Response-at-Fastly)
2. [nginx ifs evil doc](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)
3. [smokey test](https://github.com/alphagov/smokey/blob/db2af4e5803d167f006fabe7ca204f52a511620a/features/gov_uk.feature#L28)


sengi (changes requested):
I think it would be much safer and more correct to strip the etag and last-modified response headers from the error pages. (Just add `proxy_hide_header` directives for them below, just like we're already doing for the `x-amz-*` headers.)

fredericfran-gds:
Cool, I can do that, I was a bit worried it was not optimised on fastly side.

sengi:
I had a quick look at how Fastly handles 404s and it's basically just the usual semantics, so given that we're already setting `cache-control: public, max-age=30` on these error pages, we should be all good as long as we strip S3's `etag` and `last-modified` 👍

fredericfran-gds:
> I had a quick look at how Fastly handles 404s and it's basically just the usual semantics, so given that we're already setting `cache-control: public, max-age=30` on these error pages, we should be all good as long as we strip S3's `etag` and `last-modified` 👍

Yeah, it will work. it is just a bit inefficient in the sense that the object has not changed and the origin will resend it rather than confirming that the object has not changed. Not a biggie in this case.

sengi (approved):

sengi:
Nice job finding the issue here btw! Thanks!

PR #321, opened 2022-03-29
Make MongoDB connection strings match cluster nodenames.

sengi:
When a Mongo client connect to a cluster (replicaset), the client library connects to one or more of the hosts given in the connection string and requests list of nodes. It then tries to connect to the hostnames in that list. The client libraries that we use seem to remove the original hostnames (the ones in the connection string) from the list of nodes that they'll connect to in future.  In some cases, this leads to the client library being unable to reconnect because it's discarded all the reachable hosts from the connection string and replaced them with unreachable/unresolveable hosts that the cluster told it about.

The `mongoid` client library that we use in our Ruby apps seem to be more brittle regarding this issue than the `mgo` library in Router, but in both cases it's a good idea for stability to have the hostnames in the connection string match what the cluster returns.

Annoyingly, the Mongo clusters are not configured consistently in this regard: the Router one (on router_backend machines) has short hostnames as nodenames whereas the other Mongo cluster uses FQDNs (without the "stackname" component, i.e. `blue`).

We include the stackname where specifying the additional search suffix for Router, because we might as well save a DNS query. The `blue` thing isn't going to change in the remaining lifetime of the GOV.UK EC2 stack.

This should fix some stability issues with the publisher apps that we're running so far in k8s.

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -236,6 +236,9 @@ applications:
       requests:
         cpu: 2
         memory: 1Gi
+    dnsConfig:  # TODO: remove dnsConfig once MongoDB migrated to DocDB.
+      searches:
+        - blue.integration.govuk-internal.digital
     ingress:
       enabled: true
       annotations:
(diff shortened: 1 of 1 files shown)


marcpomfret (approved):
LGTM

PR #324, opened 2022-03-30
Kubescape account ID no longer required as an external secret

lwilts:
Cleaning up - this secret is no longer required, as it now gets retrieved directly from AWS by Terraform.


sengi (approved):
Cool, thanks for cleaning this up. You'll need to rebase in order for the checks to pass because I renamed the "test" one to "Chart testing" and GitHub doesn't have a way to figure that out.

PR #325, opened 2022-03-30
Add missing headers to assets

fredericfran-gds:
Some assets miss the some headers as in EC2 mentioned:
[1](https://github.com/alphagov/govuk-puppet/blob/77f85be19ceed239f6d652cc07207266b165fb9f/modules/router/templates/assets_origin.conf.erb#L36)
and [2](https://github.com/alphagov/govuk-puppet/blob/77f85be19ceed239f6d652cc07207266b165fb9f/modules/govuk/templates/asset_pipeline_extra_nginx_conf.erb#L2)


nsabri1 (approved):
LGTM

PR #326, opened 2022-03-31
Freshclam was not fetching latest virus signatures.

nsabri1:
Added extra container to the worker pod  which run `freshclam -d` and updates virus db in a `emptyDir` volume mount  @/var/lib/clamav. Mount is also mounted on the worker pod which runs Clamscan.

This whole approach needs to be updated in the near future, somewhat explained in - https://trello.com/c/uIygWIgc/789-dockerized-clamdscan-clamscan-for-scanning-assets

https://trello.com/c/02lrQ2So/881-freshclam-virus-updater-for-asset-manager

### DIFF
--- a/charts/asset-manager/templates/worker-deployment.yaml
+++ b/charts/asset-manager/templates/worker-deployment.yaml
@@ -45,6 +45,20 @@ spec:
             allowPrivilegeEscalation: false
           # TODO: consider implementing a liveness probe for Sidekiq
           # e.g. https://github.com/arturictus/sidekiq_alive
+        - name: clamdb
+          image: "{{ required "A valid .Values.appImage.repository entry required!" .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          command: ["/bin/sh"]
+          args: ["-c", "freshclam -d"]
+          resources:
+            requests:
+               memory: "64Mi"
+               cpu: "250m"
+            limits:
+              memory: "128Mi"
+              cpu: "500m"
+          volumeMounts:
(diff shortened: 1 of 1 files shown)


charts/asset-manager/templates/worker-deployment.yaml
```diff
@@ -45,6 +45,20 @@ spec:
             allowPrivilegeEscalation: false
           # TODO: consider implementing a liveness probe for Sidekiq
           # e.g. https://github.com/arturictus/sidekiq_alive
+        - name: clamdb
+          image: "{{ required "A valid .Values.appImage.repository entry required!" .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          command: ["/bin/sh"]
+          args: ["-c", "freshclam", "-d"]
```
fredericfran-gds, line 51: I would put freshclam -d in a single parameter, i.e. args: ["-c", "freshclam -d"]

Ref: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#run-a-command-in-a-shell

fredericfran-gds (approved):
looks good, just one suggestion

charts/asset-manager/templates/worker-deployment.yaml
nsabri1: i guess both approaches will have equal affect https://linux.die.net/man/1/bash. i will update thanks

nsabri1: updated

PR #327, opened 2022-03-31
Add securityContext to rails-app deployment

marcpomfret:
Adds securityContext to rails-app deployment to run containers as not root user


alexbakervr (approved):
LGTM

PR #332, opened 2022-03-31
Updating Freshclam worker

nsabri1:
Also and adding mount to sidekiq worker - resources needed to be increased due to OOM

https://trello.com/c/02lrQ2So/881-freshclam-virus-updater-for-asset-manager


fredericfran-gds (approved):
ok

PR #342, opened 2022-04-04
Updated pid to /tmp

nsabri1:
As we are now using nginx-unprivileged image, updated pid to /tmp/nginx.pid

### DIFF
--- a/charts/asset-manager/templates/nginx-configmap.yaml
+++ b/charts/asset-manager/templates/nginx-configmap.yaml
@@ -10,7 +10,7 @@ data:
     user nginx;
 
     error_log /dev/stderr warn;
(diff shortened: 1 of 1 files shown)


fredericfran-gds (approved):
ok

sengi (approved):
It baffles me that nginx doesn't appear to have a way to just not write the pidfile, but that really does seem to be the case 🤷 LGTM

sengi:
Might also be worth checking whether any of [the other changes](https://github.com/nginxinc/docker-nginx-unprivileged#nginx-unprivileged-docker-image) to the default configfile between the root and non-root containers are applicable, for example removing the `user` directive or pointing `*_temp_path` at `/tmp`.

PR #348, opened 2022-04-04
Allow aws-lb-controller to read OAuth secrets for ALB authentication.

sengi:
Also combine the Prometheus and Alertmanager ExternalSecrets for their ALB OIDC authentication into one template file so that it's easier to see what's going on, now that there's a Role and RoleBinding to go along with them.

Considered putting all these in separate files, but that way it'd be much harder to see what's going on.

[Trello](https://trello.com/c/gipscM7t/873)


lwilts (approved):
LGTM

PR #357, opened 2022-04-04
Update config to use values from values.yaml and release name.

nsabri1:
After some intial vows realised the Nginx config was using wrong ports within the config.

I have updated both Router and Asset manager.

It maybe that Asset manger Nginx config will require further updates -
which will be apparent on asset upload testing and smokey test.


sengi (approved):
Oof, thanks for fixing!

PR #359, opened 2022-04-05
Reverting nginx conf change for Router

nsabri1:
This change will not generate correct configmap for Router app, as it is generated by govuk-app-conf and not router - therefore release name will be **_govuk-app-conf_** and not Router. likewise ports will not be populated as there is no defined appPort or nginxPort.


fredericfran-gds (approved):
ok, good. I think it was currently working before so....

PR #364, opened 2022-04-07
Add Helm values for ArgoCD apps in the staging environment. 

sengi:
We only want to run the frontend apps (and Signon) in the staging environment initially. We'll bring in the publisher apps and draft stack etc. later, once they're fully configured and working properly. The main reason for this is that we want staging to be a reasonably accurate model of production, and we wouldn't want to deploy a bunch of half-working apps to production (even if we wouldn't be sending traffic to them).

### DIFF
--- a/charts/argocd-apps/ci/staging-values.yaml
+++ b/charts/argocd-apps/ci/staging-values.yaml
@@ -1,0 +1,1 @@
+../values-staging.yaml
--- a/charts/argocd-apps/values-staging.yaml
+++ b/charts/argocd-apps/values-staging.yaml
@@ -63,13 +63,13 @@ applications:
   helmValues:
     extraEnv:
       - name: GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID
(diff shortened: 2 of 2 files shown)


charts/argocd-apps/ci/staging-values.yaml
```diff
@@ -0,0 +1 @@
+../values-staging.yaml
```
fredericfran-gds, line 1: why do we need this relative path?

fredericfran-gds (approved):
ok

charts/argocd-apps/ci/staging-values.yaml
sengi: It's a symlink. [chart-testing](https://github.com/helm/chart-testing) is a bit inflexible about which values files it tests against; it forces you to put the files to test against in a directory called `ci` and only matches filenames of the pattern `*-values.yaml`. So in order to get it to test that the values files work with the chart, we have to symlink them here. Maybe someday we can upstream a fix so that we can just tell chart-testing which values files to use.

PR #369, opened 2022-04-08
Remove podMonitor from fastly-exporter Helm chart

fredericfran-gds:
Mistakenly commited the podMonitor for fastly-exporter.
We now use a serviceMonitor which is more resilient since
we can have multiple pods scraping.


nsabri1 (approved):
LGTM

PR #372, opened 2022-04-11
Grafana fastly dashboard

fredericfran-gds:
Add Grafana dashboards as configmaps, first dashboard is the fastly one.

The Grafana configmaps are deployed in a Helm chart which is then deployed by Argo.

Hence, we introduce 2 categories of apps:
 - govukApplications (renamed from applications)
 - monitoringApplications


nsabri1 (approved):
LGTM

PR #374, opened 2022-04-11
GDS-669: Security context to enforce NonRoot app containers

desenhar:

We are testing collections app to run as non root, using explicit UID, also we have changes to temp path variables to fix the unprivileged nginx image permission errors.


charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -15,8 +15,6 @@ spec:
         {{- include "govuk-rails-app.labels" . | nindent 8 }}
         app: {{ .Release.Name }}
     spec:
-      securityContext:
-        runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
```
nsabri1, line 19: You need to keep this

charts/govuk-rails-app/templates/nginx-configmap.yaml
```diff
@@ -22,6 +22,12 @@ data:
     http {
       include       /etc/nginx/mime.types;
       default_type  application/octet-stream;
+      client_body_temp_path /tmp/client_temp;
+      proxy_temp_path       /tmp/proxy_temp_path;
+      fastcgi_temp_path     /tmp/fastcgi_temp;
+      uwsgi_temp_path       /tmp/uwsgi_temp;
+      scgi_temp_path        /tmp/scgi_temp;
+
```
nsabri1, line 30: Not sure if this extra line is needed

nsabri1 (changes requested):

charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         app: {{ .Release.Name }}
     spec:
       securityContext:
-        runAsNonRoot: {{ .Values.securityContext.runAsNonRoot }}
```
nsabri1, line 19: I would not make any changes to this yet until we tested - as if we set it to **true** now it will be applied everywhere this chart is used. suggest you leave this as it for now.

nsabri1 (changes requested):

nsabri1 (approved):
LGTM - Thanks for this.

PR #379, opened 2022-04-12
Updating default chart to enable Non-Root containers and UID

nsabri1:
This change will update both App and Worker containers.

    adds UID and GID 1001


desenhar (approved):
lgtm

PR #384, opened 2022-04-13
Prevent returning error when image tag isn't promoted

theseanything:
This updates the update image tag script to successfully return instead of an error. This was confusing because the script has indeed followed the correct behaviour and not promoted the non latest release commit.


fredericfran-gds (approved):
Thanks. As discussed on Slack, probably we need to check the actual tag rather than the commit history but this is for another time. Thanks.

PR #386, opened 2022-04-19
Removing as flagged as a vulnerability in kubescape

alexbakervr:


charts/argocd-apps/values-integration.yaml
```diff
@@ -685,8 +685,6 @@ govukApplications:
         value: govuk-publishing-api-event-log-integration
       - name: EVENT_LOG_AWS_ACCESS_ID
         value: unset  # TODO: Set up Event Log S3 Bucket (not in terraform)
```
sengi, line 687:
```suggestion
        value: unset  # TODO: allow publishing-api to write to the event log S3 bucket (using IAM permissions, not long-lived keys like this; this env var should be deleted)
```

sengi (approved):
Cool, thanks for silencing this warning. I guess it's served its purpose in that it's flagged that we shouldn't be using long-lived AWS creds like this in the first place, which is really what the TODO is saying.

Suggested a clarification to the existing TODO.

PR #388, opened 2022-04-20
PSP-Baseline is now active therefore we need to update Default Helm chart

nsabri1:
govuk-rails-app Helm chart needs updating.

PSP-Baseline will not allow new root containers from running.

I have updated all Deployments within this chart to run as non root and those containers which
use the appImage now run with UID 1001 and GID 1001.

Untested at the moment


desenhar (approved):
LGTM

fredericfran-gds (approved):
ok, might be time to re-introduce the default value true, I know that there are some issues with the way helm is interpreting that value.

PR #389, opened 2022-04-20
Implement network-based ACL for `www-origin.staging`

sengi:
Trusting the network is not [good practice][1], but in this case improving on this pre-existing arrangement is nontrivial and not in scope for the Replatforming project, so we're just keeping the status quo for now. In other words, we're replicating the same access control that we already have on the current, EC2 staging environment.

Ingresses for publisher apps (including Signon) are excluded from this control because that's how it is in EC2 staging (presumably because Signon is the point of access control for the non-public-facing stuff). We'll need to apply this allow-list to assets-origin when we write the config for that. (assets-origin doesn't exist yet in our k8s rig.)

Tested: inspected the template output (`helm template ../govuk-rails-app --values <(helm template . --values values-staging.yaml |yq e '.|select(.metadata.name=="router").spec.source.helm.values')`)

[1]: https://cloud.google.com/beyondcorp


nsabri1 (approved):
LGTM

fredericfran-gds (approved):
looks good, don't forget fastly :-)

sengi:
Decided to add the Fastly netblocks here after all (thanks @fredericfran-gds for reminding me!) This way at least I won't be leaving it broken in the meantime. Added a TODO with a possible way to avoid the inevitable bitrot on the list of Fastly networks.

PR #390, opened 2022-04-21
Add chart for argo workflows RBAC resources

samsimpson1:
This adds a chart which contains the necessary resources to enable RBAC in Argo Workflows. 


charts/argo-services/templates/argo-workflows-rbac/role.yaml
```diff
@@ -0,0 +1,52 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
fredericfran-gds, line 1: can you add a reference as a comment if you got the resources list from somewhere?

charts/argo-services/values.yaml
```diff
@@ -1,3 +1,6 @@
 govukEnvironment:
 argocdUrl:
 workflowsNamespace: apps
+rbacTeams:
+  read_only: gov-uk
```
fredericfran-gds, line 5: I wonder if we should have a list in anticipation of the 3rd team `gov-uk-deployers`, maybe premature.

fredericfran-gds (approved):
Looks good, thanks.

fredericfran-gds (changes requested):
On 2nd pass, I need a modification for the second comment of list of auth teams per categories of users

charts/argo-services/templates/argo-workflows-rbac/service-account.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: argo-workflows-read-only
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    workflows.argoproj.io/rbac-rule: "'{{ .Values.rbacTeams.read_only }}' in groups"
+    workflows.argoproj.io/rbac-rule-precedence: "1"
```
fredericfran-gds, line 8: I wonder if the precedence should be swapped?

e.g. Person A is in gov-uk and gov-uk-production teams, they should be in read-write (admin) but now they will have a read-only role?

charts/argo-services/values.yaml
fredericfran-gds: On 2nd pass, I suggest a modification for the second comment of list of auth teams per categories of users. The reason for that is in integration, gov-uk and gov-uk-production will be read-write. I know gov-uk is supposed to be a super-set of gov-uk-production but it might not happen in practice?

charts/argo-services/templates/argo-workflows-rbac/service-account.yaml
samsimpson1: Counter-intuitively, the higher the number the higher the precedence with argo workflows. So if the read-write rule has a higher number than the read-only, it will be selected over the read-only one.

charts/argo-services/values.yaml
samsimpson1: It would be possible to make it take a list of groups, but it would add complexity. It may be easier to just make sure the team memberships are correct in github

charts/argo-services/templates/argo-workflows-rbac/service-account.yaml
fredericfran-gds: Thanks for that.

fredericfran-gds (approved):
thanks for the explanations, good work!

PR #391, opened 2022-04-21
Remove long-lived AWS creds for Asset Manager.

sengi:
We don't want to be using these any more; they're a hold-over from years ago when GOV.UK was hosted outside AWS. We've already configured IAM policies in Terraform to authorise Asset Manager to its S3 buckets, so let's remove these keys in order to check that the IAM stuff is working as intended.

Rollout: need to remember to delete the corresponding secrets from AWS Secrets Manager in the integration account after merging.


sengi:
Ah, turns out the IAM stuff doesn't work yet (owing to probable bug in the app). Can't merge this just yet then.

sengi:
We're now on an up-to-date version of the AWS client libs and this now works. RFAL.

nsabri1 (approved):
LGTM

theseanything (approved):

PR #392, opened 2022-04-21
Improve Fastly dashboard

fredericfran-gds:
1. Make sure all the units are present
2. Rename `rate` to `window` and convert it into a hidden constant

Updated Fastly Grafana dashboard


nsabri1 (approved):
LGTM

PR #393, opened 2022-04-21
Use ClusterRoleBindings instead of RoleBindings for argo workflows service accounts

samsimpson1:
This allows users to access workflows in all namespaces instead of just the cluster-services one


fredericfran-gds (approved):
ok...

PR #394, opened 2022-04-21
[wip] gds-669: Podsecuritypolicy now binds to default and apps namespaces.

desenhar:
Tested, with a simply nginx deployment in the default namespace that has a Hostnetwork:true parameter it fails.
 
 Warning  FailedCreate  0s (x13 over 21s)  replicaset-controller  Error creating: pods "nginx-hostnetwork-deployment-76c46fdb6-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used]


marcpomfret (approved):
LGTM

PR #400, opened 2022-04-25
Fix erroneous use of Publisher secrets in Whitehall.

sengi:
The ExternalSecrets for these already exist, as do the Secrets Manager secrets backing them. I checked that the secret values are correct for integration and staging and fixed one of the them. They're now all in sync with Puppet/EC2.


nsabri1 (approved):
LGTM

PR #401, opened 2022-04-25
Remove disused COMPANIES_HOUSE_API_KEY secret.

sengi:
The app (Smart Answers) no longer uses this. See https://github.com/alphagov/govuk-puppet/pull/11564.


fredericfran-gds (approved):
LGTM

PR #403, opened 2022-04-28
Change param name from manualCommit to manualDeploy

theseanything:
This updates the name of the parameter responsible for specifying if the deployment was manually triggered. Using the terminology to deploy to make it more clear that it's about a manual deployment rather than a git commit.


sengi (approved):
Thanks for improving the terminology ❤️  This sounds much clearer to me.

PR #405, opened 2022-04-29
Replace resources list with wildcard for argo workflows user roles

samsimpson1:
Some resources were missed in the initial list, so this just allows access to all resources in the argoproj.io API group.


fredericfran-gds (approved):
ok

PR #409, opened 2022-05-03
Added privileged policy and the namespaces it'll be binded to.

desenhar:
Adding a helm chart for the privileged policy, that will be deployed a tf helm release.

### DIFF
--- a/charts/cluster-security/README.md
+++ b/charts/cluster-security/README.md
@@ -1,0 +1,11 @@
+## Cluster Security
+
+
+This chart is installed by the cluster-services govuk-infrastructure
+Terraform deployment.
+
+# what does it do?
+
+This chart installs two pod security policies. A baseline policy with moderate restrictions applied only to 'default' and 'apps' namespaces, and a privileged policy with no restrictions applied to all other namespaces in the cluster.
+
+The policy enforces the service accounts in charge of deploying pods (via deployments,rs,sts,ds) abide by the restrictions outlined in the respective policy.
--- a/charts/cluster-security/README.md
+++ b/charts/cluster-security/README.md
@@ -4,7 +4,7 @@
 This chart is installed by the cluster-services govuk-infrastructure
 Terraform deployment.
 
-# what does it do?
+# What does it do?
 
 This chart installs two pod security policies. A baseline policy with moderate restrictions applied only to 'default' and 'apps' namespaces, and a privileged policy with no restrictions applied to all other namespaces in the cluster.
 
--- a/charts/cluster-security/templates/bindings.yaml
+++ b/charts/cluster-security/templates/bindings.yaml
@@ -26,7 +26,7 @@ subjects:
   name: system:serviceaccounts:{{ . }}
   apiGroup: rbac.authorization.k8s.io
 - kind: Group
-  name: system:cluster-admins
+  name: cluster-admins
   apiGroup: rbac.authorization.k8s.io
 roleRef:
   kind: ClusterRole
--- a/charts/cluster-security/templates/bindings.yaml
+++ b/charts/cluster-security/templates/bindings.yaml
@@ -1,30 +1,36 @@
-# This role binding allows all pods in the 'default' and 'apps' namespace to use the baseline PSP.
-apiVersion: rbac.authorization.k8s.io/v1
+ {{- range .Values.baselineNamespaces }}
 kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: psp-baseline-rolebinding
-  namespace: apps
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
   name: psp-baseline
+  namespace: {{ . }}
 subjects:
 - kind: Group
-  name: system:authenticated
-  namespace: apps
+  name: system:serviceaccounts:{{ . }}
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
+  name: psp-baseline
   apiGroup: rbac.authorization.k8s.io
 ---
-apiVersion: rbac.authorization.k8s.io/v1
+{{ end }}
+---
+{{- range .Values.privilegedNamespaces }}
 kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: psp-baseline-rolebinding
-  namespace: default
-roleRef:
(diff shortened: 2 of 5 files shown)


charts/cluster-security/templates/roles.yaml
```diff
@@ -1,13 +1,27 @@
-apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: psp-baseline
-  labels:
-    kubernetes.io/cluster-service: "true"
-    eks.amazonaws.com/component: pod-security-policy
+  name: psp-baseline-clusterrole
```
sengi, line 4: nit: the `-clusterrole` suffix is redundant since these names are already namespaced by kind. (similarly for the other one below)

charts/cluster-security/values.yaml
```diff
@@ -0,0 +1,2 @@
+baselineNamespaces: [apps, default]
+privilegedNamespaces: [armo-system, cluster-services, kube-node-lease, kube-public, kube-system, kubescape-host-scanner, monitoring]
```
sengi, line 2: nit: consider using the YAML syntax here

sengi (commented):
Looks mostly fine on paper. Couple of important questions though:

* What are the main things that this is intended to enforce? (I think I know what they are, but others - including our future selves - might not.)
* How did you test this?

sengi:
Oh also please don't forget to squash all the fixup commits before merging. (It's best to clean up the commit history before opening the PR, ideally, since it helps the reviewer.)

charts/cluster-security/templates/bindings.yaml
```diff
@@ -1,30 +1,35 @@
...
-  namespace: default
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: psp-baseline
+  name: psp-privileged-rolebinding
+  namespace: {{ . }}
 subjects:
 - kind: Group
   name: system:authenticated
```
sengi, line 27: Sorry didn't spot this earlier - it's best not to use `system:authenticated` here. The [docs](https://kubernetes.io/docs/concepts/security/pod-security-policy/#recommended-practice) recommend authorising just all service accounts in the namespace. (I'd also include cluster admins so that we don't break things like `kubectl run`, which we'll likely need for troubleshooting at least in the short term.)

charts/cluster-security/values.yaml
desenhar: I copied the same syntax from cluster-secrets/values.yaml.
So something in the form of

```namespaces:
 - apps
 - default
```
 would be more appropriate?

charts/cluster-security/templates/bindings.yaml
```diff
@@ -1,30 +1,38 @@
...
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
   name: psp-baseline
+  namespace: {{ . }}
 subjects:
 - kind: Group
-  name: system:authenticated
-  namespace: apps
+  name: system:serviceaccounts
```
sengi, line 9: It seems odd to be specifying `system:serviceaccounts` (i.e. all service accounts in all namespaces) in a RoleBinding, which is inherently specific to a namespace. Is that intended? Did you mean to include the namespace here?

sengi: Alternatively it might be simpler to use ClusterRoleBinding for some of this.

desenhar: From [this](https://kubernetes.io/docs/concepts/security/pod-security-policy/),  it advises using a rolebinding to grant usage for pods being run in the same namespace as the binding. When I tested with a clusterolebinding it didn't work properly. I'll drop the namespace argument here in the subjects since it's specified in the metadata of the rolebinding.

sengi: It should work with either - the docs that you linked to have examples with RoleBinding and ClusterRoleBinding.

desenhar: The clusterrolebinding looks straightforward enough, but when it came to testing whether the psp would restrict deployments of say for example a root container deployment, it just wouldn't work.
Did some reading and its fine to apply regular RoleBindings to ClusterRoles.
I can look to testing clusterrolebindings again, but at the moment the regular rolebindings work.

sengi: Like I said, either should work. I don't mind whether we use ClusterRoleBinding or RoleBinding here; I'm just trying to ensure that the policy intent is clear. So if we have a RoleBinding it'd make sense to have the subject be `system:serviceaccounts:{{ . }}` explicitly (i.e. the namespace's service accounts) rather than specifying all service accounts but then relying on the fact that it'll be implicitly restricted to principles in the namespace of the RoleBinding. Just makes it that much easier to understand - and by gosh this area is complicated enough already 😅

charts/cluster-security/templates/bindings.yaml
```diff
@@ -1,30 +1,36 @@
...
+  name: psp-privileged
+  namespace: {{ . }}
 subjects:
 - kind: Group
-  name: system:authenticated
-  namespace: default
+  name: system:serviceaccounts:{{ . }}
   apiGroup: rbac.authorization.k8s.io
+- kind: Group
+  name: system:cluster-admins
```
sengi, line 29: Are you sure this group exists? (i.e. does it actually contain any principals?)

charts/cluster-security/README.md
```diff
@@ -0,0 +1,11 @@
+## Cluster Security
+
+
+This chart is installed by the cluster-services govuk-infrastructure
+Terraform deployment.
+
+# what does it do?
```
sengi, line 7: nit: capital letter at start of heading?

sengi (approved):
You might need to check that you're actually including the cluster admin roles in your RoleBindings there, but otherwise ok.

charts/cluster-security/templates/bindings.yaml
desenhar: I checked `k get -n kube-system configmap/aws-auth -o yaml`
There is a "cluster-admins" group that is populated with gds team, but it might not need the "system:" prefix.
I'll change it.

sengi: Cool, thanks. Yeah it's a different group with the `system:` prefix. The string has to match.

PR #412, opened 2022-05-06
Updating Router nginx configmap to use /tmp

nsabri1:
Router's Nginx container is currently throwing error:
nginx: [emerg] mkdir() "/var/cache/nginx/client_temp" failed (30: Read-only file system)

Previously we had applied a similar change to default helm chart Nginx config:
https://github.com/alphagov/govuk-helm-charts/commit/7fd8acfb80171acf8b36528d8bf52b4379493c8d

https://trello.com/c/1IEBYxAt/917-add-volume-mount-for-nginx

### DIFF
--- a/charts/govuk-apps-conf/templates/router-configmap.yaml
+++ b/charts/govuk-apps-conf/templates/router-configmap.yaml
@@ -22,6 +22,9 @@ data:
       include       /etc/nginx/mime.types;
       default_type  application/octet-stream;
 
+      client_body_temp_path /tmp/client_temp;
+      proxy_temp_path       /tmp/proxy_temp_path;
+
       server_tokens off;
 
       sendfile        on;


charts/govuk-apps-conf/templates/router-configmap.yaml
```diff
@@ -22,6 +22,12 @@ data:
       include       /etc/nginx/mime.types;
       default_type  application/octet-stream;
 
+      client_body_temp_path /tmp/client_temp;
+      proxy_temp_path       /tmp/proxy_temp_path;
+      fastcgi_temp_path     /tmp/fastcgi_temp;
```
sengi, line 27: optional: It shouldn't use these last three anyway (at least in theory) but it doesn't hurt to set them. Up to you if you wanna leave 'em in.

sengi (approved):
🚀

charts/govuk-apps-conf/templates/router-configmap.yaml
nsabri1: thanks for pointing that out. I will remove them and see if its happy - as no need to have unneeded config.

nsabri1: will remove
 ```
 fastcgi_temp_path     /tmp/fastcgi_temp;
       uwsgi_temp_path       /tmp/uwsgi_temp;
       scgi_temp_path        /tmp/scgi_temp;
```

PR #417, opened 2022-05-10
Use `kubernetes.io/description` instead of `a8r.io/description`.

sengi:
[`kubernetes.io/description`](https://kubernetes.io/docs/reference/labels-annotations-taints/#description) was recently introduced as a well-known annotation, so let's use it in favour of `a8r.io/description`.

Generated with `find . -type f -name '*.yaml' -exec gsed -i 's/a8r.io\/description/kubernetes.io\/description/g' {} \;`


nsabri1 (approved):
LGTM

PR #423, opened 2022-05-13
Add Nginx Prometheus exporter

fredericfran-gds:
Adds a Nginx Prometheus exporter to govuk-rails-apps helm chart
as a trial to see if it can replace the Rails Prometheus exporter.

Ref:
1. [nginx log prometheus exported](https://github.com/martin-helmich/prometheus-nginxlog-exporter)


sengi:
We definitely can't replace the Rails exporter (it serves a different purpose) but it'd certainly be cool to have nginx metrics!

sengi:
Oh, it works by tailing the logs, right? That's not so great :(

fredericfran-gds:
Yeah, probably envoy/forking out for nginx plus. I will close this.

sengi:
Thanks for trying this out anyway!

PR #424, opened 2022-05-13
Fix kubeconform

fredericfran-gds:
By default, kubeconform uses master directory of k8s spec in
[here](https://github.com/yannh/kubernetes-json-schema).

We currently have a podsecuritypolicy resource which is deprecated
in k8s 1.25 and kubeconform will fail if validated against that
version

Hence, we now set the kubeconform to use the same version of the k8s
that  we use in AWS EKS, i.e. 1.21 currently


nsabri1 (approved):
Thanks for this , LGTM

sengi:
Great catch, thanks for fixing this!

PR #427, opened 2022-05-16
Update third-party GitHub Actions + style cleanup.

sengi:
Update the checkout and chart-testing-action actions to the latest releases.

Remove a step ID that we're no longer using.

Trivial style cleanup re multiline strings.

Tested: check passes on this PR. (Workflow definition comes from the PR branch.)


alexbakervr (approved):
LGTM

PR #440, opened 2022-05-19
Fix nextEnvironment value for staging

theseanything:
This fixes the value the specifies the next environment to promote a release to. Staging promotes to production, otherwise it'd try to update the image tag for the same environment.


kentsanggds (approved):

sengi:
Haha whoops 😅 Good catch and thanks for fixing!

PR #441, opened 2022-05-23
Send deployment annotations to Grafana

theseanything:
This enable Argo Notifications to send annotation events to Grafana. The annotations represent when Argo Application (for GOV.UK apps) sync successfully. Whilst a sync can occur for other reasons than a deployment, it's the closest event to map to a deployment (and it's useful when other application state might change). Annotations are tagged with the Argo Application name and "deployment". This also updates the "App: request rates, errors, durations" Grafana dashboard to display the annotations for the selected app.


kentsanggds (approved):
LGTM

PR #446, opened 2022-05-25
Empty appProbe does not need to be defined here.

nsabri1:


nsabri1:
> I presume this is intended to be a no-op cleanup - is that right? I had to read the template in order to check, and I think it shouldn't affect the output. (Did you check using `helm template`?)

yes no-op cleanup no impact - as appProbe is for custom app probes - but we are not defining any custom probe here so its unneeded

PR #449, opened 2022-05-26
Re-label container to app name

marcpomfret:
Removing the app name from the container name as part of [#444](https://github.com/alphagov/govuk-helm-charts/pull/444) removed the ability to filter by app in grafana this allows that filter to work again

### DIFF
--- a/charts/govuk-rails-app/templates/podMonitor.yaml
+++ b/charts/govuk-rails-app/templates/podMonitor.yaml
@@ -7,7 +7,7 @@ metadata:
     app.kubernetes.io/name: prometheus-operator
     app.kubernetes.io/part-of: kube-prometheus
   name: {{ .Release.Name }}
-  namespace: monitoring
+  namespace: apps
 spec:
   selector:
     matchLabels:


fredericfran-gds (approved):
looks good. Can you change asset manager PodMonitor too. Thanks.

sengi (commented):
Doesn't this mean we end up with a label called "container" that really represents the pod and not the container? Shouldn't we fix the dashboard instead?

sengi:
I'm suspicious that the dashboard queries refer to `container`, e.g. `sum(rate(http_requests_total{container="${app}"}[1m]))`. Pretty sure that should be `job` instead of `container`. Also it seems we have another bug: the job labels are all prefixed with `monitoring/` for some reason. (Surely it should be `app/` if it's meant to be the namespace?)

marcpomfret:
> Doesn't this mean we end up with a label called "container" that really represents the pod and not the container? Shouldn't we fix the dashboard instead?

yeah see where your coming from, how about if i change it to app_name, as pod is already there with the extra -xxxxx bits on that i'd been trying to get off with regex, dont really want to overwrite that as we'll prob want them down the line

fredericfran-gds:
> I'm suspicious that the dashboard queries refer to `container`, e.g. `sum(rate(http_requests_total{container="${app}"}[1m]))`. Pretty sure that should be `job` instead of `container`. Also it seems we have another bug: the job labels are all prefixed with `monitoring/` for some reason. (Surely it should be `app/` if it's meant to be the namespace?)

Maybe  I should add the namespace filter, I already created the variable in the dashboard but for some reason, I seem to have removed them from the query. Sorry.

marcpomfret:
ok updated the namespace to be apps instead of monitoring as discussed and then will change the grafana board separately

PR #451, opened 2022-05-26
Simplify PodMonitors

fredericfran-gds:
Simplify PodMonitors by removing the namespace fields so that
the PodMonitor will find the endpoint to scrape base on the
`current` namespace of the PodMonitor.

### DIFF
--- a/charts/asset-manager/templates/podMonitor.yaml
+++ b/charts/asset-manager/templates/podMonitor.yaml
@@ -7,13 +7,9 @@ metadata:
     app.kubernetes.io/name: prometheus-operator
     app.kubernetes.io/part-of: kube-prometheus
   name: {{ .Release.Name }}
-  namespace: monitoring
 spec:
   selector:
     matchLabels:
       app: {{ .Release.Name }}
-  namespaceSelector: 
-    matchNames: 
-    - apps
   podMetricsEndpoints:
   - port: metrics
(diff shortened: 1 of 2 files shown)


charts/asset-manager/templates/podMonitor.yaml
```diff
@@ -7,13 +7,9 @@ metadata:
     app.kubernetes.io/name: prometheus-operator
     app.kubernetes.io/part-of: kube-prometheus
   name: {{ .Release.Name }}
-  namespace: monitoring
 spec:
   selector:
     matchLabels:
       app: {{ .Release.Name }}
-  namespaceSelector:
```
sengi, line 15: Is the default behaviour equivalent to this (i.e. only match stuff in the apps namespace because that's the namespace that the PodMonitor's in), or does it default to matching all namespaces? If it's all namespaces then that's problematic because it means the metrics will get polluted if there's ever an app with the same name running in another namespace.

sengi (approved):

charts/asset-manager/templates/podMonitor.yaml
sengi: (I tried to look this up, but the API docs for namespaceSelector weren't very helpful on this point.)

sengi:
Thanks for the cleanup!

charts/asset-manager/templates/podMonitor.yaml
fredericfran-gds: I think it is ok, based one: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#namespaceselector

PR #453, opened 2022-05-27
Prevent crawling of draft stack and whitehall-admin

samsimpson1:
This adds a header to draft router and whitehall-admin which should stop search engines from indexing pages on these services. This also puts the router nginx config in its own file, instead of being in-line in the config map.

Tested both of the modified charts with helm template, with different variations of govukEnvironment and nginxDenyCrawlers.

Trello: https://trello.com/c/QRusI25c/857-ensure-no-indexing-of-draft-stack-and-other-domains


fredericfran-gds (approved):
looks good, I wonder if we need to prepend `live` and `draft` to the config map names rather than append to keep the convention?

PR #454, opened 2022-05-27
Fix router not updating due to incorrect volume mount name

samsimpson1:
This was missed in the original PR.

Trello: https://trello.com/c/QRusI25c/857-ensure-no-indexing-of-draft-stack-and-other-domains


fredericfran-gds (approved):
ok

PR #455, opened 2022-05-30
Fix router Prometheus metrics

fredericfran-gds:
Prometheus can't currently scrape router metrics on port 9394.
We need to set the api_port to 9394 in router via an env var.
Note that the api_port will be use only for metrics and
healthchecks and it is no longer used by router_api to trigger
router to reload routes in its memory.

Ref:
1. [trello card](https://trello.com/c/ozKdr7Et/933-fix-scraping-of-router-metrics)


kentsanggds (approved):
LGTM

PR #457, opened 2022-05-31
Mark assets as public and immutable to aid caching

samsimpson1:
This marks assets as public and immutable, meaning they can be cached indefinitely, will never change and can be stored in public caches.

Trello: https://trello.com/c/uCGahIqu/623-custom-response-headers-at-origin


nsabri1 (approved):
ok

PR #462, opened 2022-06-14
Strip cookies in router's nginx config

samsimpson1:
This will replicate current production functionality without interfering with AB testing done at Fastly.

Trello: https://trello.com/c/HXtsiNCO/855-strip-cookies-at-fastly


nsabri1 (approved):
LGTM

PR #463, opened 2022-06-14
Using helm to define prometheus rules, a simple error ratio alert for Router.

desenhar:
https://trello.com/c/nrmpJKr4/934-essential-alerting

This is a PR for creating a new helm chart in the monitoring namespace, this helm chart solely concerns prometheus rules.
Any new alerts should be added to alerts.yml and its respective unit test added to test.yml

This alert has been tested in integration, this was done by scaling deployments of several frontend apps to zero, pausing the autosync in ArgoCD, and then finally using 'wrk' tool to simulate loads on smartanswers endpoint, this resulted in the alert firing in alertmanager.


charts/monitoring/rules/README.md
```diff
@@ -0,0 +1,5 @@
+## Run prometheus unit tests for alerts
```
nsabri1, line 1: small nit - Prometheus should be upper case

nsabri1 (approved):
OK

PR #466, opened 2022-06-20
Essential Alerting - refactored

desenhar:
https://trello.com/c/nrmpJKr4/934-essential-alerting
This is a simple error ratio alert for Router (job="apps/router"). The alert should fire if the overall ratio of errors to requests exceeds 0.1 (i.e. 10%).

Any new alerts should be added to the subdirectory named after the respective app within alerts.yaml.

This alert has been tested in integration, this was done by scaling deployments of several frontend apps to zero, pausing the autosync in ArgoCD, and then finally using 'wrk' tool to simulate loads on smartanswers endpoint, this resulted in the alert firing in alertmanager.


charts/govuk-apps-conf/templates/alert-rules/router.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: router-alert-rules
+  namespace: monitoring
+  labels:
+    release: prometheus
+    app: router
+spec:
+  {{- $.Files.Get "rules/router/alerts.yaml" | nindent 2}}
```
sengi, line 10: What's the reason for not putting these inline? (similarly for the alert-rules one)

desenhar: what do you mean?

desenhar: Also, I deleted the other alert-rules.yaml, that was brought in when I cherry picked a commit, shouldn't have been there.

kentsanggds: Reason for not bringing them inline is to allow us to make changes in only 1 place to run the `promtool` unit tests and to deploy it using a helm chart.

charts/govuk-apps-conf/rules/router/tests.yaml
```diff
@@ -0,0 +1,41 @@
+rule_files:
+    - router-alerts.yaml
```
kentsanggds, line 2: This should probably be alerts.yaml

charts/govuk-apps-conf/rules/router/README.md
```diff
@@ -0,0 +1,5 @@
+## Run Prometheus unit tests for alerts
```
fredericfran-gds, line 1: I would have preferred a separate helm chart to include all the Prometheus alerts.
Usually, this chart is for shared config between apps.

I would rename `rules` to `prometheus-rules` if kept in this chart and move the README.md file directly in that folder as I guess it applies to multiple apps that we will add here.

charts/govuk-apps-conf/rules/router/alerts.yaml
```diff
@@ -0,0 +1,11 @@
+groups:
+- name: RouterAlerts
+  rules:
+  - alert: RouterBackendHandlerResponseErrorRate
+    expr: sum(rate(router_backend_handler_response_duration_seconds_count{response_code=~"^5..", job="apps/router"}[5m])) / sum(rate(router_backend_handler_response_duration_seconds_count{job="apps/router"}[5m])) > 0.1
```
fredericfran-gds, line 5: I know it is a test but it seems it only considers 5xx HTTP codes as error, might need to expand that to include 4xx also?

charts/govuk-apps-conf/rules/router/README.md
fredericfran-gds: I think some automated test should be run in GitHub Actions for that, might be a separate story.

fredericfran-gds (changes requested):
great work. see inline comments.

charts/govuk-apps-conf/rules/router/README.md
desenhar: We were deciding on whether to relocate the prometheus rules to either rails-app or apps-conf, we ultimately chose the latter as the structure made more sense, as we could group alerts with their respective apps, this also provided an easier way to deploy using ArgoCD as the rules would deploy alongside its app.
As opposed to configuring automation for an entirely new helm chart.

charts/govuk-apps-conf/rules/router/alerts.yaml
fredericfran-gds: So in [ec2 production grafana](https://grafana.blue.staging.govuk.digital/dashboard/file/2ndline_health.json?refresh=1m&orgId=1), we care about 4xx (excluding 429-too many requests), 5xx and 429 as separate. We have a graph for them each.





[4:58](https://gds.slack.com/archives/C013F737737/p1655999882630559)
I was wondering about combining the 3 so that we have one alert. Obviously the alert will trigger more often as it is the sum…

fredericfran-gds (approved):
thanks for doing the modifications!

PR #470, opened 2022-06-28
Revert "Using helm to define prometheus rules, a simple error ratio alert for Router."

kentsanggds:
Reverts alphagov/govuk-helm-charts#463

No longer needed as we will be adding prometheus rules using another means.


nsabri1 (approved):
OK

nsabri1 (commented):
charts/monitoring/values.yaml you can remove this also

PR #471, opened 2022-06-29
Add PrometheusRule schema

kentsanggds:
This was removed in this revert PR - https://github.com/alphagov/govuk-helm-charts/pull/470, but is actually needed for PrometheusRules kind to be allowed. 

This schema was created by creating an output of the yaml, generating the json from the yaml and then inferring the schema from the yaml.

The descriptions and definitions were based on

https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/prometheus.rules.json


nsabri1 (approved):
LGTM

PR #474, opened 2022-07-08
Fix external-secrets interpretation of refreshInterval

fredericfran-gds:
Argo is trying to always rewrite the refresh interval to 1h from
1h0m0s.

See bug description:
https://github.com/external-secrets/external-secrets/pull/1028


marcpomfret (approved):
LGTM

PR #476, opened 2022-07-13
Convert image_tag files to yaml

fredericfran-gds:
See individual commits but the main aim of this PR is to convert the image_tag files to yaml and then modify all read/write functions to take into account this change.


nsabri1 (approved):
LGTM

charts/argo-workflows/scripts/update-image-tag.sh
```diff
@@ -20,7 +20,7 @@ if [[ $(<"${FILE}") = "${IMAGE_TAG}" ]]; then
 elif [[ "${LATEST_GIT_SHA}" = "${IMAGE_TAG}" ]] || [[ "${MANUAL_DEPLOY}" = true ]]; then
   git checkout -b "${BRANCH}"
 
-  echo "${IMAGE_TAG}" >"${FILE}"
+  tag="${IMAGE_TAG}" yq -i '.image_tag = env(tag)' "${FILE}"
```
sengi, line 23: Might be slightly more readable without the extra variable. (I'm assuming IMAGE_TAG is already in the environment; if not then just needs an `export IMAGE_TAG` before this line.)
```suggestion
  yq -i '.image_tag = env(IMAGE_TAG)' "${FILE}"
```

sengi (approved):

charts/argo-workflows/scripts/update-image-tag.sh
fredericfran-gds: thanks, it is an env already so I will not create a new one.

PR #477, opened 2022-07-14
Add new argo-workflow to set the automatic_deploys_enabled

fredericfran-gds:
This new argo-workflow will set the automatic_deploys_enabled and
optionally the image tag


nsabri1 (approved):
Logically LGTM

PR #481, opened 2022-07-18
fix #478 if_statement issue

fredericfran-gds:


charts/argo-workflows/scripts/update-image-tag.sh
```diff
@@ -62,7 +62,7 @@ else
   echo "Image tag not updated for ${ENVIRONMENT}: image tag not the latest commit on main."
 fi
 
-if [[ "${CHANGED}" ]]; then
+if "${CHANGED}"; then
```
sengi, line 65: Bash doesn't really have booleans in the way that you want. (This line has a security issue, for example: if I set $CHANGED to `rm -rf /` it'll run that in this statement.)

You're looking for `CHANGED=0` / `CHANGED=1` for setting it and `if [[ $CHANGED -ne 0 ]]` for checking it.

fredericfran-gds: cool, I will fix that.

sengi: Thanks!

sengi (approved):

PR #482, opened 2022-07-18
Updates to staging apps after loadlesting

marcpomfret:
Increased Pod CPU and WEB_CONCURRENCY app value to increase app request processing to the point where connections are not dropped


sengi (approved):
I'd recommend making the same changes in `values-production.yaml` in this PR just so there's less chance of them getting forgotten / out of sync.

PR #484, opened 2022-07-19
fix invalid branch due to double / when no image tag

fredericfran-gds:


charts/argo-workflows/scripts/set-automatic-deploys-enabled.sh
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-BRANCH="update-image-tag/${REPO_NAME}/${ENVIRONMENT}/${IMAGE_TAG}/${AUTOMATIC_DEPLOYS_ENABLED}"
+BRANCH="update-image-tag/${REPO_NAME}/${ENVIRONMENT}/${AUTOMATIC_DEPLOYS_ENABLED}"
+if [[ -z "${IMAGE_TAG}" ]]; then
```
sengi, line 5: Is the condition the right way around?

fredericfran-gds: nah, fixed

charts/argo-workflows/scripts/set-automatic-deploys-enabled.sh
```diff
@@ -21,7 +25,7 @@ commit() {
 }
 
 update_image_tag() {
-  if [[ -z "${IMAGE_TAG}" ]]; then
+  if [[ -n "${IMAGE_TAG}" ]]; then
```
sengi, line 28: Good catch!

sengi (approved):
Thanks!

PR #489, opened 2022-07-29
Remove base 64 decoding for repository SSH keys

theseanything:
The key can be stored without being base 64 encoded and no longer needs to be decoded by External Secrets. Originally it might have been encoded to preserve newline formatting when inputting into AWS Secret Manager. However, newlines are preserved if secret is created using JSON with '\n' characters. Ultimately this removes unnecessary complexity in the External Secret configuration.


sengi (approved):
Thanks for doing the hard work to simplify this! Helps keep the system robust and understandable ❤️

PR #491, opened 2022-08-02
Add initial workflow for updating image tag

theseanything:
This adds the workflow into the default branch so it can be tested. This workflow is used to update the image tag and can be used by our deployment process. We are migrating this logic from an existing Argo Workflow.

### DIFF
--- a/.github/workflows/update-image-tag.yml
+++ b/.github/workflows/update-image-tag.yml
@@ -1,0 +1,76 @@
+name: Update image tag
+
+on:
+  repository_dispatch:
+    types: [update-image-tag]
+  workflow_dispatch:
+    inputs:
+      image_tag:
+        description: 'Image tag to deploy'
+        required: true
+        type: string
+      repo:
+        description: 'The application repository'
+        required: true
+        type: string
+        default: 'collections'
+      environment:
+        description: 'Environment to deploy to'
+        required: true
+        type: choice
+        options:
+        - integration
+        - staging
+        - production
+        default: 'integration'
+      manual_deploy:
+        description: 'Manual deploy'
+        required: true
+        type: boolean
+
+jobs:
+  update-image-tag:
+
+    permissions: write-all
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-image-tag.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ubuntu-latest
+    env:
+      repo: ${{ github.event.client_payload.repository || inputs.repo }}
+      image_tag: ${{ github.event.client_payload.image_tag || inputs.image_tag }}
+      environment: ${{ github.event.client_payload.environment || inputs.environment }}
+      branch: "update-image-tag/${{ env.repo }}/${{ env.environment }}/${{ env.image_tag }}"
+      config_file: "charts/argocd-apps/image-tags/${{ env.environment }}/${{ env.repo }}"
+      manual_deploy: ${{ github.event.client_payload.manual_deploy || inputs.manual_deploy }}
+    steps:
+      - uses: chrisdickinson/setup-yq@latest
```
sengi, line 44: security: consider pinning this to a specific commit since it's third-party.

.github/workflows/update-image-tag.yml
```diff
@@ -0,0 +1,76 @@
...
+    inputs:
+      image_tag:
+        description: 'Image tag to deploy'
+        required: true
+        type: string
+      repo:
+        description: 'The application repository'
+        required: true
+        type: string
+        default: 'collections'
```
sengi, line 16: Do we want a default here? Or is this just for testing for now?

sengi (approved):

.github/workflows/update-image-tag.yml
theseanything: Yeah this is just a demo config for testing.

.github/workflows/update-image-tag.yml
theseanything: Yup - will probably look at a better way of using yq in GHA

PR #492, opened 2022-08-02
Update cpu for gov-frontend based on loadtesting

marcpomfret:


martynipratt (approved):
Working with Marc on this. Have seen the load test results and confirm they are passing the benchmarks we require.

fredericfran-gds (approved):
ok,

PR #495, opened 2022-08-05
Add Prometheus podmonitor to workers

fredericfran-gds:
Sidekiq-based workers will now exposed Sidekiq metrics in a
Prometheus format so adding a Prometheus pod monitor.

While all the workers of the same app will exposed "same" metrics
since they have the "same" view of the queue, this is better than
other discussed alternatives.

Ref:
1. [trello card](https://trello.com/c/K8aUtwz4/974-graphs-for-sidekiq-queues-in-grafana)

### DIFF
--- a/charts/govuk-rails-app/templates/worker-deployment.yaml
+++ b/charts/govuk-rails-app/templates/worker-deployment.yaml
@@ -26,6 +26,10 @@ spec:
       containers:
         - name: app
           image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          imagePullPolicy: {{ .Values.appImage.pullPolicy | default "Always" }}
+          ports:
+            - name: metrics
+              containerPort: {{ .Values.metricsPort }}
           command: ["bundle"]
           args: ["exec", "sidekiq", "-C", "config/sidekiq.yml"]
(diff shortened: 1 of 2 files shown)


charts/govuk-rails-app/templates/worker-deployment.yaml
```diff
@@ -26,6 +26,10 @@ spec:
       containers:
         - name: app
           image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          imagePullPolicy: {{ .Values.appImage.pullPolicy | default "Always" }}
```
sengi, line 29: I was dubious about specifying a default here, but then found it's consistent with what we already have in `deployment.yaml` so 🤷👍

sengi (approved):
LGTM, thanks!

FWIW I dislike the proliferation of these ugly PodMonitor objects, but I still haven't gotten around to eliminating the need for them so they're still needed for now unfortunately.

charts/govuk-rails-app/templates/worker-deployment.yaml
fredericfran-gds: yeah, I copied it across. Thanks!

PR #496, opened 2022-08-05
Aligned values files for int, staging and prod

marcpomfret:
Added appResources limits for Staging and Production
Removed left over securityContext values for Staging and Production
Re-arranged a few vars to make file comparisons clearer


fredericfran-gds (approved):
ok

PR #497, opened 2022-08-06
Fix whitehall-admin attachment/image uploads.

sengi:
`whitehall-admin` assumes that all of its processes share the same filesystem for uploading asset files.

- Rename nginxExtraVolumes to extraVolumes. There's never been anything nginx-specific about it; it's inherently a pod-level thing.
- Allow passing extra volumeMounts into the app pod spec in charts/govuk-rails-app.
- Use the same NFS (EFS) filer for whitehall-admin asset uploads as for asset-manager asset uploads.
- Use the same UID/GID for whitehall-admin as the old EC2 stack so that we can run in parallel without filesystem permissions issues on the NFS filer.

Tested: installed `whitehall-admin` and `frontend` via Helm and checked that they start up without errors. e.g. `cd charts/argocd-apps && helm install chris-test-wh-admin ../govuk-rails-app --values <(helm template . --values values-integration.yaml |yq e '.|select(.metadata.name=="whitehall-admin").spec.source.helm.values')`

### DIFF
--- a/charts/govuk-rails-app/templates/deployment.yaml
+++ b/charts/govuk-rails-app/templates/deployment.yaml
@@ -71,6 +71,10 @@ spec:
             runAsUser: 1001
             runAsGroup: 1001
           {{- end }}
+          {{- with .Values.appExtraVolumeMounts }}
+          volumeMounts:
+            {{ . | toYaml | trim | nindent 12 }}
+          {{- end }}
         - name: nginx
           image: "{{ .Values.nginxImage.repository }}:{{ .Values.nginxImage.tag }}"
           imagePullPolicy: {{ .Values.nginxImage.pullPolicy | default "Always" }}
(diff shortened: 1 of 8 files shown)


charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -65,11 +65,13 @@ spec:
...
             periodSeconds: 5
           {{- end }}
-          {{- if .Values.securityContext.appContainer }}
           securityContext:
-            allowPrivilegeEscalation: false
-            runAsUser: 1001
-            runAsGroup: 1001
+            allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
+            runAsUser: {{ .Values.securityContext.runAsUser }}
+            runAsGroup: {{ .Values.securityContext.runAsGroup }}
```
nsabri1, line 71: You probably want to apply same to Worker-deployment to keep both containers consistent?

nsabri1 (approved):
LGTM

charts/govuk-rails-app/templates/deployment.yaml
sengi: Whoops, good catch, thanks!

nsabri1 (approved):
LGTM

PR #498, opened 2022-08-08
Add Prometheus podmonitor to Asset-Manager workers

fredericfran-gds:
Similar to #495 but added here because Asset-Manager has its own Helm
chart


nsabri1 (approved):
LGTM

PR #501, opened 2022-08-10
Fix ownership of mounted volumes for Asset-manager

fredericfran-gds:
set fsGroup in securityContext:
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/


nsabri1 (approved):
OK

PR #503, opened 2022-08-10
Fix Clamav of Asset-manager

fredericfran-gds:
It is currently complaining about missing the clamd.conf so added it
from EC2 integration

### DIFF
--- a/charts/asset-manager/templates/worker-deployment.yaml
+++ b/charts/asset-manager/templates/worker-deployment.yaml
@@ -23,7 +23,7 @@ spec:
       securityContext:
         seccompProfile:
           type: RuntimeDefault
-        # Asset Manager needs to run as 2899 since it shares the same NFS volume with 
+        # Asset Manager needs to run as 2899 since it shares the same NFS volume with
         # EC2 counterpart
         runAsUser: 2899
         runAsGroup: 2899
@@ -69,6 +69,9 @@ spec:
           volumeMounts:
             - name: clam-virus-db
               mountPath: /var/lib/clamav
+            - name: clamd-conf
+              mountPath: /etc/clamav/clamd.conf
+              subPath: clamd.conf
           securityContext:
             allowPrivilegeEscalation: false
       {{- with .Values.dnsConfig }}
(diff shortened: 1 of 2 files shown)


charts/asset-manager/templates/worker-deployment.yaml
```diff
@@ -69,6 +69,9 @@ spec:
           volumeMounts:
             - name: clam-virus-db
               mountPath: /var/lib/clamav
+            - name: {{ .Release.Name }}-clamd-conf
+              mountPath: /etc/clamav/clamd.conf
+              subPath: clamd.conf
```
sengi, line 74: It's best not to use subPath because this prevents updating the configmap.

sengi (approved):

charts/asset-manager/templates/worker-deployment.yaml
sengi: I dunno to what extent we care in this case though. Not saying you need to change it, just something to be aware of when using it. If the dir is empty anyway apart from this file then I would get rid of the subPath and just mount as `/etc/clamav`, but if not then maybe it's not worth the faff.

charts/asset-manager/templates/worker-deployment.yaml
```diff
@@ -69,6 +69,9 @@ spec:
           volumeMounts:
             - name: clam-virus-db
               mountPath: /var/lib/clamav
+            - name: {{ .Release.Name }}-clamd-conf
```
sengi, line 72: No need for `Release.Name` here. The name is scoped to within the pod.

sengi, line 72: Looks like you forgot to define this `Volume`?

sengi (approved):
Noticed a couple of other things

charts/asset-manager/templates/worker-deployment.yaml
fredericfran-gds: ok, let me see what's in the volume.

fredericfran-gds: other stuff inside unfortunately.

charts/asset-manager/templates/worker-deployment.yaml
fredericfran-gds: I can fix that, thanks

PR #505, opened 2022-08-11
Add Clamav config

fredericfran-gds:
Add missing freshclam [config](https://github.com/alphagov/govuk-puppet/blob/32c1bbbb10067078c1406170666a135b4a10aaea/modules/clamav/manifests/config.pp#L14).

Add clamd config to worker containers as well, before it was only on the
freshclam containers but this is a shared config.


nsabri1 (approved):
LGTM

PR #508, opened 2022-08-13
Fix Clamav for Asset-manager

fredericfran-gds:
Fixes:

1. In the worker pod, we run clamd (daemon verion of clamav).
   This seems to lead to no killed clamav scans and also this is
   how it is currently run in EC2 rather than being run standalone.

2. Remove the freshclam (updates virus database of clamav)
   container from continuously running in the worker pod and set
   it to run as presync and then run a cronjob every hour for it.

3. Use clamav EFS to share virus databases across all clamav containers.
   (https://github.com/alphagov/govuk-infrastructure/pull/725)

Related PRs:
1. [asset-manager dockerfile](https://github.com/alphagov/asset-manager/pull/938)

### DIFF
--- a/charts/asset-manager/templates/clamd-configmap.yaml
+++ b/charts/asset-manager/templates/clamd-configmap.yaml
@@ -1,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-clamd-conf
+  labels:
+    {{- include "asset-manager.labels" . | nindent 4 }}
+    app.kubernetes.io/component: clamd
+data:
+  clamd.conf: |-
+
+    LogTime yes
+    LogSyslog yes
+    LogVerbose yes
+    ExtendedDetectionInfo yes
+    TCPSocket 3310
+    TCPAddr 127.0.0.1
+    MaxConnectionQueueLength 50
+    StreamMaxLength 50M
+    MaxThreads 20
+    ReadTimeout 120
+    DetectPUA no
+    AlgorithmicDetection yes
+    ScanPE yes
+    ScanELF yes
+    ScanOLE2 yes
+    OLE2BlockMacros no
+    ScanPDF yes
+    ScanMail yes
+    MaxScanSize 250M
+    MaxFileSize 105M
+    TemporaryDirectory /tmp
--- a/charts/asset-manager/templates/freshclam-cronjob.yaml
+++ b/charts/asset-manager/templates/freshclam-cronjob.yaml
@@ -1,0 +1,12 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ .Release.Name }}-freshclam
+spec:
+  schedule: "19 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          {{- include "asset-manager.freshclam.podspec" . | indent 8 }}
+          restartPolicy: Never
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -22,13 +22,14 @@ govukApplications:
       hosts:
       - name: asset-manager.eks.integration.govuk.digital
     workerReplicaCount: 1
-    nfsServer: &assets-nfs assets.blue.integration.govuk-internal.digital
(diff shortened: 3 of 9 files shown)


charts/asset-manager/templates/clamd-configmap.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-clamd-conf
+  labels:
+    {{- include "asset-manager.labels" . | nindent 4 }}
+    app.kubernetes.io/component: web
```
sengi, line 7: Probably want this to be something other than web. `clam`?

charts/asset-manager/templates/freshclam-cronjob.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ .Release.Name }}-freshclam
+spec:
+  schedule: "0 * * * *"
```
sengi, line 6: Just to be nice neighbours, I'd probably set the minute to something randomly-chosen that isn't 0. Unfortunately it looks like k8s CronJob still doesn't do fuzzing of start times.

charts/asset-manager/templates/freshclam-cronjob.yaml
```diff
@@ -0,0 +1,49 @@
...
+            # Asset Manager needs to run as 2899 since it shares the same NFS volume with
+            # EC2 counterpart
+            runAsUser: 2899
+            runAsGroup: 2899
+            fsGroup: 2899
+          containers:
+            - name: freshclam
+              image: "{{ required "A valid .Values.appImage.repository entry required!" .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+              imagePullPolicy: {{ .Values.appImage.pullPolicy | default "Always" }}
+              command: ["/bin/sh"]
```
sengi, line 23: Does it really need a shell? Ditto below in worker-deployment.yaml.

sengi (approved):

charts/asset-manager/templates/freshclam-cronjob.yaml
sengi: (I'm assuming we're not planning on running more than like 3 of these, right? If not then we'll need to go a bit further, say by adding a random starting delay inside the job.)

nsabri1: I also believe there is a rate limit on how often you can poll also

charts/asset-manager/templates/clamd-configmap.yaml
fredericfran-gds: thanks, I will fix

charts/asset-manager/templates/freshclam-cronjob.yaml
fredericfran-gds: copy and paste from previous art but I can test that. I believe not.

charts/asset-manager/templates/freshclam-cronjob.yaml
fredericfran-gds: ok, maybe I will change the architecture. I will have only one freshdaemon per cluster and the volume will be mounted in the needed containers.

I need to do some more checks on PVs for our cluster thanks.

charts/asset-manager/templates/freshclam-cronjob.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: {{ .Release.Name }}-freshclam
+spec:
+  schedule: "19 * * * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          {{- include "asset-manager.freshclam.podspec" . | indent 8 }}
+          restartPolicy: OnFailure
```
sengi, line 12: optional: `Never` might be preferable here since it'll make it easier to debug failures and we'll try again automatically the next hour anyway.
```suggestion
          restartPolicy: Never
```

sengi (approved):
Cool 👍

charts/asset-manager/templates/freshclam-cronjob.yaml
fredericfran-gds: Ok, I will change that.

PR #509, opened 2022-08-15
Remove stray occurrences of securityContext.appContainer.

sengi:
Missed these from #497. Should fix `Error: container has runAsNonRoot and image will run as root (pod: "...", container: nginx)`.


fredericfran-gds (approved):
ok

PR #514, opened 2022-08-16
Asset-Manager: switch to EFS CSI driver and PersistentVolumeClaim

fredericfran-gds:
Due to the issue with non-root asset-manager trying to use a root
NFS volume created in AWS/terraform directly. We move to using the
EFS CSI driver and PersistentVolumeClaim.

See related PR: https://github.com/alphagov/govuk-infrastructure/pull/725

### DIFF
--- a/charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
+++ b/charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: clamav-db
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: clamav-db-efs-sc
+  resources:
+    requests:
+      storage: 1Gi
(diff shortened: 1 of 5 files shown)


charts/asset-manager/templates/worker-deployment.yaml
```diff
@@ -90,9 +90,8 @@ spec:
           server: "{{ .Values.assetManagerNFS }}"
           path: /
       - name: clam-virus-db
-        nfs:
-          server: "{{ .Values.clamavNFS }}"
-          path: /
+        persistentVolumeClaim:
+            claimName: clamav-db
```
sengi, line 94: nit: double indent?

```suggestion
          claimName: clamav-db
```

charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
```
sengi, line 2: I'm guessing the EFS CSI driver takes care of the PersistentVolume side of things and that's why we don't define a PersistentVolume ourselves? Is that right?

nsabri1 (approved):
LGTM

sengi (approved):

charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
fredericfran-gds: yep

PR #518, opened 2022-08-17
Fix Grafana Dashboards

fredericfran-gds:
1. some config maps shared the same key: dasboards.json which
   can lead to conflicts because they are projected on the same files.
2. remove some notes copied by error


sengi (approved):
Aha! Thanks for finding + fixing! 🙌

PR #520, opened 2022-08-18
Fix grafana router dashboard

fredericfran-gds:
The request latency graph is not working, this commit fixes that.

grafana router request latency


sengi (approved):
Nice — thanks for fixing this!

PR #521, opened 2022-08-18
Govuk-rails-app: fix security context

fredericfran-gds:
Set runAsUser and runAsGroup at the pod level throughout.
Keeps things consistent for an app.

Solve the issue with upload-assets job with error:
"Init:container has runAsNonRoot and image has non-numeric user (app), cannot verify user is non-root"


sengi (approved):
Cool, nice bit of cleanup there! Thanks!

PR #523, opened 2022-08-22
Add Sentry DSN configuration for govuk-apps

theseanything:
This updates the configuration for govuk-rails-apps to support Sentry DSN configuration. An External Secret is now created for every application which will synchronise a secret from AWS Secrets Manager with a K8s secret in the cluster that contains the DSN. The SENTRY_DSN environment variable is then set to the value of that secret. This allows application to use Sentry for error reporting. Each application requires its own unique DSN as they are project specific.


fredericfran-gds (commented):
I don't think the secret should be under govuk/common, it should have been govuk/<app_name>/sentry-dsn to follow the conventions.

theseanything:
> I don't think the secret should be under govuk/common, it should have been govuk/<app_name>/sentry-dsn to follow the conventions.

Changed the implementation to use a shared AWS Secret Manager secret - so will remain as a common secret

fredericfran-gds (approved):
It's good, thanks. As discussed, we will clean up later. thanks

PR #524, opened 2022-08-22
Add option to disable Sentry for govuk app

theseanything:
The test govuk app does not need Sentry configured. This adds configuration to disable creating Sentry secrets and setting env vars.

https://trello.com/c/kbrv62b4


sengi (approved):
Cool 👍

PR #525, opened 2022-08-22
Add SENTRY_DSN to workers containers

theseanything:
This adds missing env var needed to configure sentry on govuk-app worker containers.

https://trello.com/c/kbrv62b4


sengi (approved):
Good catch!

PR #527, opened 2022-08-23
Remove the duplicate Nginx logging

theseanything:
Nginx was configured to log access logs twice (once to /dev/stderr and /dev/stdout) - this cause logs to show up twice in logit, where one set is not parse as is plaintext formatted. This removes the duplicate logs logged to /dev/stderr using the timed_combined format (which is a plaintext format).


nsabri1 (approved):
Looks good

PR #528, opened 2022-08-23
Enable JSON formatted Ngnix logs for apps

theseanything:
This changes ngnix logs to be json formatted so they can be parsed by logit.


nsabri1 (approved):
LGTM

PR #531, opened 2022-08-23
Fix content store BACKNED_URL for router

theseanything:
Content store is currently not running in the EKS infra, hence requests should be forwarded to content store in the old infra.


sengi (approved):
Good catch!

PR #533, opened 2022-08-23
Update Nginx logs to JSON output for router

theseanything:
This updates the log format to use json format so it can be parsed by Logit.


samsimpson1 (approved):
LGTM

PR #535, opened 2022-08-23
Update all 3 whitehall probes

marcpomfret:
Have to update all probe values, can't just sub in certain values with current setup


sengi:
Ah sorry, can blame me for that, think I wrote that sketchy bit of Helm template 😅

sengi (approved):

PR #536, opened 2022-08-23
Improve dashboards to make it easier to detect deviations

fredericfran-gds:
Dashboards improved:
- apps
- router

Still to do:
Request duration for router.

new apps dashboard

new router dashboard

### DIFF
--- a/charts/grafana-dashboards/apps-dashboard.json
+++ b/charts/grafana-dashboards/apps-dashboard.json
@@ -28,7 +28,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "iteration": 1653573951592,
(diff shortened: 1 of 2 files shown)


sengi:
Graphs look fine but could you elaborate a bit on what's changing here? It's super hard to tell from the generated JSON 😅

sengi (approved):

fredericfran-gds:
Yeah, I can. So before, there was no overall `all` plot in the app's graphs to see the general trend, we had to switch between each app so likely to miss something.

The first 2 routers graphs were challenging to use to see the errors, here we could see the errors coming clearly from Signon and Email-alert-frontend.

sengi:
Ah I see — thanks that's a nice improvement!

PR #538, opened 2022-08-24
Don't apply article slug rewrite rules to Rails assets URLs.

sengi:
The rewrites for article slugs (e.g. rejecting mixed case paths, downcasing all-caps paths, etc.) were being applied to Rails assets as well as other content.

This was causing 404s on resources such as the picture of Churchill on https://www.eks.production.govuk.digital/government/history because the asset filename contained both uppercase and lowercase letters (`Churchill_01.jpg`).

Nest the slug rewrite rules under the default `location /` block so that they only apply to paths which don't match a more specific top-level `location` block.


samsimpson1 (approved):
LGTM

theseanything (approved):

PR #540, opened 2022-08-24
Fix ASSET_HOST domain

fredericfran-gds:
Quickfix for ASSET_HOST which must have value: https://www.gov.uk
in production.


nsabri1 (approved):
OK

sengi:
Nice!

PR #542, opened 2022-08-24
Fix setting proxy headers in Ngnix for Router

theseanything:
The proxy headers were not being set by Nginx for request handled by
location /. This is because the proxy_set_header directive is defined in
the location / block, which leads Ngnix to ignore any proxy_set_headers
in previous configuration blocks. From Nginx documentation:

> These directives are inherited from the previous configuration level
if and only if there are no proxy_set_header directives defined on the
  current level

This meant headers such as host, x-forwarded etc were not being set and
cause user facing issues as downstream application rely on them being
set correctly. For example, Whitehall requires the x-forwarded-host
header to correctly redirect users.

### DIFF
--- a/charts/govuk-apps-conf/templates/router-nginx-config.tpl
+++ b/charts/govuk-apps-conf/templates/router-nginx-config.tpl
@@ -95,11 +95,6 @@ http {
   server {
     listen 8080;
 
-    proxy_set_header   Host $http_host;
-    proxy_set_header   X-Real-IP $remote_addr;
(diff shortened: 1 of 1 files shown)


sengi (approved):
Should we nuke lines 98–102? Am I right in thinking those don't actually do anything any more?

theseanything:
> Should we nuke lines 98–102? Am I right in thinking those don't actually do anything any more?

Ah good catch - forgot to remove.

PR #543, opened 2022-08-24
Fix assets redirect for government-frontend

fredericfran-gds:
Needs a plek entry to redirect assets to the ec2 assets endpoint.

See government-frontend code snippet:
https://github.com/alphagov/government-frontend/blob/main/app/controllers/asset_manager_redirect_controller.rb#L3


nsabri1 (approved):
LGTM

PR #546, opened 2022-08-24
Remove promotion step when nextEnvironment is not set

theseanything:
This removes the deployment promotion step from the postsync workflow when the nextEnvironment variable is not set. This fixes an issue in production where no promotion is required - however the postsync job fails as workflow tries to promote the deploy without a nextEnvironment.


fredericfran-gds (approved):
ok

PR #550, opened 2022-08-24
Allow cross communication between EKS and EC2 www-origins

fredericfran-gds:
See [PR](https://github.com/alphagov/govuk-infrastructure/pull/730)
for security group created and consequences.


sengi (approved):
This is awesome — love that it deletes two of these duplicated ACLs 🎉

theseanything:
@fredericfran-gds is this still something we want?

PR #553, opened 2022-08-25
Sync k8s auto-deploys-enabled flags with EC2.

sengi:
Router, licence-finder, static and signon are still manually deployed; everything else should be auto. Everything in integration should be auto.


nsabri1 (commented):
LGTM

nsabri1 (approved):
LGTM

PR #556, opened 2022-08-30
Currently Origin LB is stating unhealthy Router pods

nsabri1:
This is due to using incorrect HTTP path on port 8080.

Updating health check to` /readyz `- this is the Nginx container health check on port 8080.


samsimpson1 (approved):
LGTM

PR #558, opened 2022-08-30
Set ASSET_HOST for only whitehall

theseanything:
ASSET_HOST doesn't need to be set for all GOV.UK Rails applications. It's used by Rails to specify a alternative domain for links to assets, however we now serve all assets from the same origin and can use relative paths. The only exception is Whitehall which renders CSV preview pages from the assets.publishing.services.gov.uk domain and needs to know to retireve assets from the original origin.

Unclear for previous history, why we set this in the first place.


sengi (approved):
Awesome, thanks!

PR #559, opened 2022-08-31
default health check path update

nsabri1:
As the Ingress target port is set to the Nginx container, therefore updating the health check path to match what is set in Nginx deployment.
https://github.com/alphagov/govuk-helm-charts/blob/b36296164fa7e0fbd469bb593ae7699fa95c76b1/charts/govuk-rails-app/templates/deployment.yaml#L91

Annotations are used to customise kubernetes Ingress and Service
objects:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/guide/ingress/annotations.md

### DIFF
--- a/charts/govuk-rails-app/values.yaml
+++ b/charts/govuk-rails-app/values.yaml
@@ -12,7 +12,7 @@ ingress:
     alb.ingress.kubernetes.io/target-type: ip
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: "443"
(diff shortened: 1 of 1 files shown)


sengi (commented):
What are the practical consequences of this change and how have you tested it? (for example do rolling updates still work without serving errors, what happens when an app container starts failing its readiness check, etc.)

sengi (approved):
I'm still not sure I 100% understand the purpose of the ALB healthcheck as it relates to **our** overall system, but we determined empirically the other day that the (pod/container) readiness probes failing/succeeding results in endpoints being removed/added from/to the LB target group — so I can't think of a scenario where this breaks anything.

It should reduce the rate of (unnecessary?) probes hitting the Ruby containers as well, which is a bonus.

LGTM

PR #560, opened 2022-09-05
Update govuk-replatform-test-app in integration.

kentsanggds:
## What

Move govuk-replatform-test-app section to keep list of apps alphabetical and support multiple devs on codelabs.

## Reference

https://trello.com/c/X357aij8/988-write-up-codelab-sessions


charts/argocd-apps/values-integration.yaml
```diff
@@ -497,6 +497,55 @@ govukApplications:
...
+        failureThreshold: 2
+        periodSeconds: 5
+    ingress:
+      enabled: true
+      annotations:
+        alb.ingress.kubernetes.io/load-balancer-name: govuk-replatform-test-app
+      hosts:
+      - govuk-replatform-test-app.eks.integration.govuk.digital
+    extraEnv:
+      - name: HELM_MESSAGE_1
```
theseanything, line 531: Could the app just pick up any Env Var with a prefix? Rather than having a pre-defined number of slots?

theseanything: i.e. prefix like `MESSAGE_`

theseanything: Nit: Not sure about using the name HELM_MESSAGE as the environment variable doesn't actually have anything to do with Helm, which might be confusing for people learning. (Yes, Helm does template the env vars for the app config, but isn't actually effected by their value.)

theseanything (approved):

charts/argocd-apps/values-integration.yaml
kentsanggds: Not sure about picking up any Env Var, I can simplify the prefix to just MESSAGE, reason I put HELM_MESSAGE is that its delivered from the helm charts and not from the test app

kentsanggds:
Sure we can put the app in a different namespace, not sure if other changes need to happen for that to work

nsabri1 (commented):
Have you considered using its on chart named replatform-test-app? rather then using the Default chart - i just feel its wrong that we making changes to the Default app chart just to accommodate the test app, not to say it is wrong.

kentsanggds:
I had considered that but I thought that adding it as a different chart will probably make some things more complicated and I wanted to align the test app as closely as possible to existing apps in terms of configuration in this repo.

nsabri1:
> I had considered that but I thought that adding it as a different chart will probably make some things more complicated and I wanted to align the test app as closely as possible to existing apps in terms of configuration in this repo.

Curious to know what things do you think will become more complicated - i feel it could potentially make things less complicated and keep those ENV values which you ave added to in integration.yaml, could just sit in the new test charts values.yaml

kentsanggds:
I think the pipeline will have to change to deploy things, also remember that the point of the test app is to guide a developer how making changes to a standard govuk app config will look. If the test app is set up completely differently it will be a less true reflection of a standard govuk app.

charts/argocd-apps/values-integration.yaml
```diff
@@ -497,6 +497,44 @@ govukApplications:
             key: bearer_token
       - name: PUBLISH_TOPICS
         value: "1"
+- name: govuk-replatform-test-app
+  repoName: govuk-replatform-test-app
```
sengi, line 501: nit: `repoName` is redundant here since it defaults to the value of `name`, so for clarity we would omit it.

sengi (approved):

charts/govuk-rails-app/templates/deployment.yaml
```diff
@@ -31,8 +31,10 @@ spec:
             - name: metrics
               containerPort: {{ .Values.metricsPort }}
           envFrom:
+          {{- if .Values.govukAppsEnv.enabled }}
```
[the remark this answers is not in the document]
kenlt-uk: It's been a while since I did this, but I think it was complaining about something not set which was why I have it as being optional. Could revisit it to see if there's a way to leave it in without causing the app to break

charts/argocd-apps/values-integration.yaml
```diff
@@ -497,6 +497,44 @@ govukApplications:
...
+  helmValues:
+    sentry:
+      enabled: false
+    uploadAssets:
+      enabled: false
+    appPort: 9292
```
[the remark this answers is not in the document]
kenlt-uk: yes, thats possible, the PR on the test app hasn't been merged in yet so could update it

charts/argocd-apps/values-integration.yaml
kenlt-uk: ok, will remove it

charts/govuk-rails-app/templates/deployment.yaml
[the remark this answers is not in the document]
sengi: Oh sorry, I see it - it's because we're installing this app in a different namespace, where that configmap won't exist. I don't mind which way we solve it for now - we can always iterate :)

PR #561, opened 2022-09-06
Remove ASSET_HOST for Whitehall

theseanything:
The ASSET_HOST doesn't need to be set as we serve assets on the same domain. This was originally needed for CSV preview pages which are rendered by Whitehall, however hosted on the asset.publishing.service.gov.uk domain. This currently only applies to our old infra (we don't serve asset traffic yet). This is being fixed in the old infra by adding a proxy to deal with assets request, rather than setting ASSET_HOST.


sengi (approved):
Cool

PR #562, opened 2022-09-06
Tag images that are in use 

desenhar:
https://trello.com/c/MK34KlBn/980-lifecycle-rules-for-ecr-to-clean-up-images

This is a PR to introduce an additional function in the image tagging process, the workflow will trigger the function, the image will then be tagged with a 'deployed-to-{{env}}' tag, this ecr cli commands have been tested in isolation and do not override existing tags.

### DIFF
--- a/charts/argo-workflows/scripts/update-image-tag.sh
+++ b/charts/argo-workflows/scripts/update-image-tag.sh
@@ -17,6 +17,13 @@ change_image_tag() {
   CHANGED=true
 }
 
+add_image_deployment_tag() {
+  aws ecr get-login-password --region eu-west-1 | docker login --username AWS --password-stdin "$ECR_REGISTRY"
+  MANIFEST=$(aws ecr batch-get-image --repository-name "${REPO_NAME}" --image-ids imageTag="${IMAGE_TAG}" --region eu-west-1 --query 'images[].imageManifest' --output json)
+  # Every image that has a tag will then have 'deployed-to-${ENVIRONMENT}' tag added as well.
+  aws ecr put-image --repository-name "${REPO_NAME}" --image-tag "deployed-to-${ENVIRONMENT}" --image-manifest "$MANIFEST"
+}
+
 git config --global user.email "${GIT_NAME}@digital.cabinet-office.gov.uk"
 git config --global user.name "${GIT_NAME}"
 
@@ -40,6 +47,7 @@ elif [[ "${LATEST_GIT_SHA}" = "${IMAGE_TAG}" ]]; then
   # Auto deploys are enabled unless explicitly set to "false" (case insensitive).
   if [[ "${auto_deploys,,}" != "false" ]]; then
     change_image_tag
+    add_image_deployment_tag
   else
     echo "Did not update image tag because automatic_deploys_enabled is set to false for app"
   fi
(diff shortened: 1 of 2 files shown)


charts/argo-workflows/scripts/update-image-tag.sh
```diff
@@ -37,6 +45,7 @@ elif [[ "${MANUAL_DEPLOY}" = true ]]; then
 # Relies on the assumption the IMAGE_TAG is a commit SHA
 elif [[ "${LATEST_GIT_SHA}" = "${IMAGE_TAG}" ]]; then
   auto_deploys="$(yq '.automatic_deploys_enabled' "${FILE}")"
+  add_tag
```
theseanything, line 48: We should only tag the image after we've successfully update the helm chart image tag value (on line 51).

theseanything (commented):
Also just a reminder to tidy up your commits before putting this review: https://gds-way.cloudapps.digital/standards/source-code/working-with-git.html#commit-messages

.github/workflows/update-image-tag.yml
```diff
@@ -27,6 +27,16 @@ on:
         description: 'Manual deploy'
         required: true
         type: boolean
+      ecr_registry:
```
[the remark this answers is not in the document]
theseanything: If we aren't changing this value, as it's always production ECR - this doesn't need to be a workflow input from the GitHub. I.e. just put the value in the Argo Workflow - charts/argo-workflows/templates/update-image-tag.yaml:47

charts/argo-workflows/scripts/update-image-tag.sh
```diff
@@ -17,6 +17,14 @@ change_image_tag() {
   CHANGED=true
 }
 
+add_image_deployment_tag() {
+  current_image_tag="$IMAGE_TAG"
```
theseanything, line 21: Why not just use IMAGE_TAG?

charts/argo-workflows/scripts/update-image-tag.sh
```diff
@@ -17,6 +17,14 @@ change_image_tag() {
...
+  aws ecr get-login-password --region eu-west-1 | docker login --username AWS --password-stdin "$ECR_REGISTRY"
+  MANIFEST=$(aws ecr batch-get-image --repository-name "${REPO_NAME}" --image-ids imageTag="${current_image_tag}" --region eu-west-1 --query 'images[].imageManifest' --output json)
+  # Every image that has a tag will then have 'deployed-to-production' tag added as well.
+  aws ecr put-image --repository-name "${REPO_NAME}" --image-tag "deployed-to-production" --image-manifest "$MANIFEST"
```
theseanything, line 25: We have to tag images being used in staging and integration also

```suggestion
  aws ecr put-image --repository-name "${REPO_NAME}" --image-tag "deployed-to-${ENVIRONMENT}" --image-manifest "$MANIFEST"
```

charts/argo-workflows/templates/update-image-tag.yaml
```diff
@@ -19,6 +20,7 @@ spec:
           - name: imageTag
           - name: manualDeploy
             default: "false"
+          - name: ecrRegistry
```
theseanything, line 23: Doesn't need to be an input as we're not changing it per workflow run

theseanything (changes requested):

theseanything:
> Also just a reminder to tidy up your commits before putting this review: https://gds-way.cloudapps.digital/standards/source-code/working-with-git.html#commit-messages

Can you also format the commit message as per the style guide and add in detail about why these changes?

desenhar:
The commit's have been squashed and one has been left with a description below justifying why this change was needed.

theseanything (approved):

PR #568, opened 2022-09-09
updated production external domain suffix to gov.uk in line with pupp…

martynipratt:
…et environment

Checking the production config in govuk puppet it looks like we pass the gov.uk as to root domain


marcpomfret (approved):
ok

PR #569, opened 2022-09-09
Updated PVC to point to new EFS static storage class.

nsabri1:
Also replied PreSync and removed SyncWave

Relates to:
https://github.com/alphagov/govuk-infrastructure/pull/740

### DIFF
--- a/charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
+++ b/charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
@@ -6,11 +6,11 @@ metadata:
     {{- include "asset-manager.labels" . | nindent 4 }}
     app.kubernetes.io/component: clamav-pvc
   annotations:
-    argocd.argoproj.io/sync-wave: "-1"
+    argocd.argoproj.io/hook: PreSync
 spec:
   accessModes:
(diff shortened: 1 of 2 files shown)


charts/asset-manager/templates/clamav-db-persistentvolumeclaim.yaml
```diff
@@ -6,11 +6,11 @@ metadata:
     {{- include "asset-manager.labels" . | nindent 4 }}
     app.kubernetes.io/component: clamav-pvc
   annotations:
-    argocd.argoproj.io/sync-wave: "-1"
+    argocd.argoproj.io/hook: PreSync
 spec:
   accessModes:
     - ReadWriteMany
-  storageClassName: clamav-db-efs-sc
+  storageClassName: clamav-db-efs
```
sengi, line 13: Since we're no longer using dynamic provisioning, this needs to be empty so that the PVC will match the PV. (It might also match if you omitted storageClassName entirely, but that's less robust because it depends whether there's a default storage class configured - see https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1.)
```suggestion
  storageClassName: ""
```

sengi (approved):

PR #570, opened 2022-09-09
Rename clamav-db EFS PVC to match static binding in TF.

sengi:
See https://github.com/alphagov/govuk-infrastructure/pull/740.

Also making the PVC name match the PV name, for clarity.


nsabri1 (approved):
LGTM

PR #571, opened 2022-09-09
Move clamav-db-efs PVC to Terraform.

sengi:
Its object lifecycle needs to match that of the PVC, not the app deployment etc.


nsabri1 (approved):
OK

PR #572, opened 2022-09-12
Updating ClamAV-DB path to NFS share which ClamAV daemon already uses…

nsabri1:
Updating ClamAV-DB path to AWS EFS/NFS share which ClamAV daemon already uses for storing scanned assets and infected assets.

Currently testing this change.


fredericfran-gds (approved):
looks good, thanks

PR #573, opened 2022-09-12
Notify slack only on failure for post sync workflows

theseanything:
This changes the notifications to only be sent when the workflow fails or errors, instead of also when it succeeds. Send notification aren't actionable and create noise in Slack. We really only care when it fails.


fredericfran-gds (approved):
ok, hope the double templating "fix" is working. Thanks.

PR #574, opened 2022-09-12
Remove Signon Resources

theseanything:
This was used to manage signon resources declaritively and synchornise secrets with Kubernetes. However the implementation is incomplete and cannot handle Staging being sychronised with Prod everyday. We currently only need to sync secrets with K8s, which we can do with the External Secrets and the webhook provider, as we're not concerned with spining entirely new cluster yet. Therefore we don't want to manage or maintain the complexity of moving Signon to a declaritive state as of yet. The API for signon will also be modified, so switching this off to prevent any issue with breaking changes to the API.


fredericfran-gds (approved):
looks good. Thanks for simplifying stuff...

I guess the reference to existing secrets created here are removed?

theseanything:
> I guess the reference to existing secrets created here are removed?

Sadly not, the secrets created by Signon resources will have to manually removed. But will do after it's removed - otherwise it'll recreate them ahah.

PR #576, opened 2022-09-14
Add the draft flavour of content-store in integration.

sengi:
The differences in env vars between regular and draft content-store come from
https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/class/draft_content_store.yaml

Recall that YAML alias overrides (`<<: *foo`) are shallow, not recursive/deep.

I don't believe the k8s secrets for the Signon tokens for this exist yet, but that's currently in a state of flux and I don't think there's anywhere for us to configure those right now. So for now, I expect the pods to fail to start because one or more non-optional Secrets won't exist. The idea is that those k8s Secrets be automatically populated one way or another.

Tested by inspecting `helm template` output, for example:

    helm template ../govuk-rails-app --name-template=draft-content-store --values <(
      helm template . --values values-integration.yaml | \
        yq e '.|select(.metadata.name == "draft-content-store").spec.source.helm.values'

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -148,7 +148,7 @@ govukApplications:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
 - name: content-store
-  helmValues:
+  helmValues: &content-store
     uploadAssets:
       enabled: false
     extraEnv:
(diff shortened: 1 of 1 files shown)


theseanything (approved):
Yup - makes sense, we can update the secrets once being synced correctly from Signon.

PR #578, opened 2022-09-15
Set Sentry environment for Router

theseanything:
Router app uses the Go Sentry SDK which looks for the SENTRY_ENVIRONMENT env var instead of SENTRY_CURRENT_ENV (set in common config). This fixes missing environment tags on issue reported from EKS.


sengi (approved):
Aha! Thanks!

PR #580, opened 2022-09-16
Added three new publishing apps with barebone values.

nsabri1:
This is so we can have them run in integration EKS cluster.

More values will be added as we look into the apps further.

https://trello.com/c/WwdA4FxT/721-package-travel-advice-publisher-for-migration 
https://trello.com/c/V3qwIgq3/720-package-manuals-publisher-for-migration 
https://trello.com/c/4idzCwO7/712-package-collections-publisher-for-migration


fredericfran-gds (approved):
ok

PR #581, opened 2022-09-16
Use the in-cluster Publishing API instead of EC2.

sengi:
This applies to all environments, but that's ok because it turns out we never had a security group rule to allow traffic from k8s to EC2 publishing-api anyway. So if anything was trying to talk to pubapi it would have been failing already.

In other words, this should not affect the split k8s frontend / EC2 backend setup in the staging and prod k8s environments.


marcpomfret (approved):
LGTM

PR #582, opened 2022-09-20
Enable ELB logging to S3 for router's load balancer in all environments

samsimpson1:
https://trello.com/c/aHggiTgs/1009-enable-alb-logging-to-s3

### DIFF
--- a/charts/argocd-apps/values-integration.yaml
+++ b/charts/argocd-apps/values-integration.yaml
@@ -768,6 +768,7 @@ govukApplications:
         alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
         alb.ingress.kubernetes.io/load-balancer-name: www-origin
         alb.ingress.kubernetes.io/healthcheck-path: /readyz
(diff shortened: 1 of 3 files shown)


nsabri1 (commented):
I wonder if this is only needed in Production

samsimpson1:
@nsabri1 it's enabled in all environments for the LBs for the EC2-based system, and it won't incur much cost since the logs are deleted after a few days in each environment. We may as well have them available in case they're needed

sengi:
Yeah I'd recommend keeping logs in all three environments but just having shorter retention on integration and staging.

nsabri1:
> Member

Ok thats answered my question. will approve thanks for this

nsabri1 (approved):

samsimpson1:
The bucket specified is actually the same one used for the old load balancers. I didn't see the need to create a new one when we could use the one already there. We will need to move it into the new terraform after we decommission the old environments, however.

PR #583, opened 2022-09-20
Revert "Tag images that are in use "

desenhar:
Reverts alphagov/govuk-helm-charts#562

At the moment, the workflows are failing as the github-cli image doesn't have aws-cli and docker packages needed to execute the calls in the new function.
Reverting for the meanwhile while I focus on the fix.


samsimpson1 (approved):
LGTM

PR #585, opened 2022-09-21
Add draft-router-api to ArgoCD.

sengi:
Values that differ from regular router-api come from https://github.com/alphagov/govuk-puppet/blob/6b60080/hieradata_aws/class/draft_cache.yaml

Tested: Helm template parses ok, inspected output with e.g.

    helm template ../govuk-rails-app --name-template=draft-router-api --values <(
        yq e '.|select(.metadata.name == "draft-router-api").spec.source.helm.values'


theseanything (approved):

marcpomfret (approved):
LGTM

PR #586, opened 2022-09-21
Eliminate duplicate secrets for SENTRY_DSN.

sengi:
Sentry DSNs (API keys) are unique per app repo but are not unique per app. For example, `whitehall-admin`, `whitehall-frontend` and `draft-whitehall-frontend` are all the same Sentry project and so they have the same Sentry DSN.

This change makes our k8s Secrets reflect this. We will no longer have duplicate entries in the `govuk/common/sentry` AWS Secrets Manager secret for apps which are deployed multiple times, such as apps with `draft-` versions.

This should also simplify automating fetching the Sentry secrets from the Sentry API during bootstrap, when we come to automate that.

Tested: Helm templates parse ok; inspected output with e.g.

```
helm template ../govuk-rails-app --name-template=draft-router-api --values <(
    yq e '.|select(.metadata.name == "draft-router-api").spec.source.helm.values'
```
and similarly for a handful of representative combinations of `.kind == "ExternalSecret"` and `.metadata.name == "draft-whitehall-frontend"` vs. `whitehall-frontend` etc.

No changes to `values-{staging,production}.yaml` because we don't yet have any `draft-` apps there yet, nor `whitehall-admin`.

Rollout: update the `govuk/common/sentry` secretsmanager secret **in all three environments** to remove duplicates and make the keys match the app repo names, just before merging this.

### DIFF
--- a/charts/argocd-apps/templates/govuk-application.yaml
+++ b/charts/argocd-apps/templates/govuk-application.yaml
@@ -29,6 +29,7 @@ spec:
       # chart's values.yaml.
       values: |
         {{- toYaml $.Values.globalHelmValues | nindent 8 }}
+        repoName: {{ $reponame }}
         appImage:
           repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/{{ $reponame }}
           tag: {{ $imageTag }}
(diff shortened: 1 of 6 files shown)


sengi:
Whoopsee, forgot to update staging and prod values.

theseanything (approved):
Nice!

marcpomfret (approved):
Thanks for this

PR #588, opened 2022-09-21
Disable k8s service links for asset-manager.

sengi:
This declutters the environment by removing the 5 unused env vars for every k8s Service in the namespace. This makes troubleshooting anything to do with environment variables just a little bit more operator-friendly.

We already do this for all the other apps, so this just brings asset-manager into line.


fredericfran-gds (approved):
ok

PR #590, opened 2022-09-23
Use draft backends for draft-content-store.

sengi:
See individual commit messages for context and rationale as to why this particular approach.

Other ideas welcome — it's always possible we've missed a simpler way.

I plan to clean up Plek so that we at least don't have to specify all the non-draft backend addresses on every draft app.

Tested: inspected `helm template` output (see commit messages for example commands)


theseanything (approved):
Nice temp fix - will be really great to fix Plek eventually too!

PR #594, opened 2022-09-26
Add notifications for failed image deployments

theseanything:
This creates a new workflow that wraps update-image-tag with an exit
handler. This exit handler will send a notification if the workflow
fails. This is improves the observability of failure when deployments
are triggered from GitHub whereby only the update-image-tag workflow
runs, not the post-sync workflow.


nsabri1 (approved):
OK

PR #595, opened 2022-09-26
Improve formatting for slack notification for failed workflows

theseanything:
This adds the ability to format slack messages using BlockKit, which makes it easier to add links and improves the UI for messages. This updated the messages to contain link to the failed workflow instead of links ArgoCD, which isn't helpful.


sengi (approved):
Can't really review the code here beyond seeing that it looks vaguely reasonable considering that it has to be YAML spaghetti by its nature 😅 Happy to take your word for it that the end results are much nicer than before! Thanks for polishing this!

PR #596, opened 2022-09-27
Reference non-ec2 secrets names for signon bearer tokens

theseanything:
This updates the secret names for the referenced bearer tokens secrets. We no longer use the ec2 suffixed secrets and they will be removed.


nsabri1 (approved):
OK

PR #597, opened 2022-09-28
Update image tag script for workflow to allow ECR calls to add tags

desenhar:
https://trello.com/c/MK34KlBn/980-lifecycle-rules-for-ecr-to-clean-up-images

Here is the update-image-tag script that is ran whenever a new Image-tag is pushed to ECR, the new function is triggered, the registry is logged into using podman, and the image in question is tagged.
The pod that runs the tagging is assigned a service account that is annotated with a IAM role that is authorised to tag images in production ECR. (see https://github.com/alphagov/govuk-infrastructure/pull/747)

### DIFF
--- a/charts/argo-workflows/scripts/update-image-tag.sh
+++ b/charts/argo-workflows/scripts/update-image-tag.sh
@@ -18,10 +18,10 @@ change_image_tag() {
 }
 
 add_image_deployment_tag() {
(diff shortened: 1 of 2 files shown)


desenhar:
registry ID is the account id for production, I would prefer an alternative way of injecting that value to hardcoding.

PR #602, opened 2022-10-04
Stop overriding `PLEK_SERVICE_*_URI` for every service.

sengi:
Turns out this isn't practical when it comes to configuring the draft stack vs. regular stack x 3 environments. It's way simpler and less error-prone to avoid constructing FQDNs entirely and let the local resolver do its thing. This also paves the way for getting rid of Plek in the future.

So for internal (app-to-app) API traffic, apps will just use short hostnames like `frontend` when calling `getaddrinfo()` and the local resolver will expand that to `frontend.apps.svc.cluster.local.` as nature intended.

We have to keep most of the overrides in staging/production for now, until the second live-traffic experiment is out of the way.

We no longer need `internalDomainSuffix`, so get rid of that.

See https://github.com/alphagov/plek/blob/main/CHANGELOG.md#410 for descriptions of the new Plek env vars.

https://trello.com/c/R86kPYPL

### DIFF
--- a/charts/govuk-apps-conf/templates/configmap.yaml
+++ b/charts/govuk-apps-conf/templates/configmap.yaml
@@ -7,15 +7,13 @@ metadata:
   labels:
     {{- include "govuk-apps-conf.labels" . | nindent 4 }}
 data:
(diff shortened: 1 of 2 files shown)


marcpomfret (approved):
Nice

PR #607, opened 2022-10-08
Refactor app config.

sengi:
Clean up the structure of our config for applications' Helm values to remove some of the ambiguity and arbitrariness that's crept in over the course of its development.

- Rename the `argocd-apps` chart to `app-config` and add a README to explain what is/isn't intended to belong in it.
- Merge the contents of the `govuk-apps-conf` chart into `app-config`. Having two charts for global/shared app configuration was arbitrary and confusing.
- Simplify the way global Helm values are passed to app charts and signpost it better.
- Rename a couple of the configmap templates in `app-config` to make it clearer as to their purpose.

#### Tips for reviewing

- Best reviewed commit-by-commit.
- For generated commits (as opposed to hand-edited ones), focus on whether the command in the commit message looks right to you and whether you see anything unexpected in the diff. (I've already done this myself of course, but please do have a skim through for yourself too.)
- You can evaluate the template output by running e.g. `helm template . --values values-integration.yaml` and diffing before/after etc. `yq` is useful for slicing and dicing the output, for example `| yq e '.|select(.kind == "Deployment")'` to list only Deployments, or `select(.metadata.name == "draft-content-store").spec.source.helm.values` to show the Helm values for a particular app.
- It's safe to ignore any merge conflicts under `charts/argocd-apps/image-tags`; I'll rebase to fix these once this is approved.


theseanything (approved):
Nice - this simplifies things. Not for this PR, but I wonder if we injected the shared env vars directly into the deployment config instead of via a config maps, could be another way to get apps to automatically reload.

sengi:
Rebased to fix merge conflicts in `charts/argocd-apps/image-tags`.

PR #610, opened 2022-10-10
Increment argo-services chart version for rollout.

sengi:
This chart is rolled out via `helm install` (by Terraform) so the chart version matters.


nsabri1 (approved):
LGTM

PR #612, opened 2022-10-10
Rename argo-services and argo-workflows charts

theseanything:
This PR renames the following charts:

- argo-services -> argo-bootstrap
- argo-workflows -> argo-services

argo-bootstrap better conveys its purpose i.e. to be used by terraform to bootstrap ArgoCD with the initial App of Apps. With this we've moved resources from argo-bootstrap to argo-services (previously argo-workflows) that are not needed to bootstrap Argo. This allows those resources to managed by ArgoCD and removes the need to deploy them with Terraform, which requires more effort to manage.

We've renamed argo-workflows to argo-services, as in now contains a more generalised set of resources to do with configuring Argo services (e.g. Argo Events, Argo notifications, Argo Workflows etc..)


sengi (approved):
Cool! LGTM 🚀

PR #615, opened 2022-10-10
Rename `govuk-rails-app` to `generic-govuk-app`.

sengi:
The chart isn't specific to Rails and we're already using it for Router, which isn't even written in Ruby. Hopefully the new name better conveys what's going on with this chart.

Best reviewed commit by commit.

The first commit was generated with

    git grep -l govuk-rails-app |xargs gsed -i 's/govuk-rails-app/generic-govuk-app/g'

(except for one tiny edit in app-config/README.md).


nsabri1 (approved):
LGTM

PR #616, opened 2022-10-10
Install only the Application resources into the Argo namespace.

sengi:
Having moved a bunch of externalsecrets and configmaps into this chart which used to be in their own ArgoCD `Application`, we need those objects to be created in the `apps` namespace.

I think it's simpler and less error prone for the `apps` namespace to be the default and just set `namespace: cluster-services` in the ArgoCD `Application` resource template, rather than having to override the namespace on everything else.

Fixes alphagov/govuk-helm-charts#607.


sengi:
Oh wait no it's not a Terraform thing — sorry, one sec

theseanything (approved):

PR #623, opened 2022-10-13
Remove finalizer from monitoring-config Argo app.

sengi:
This isn't needed (for monitoring config) and just seems to cause deadlocks when deleting or moving the app.

\the_architect{this is the sixth time I've kubectl edited out the finaliser on the monitoring chart and I'm becoming very efficient at it}

#620.

### DIFF
--- a/charts/app-config/templates/monitoring-application.yaml
+++ b/charts/app-config/templates/monitoring-application.yaml
@@ -3,8 +3,6 @@ kind: Application
 metadata:
   name: monitoring-config
   namespace: {{ .Values.argoNamespace | default .Release.Namespace }}
-  finalizers:
-  - resources-finalizer.argocd.argoproj.io
   annotations:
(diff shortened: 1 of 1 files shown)


theseanything:
My understanding is that this finaliser makes Argo delete the App resources if the App is deleted (i.e. a cascaded deletion) - how come this isn't something we want for the monitoring-config App? Seems like a sensible thing to have to make sure resources are cleaned. (Unless its being buggy and just isn't working.)

sengi:
All it seems to do is deadlock when trying to delete the app 🤷

theseanything (approved):

sengi:
\the_matrix_architect{this will be the sixth time we've destroyed it, and we have become exceedingly efficient at it}

sengi:
(my understanding is the same fwiw)

PR #625, opened 2022-10-13
Rightsize response buffers in nginx reverse proxies.

sengi:
This allows:
- the sum of the response headers on any given request to be up to 16 kB (up from 4 kB) without failing with `upstream sent too big header while reading response header`.
- the response body to be up to 384 kB (up from 32 kB) before nginx starts spooling to disk (which is deleterious to performance — we want to avoid disk writes for the vast majority of requests).

This was prompted by alphagov/govuk-puppet#11846.

See
https://www.getpagespeed.com/server-setup/nginx/tuning-proxy_buffer_size-in-nginx for a good explanation of what these parameters mean in practice.

Nginx docs: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffers

### DIFF
--- a/charts/app-config/templates/router-nginx-config.tpl
+++ b/charts/app-config/templates/router-nginx-config.tpl
@@ -20,6 +20,11 @@ http {
   uwsgi_temp_path       /tmp/uwsgi_temp;
   scgi_temp_path        /tmp/scgi_temp;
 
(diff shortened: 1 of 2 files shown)


samsimpson1 (approved):

sengi:
Supporting data:
Response size quantiles 2022-10-13 nq8

PR #629, opened 2022-10-14
Add workflow to retag images with `deployed-to-<env>`

theseanything:
This PR adds the `add-tag-to-image` workflow which can retag an image in AWS ECR with a new tag. This workflow is now called by the deploy image workflow to add a `deployed-to-<env>` tag to the image that's been deployed. This allows us to create AWS ECR lifecycle rules and have a rule to protect images that are currently in use.

This also creates a Service Account the use an IAM Role to gain AWS permissions to push the new image tag. The IAM Role is created in Terraform and the ARN is passed via argo-bootstrap in Helm Values. https://github.com/alphagov/govuk-infrastructure/pull/758

Tested and works by manually syncing with Argo.

### DIFF
--- a/charts/argo-services/values.yaml
+++ b/charts/argo-services/values.yaml
@@ -9,3 +9,7 @@ enableWebhookIngress: false
 rbacTeams:
   read_only:
   read_write:
+iamRoleServiceAccounts:
+  tagImageWorkflow:
+    name: service-account-name
+    iamRoleArn: arn:aws:iam::$account_id:role/$role_name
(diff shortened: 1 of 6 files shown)


charts/argo-services/values.yaml
```diff
@@ -1,3 +1,4 @@
+awsAccountId: "1112223334"
```
nsabri1, line 1: Does there need to be a value added for awsAccountID?

theseanything: There needs to dummy value  here for the tests. When w deploy this the actual AccountID will be passed from Terraform

nsabri1: Maybe a comment here would be good idea.

nsabri1 (approved):
OK

PR #634, opened 2022-10-18
Fix Argo Notifications secret namespace

theseanything:
Argo Notifications cannot access these secrets as they deployed into the apps namespace, rather than same namespaces as Argo Notification and the Argo Applications.


nsabri1 (approved):
LGTM

PR #635, opened 2022-10-18
Fix Argo Notification configmap namespaces

theseanything:
This corrects the namespace references for Argo Events and the configmap used by Argo Notifications.


nsabri1 (approved):
OK

PR #636, opened 2022-10-18
Remove ArgoCD slack notifications

theseanything:
We don't use these at the moment. Removing to simplify configuration.

These are different from the Argo Workflow notification that we currently use to monitor deployments.


nsabri1 (approved):
OK

PR #638, opened 2022-10-19
Adding external secrets.

nsabri1:
Secrets added for the following apps:
  
    content-tagger
    link-checker-api
    local-links-manager
    search-admin
    support-api
    support
    transition

### DIFF
--- a/charts/app-config/templates/external-secrets/basic-auth.yaml
+++ b/charts/app-config/templates/external-secrets/basic-auth.yaml
@@ -1,0 +1,19 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: basic-auth
+  labels:
+    {{- include "app-config.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      GOVUK Basic auth creds.
+spec:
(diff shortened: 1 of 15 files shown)


charts/app-config/templates/external-secrets/link-checker-api/basic-auth.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: link-checker-api-basic-auth
```
sengi, line 4: Are these basic auth secrets just the `betademo` login? If so, we need to have just one secret for them.

nsabri1: I was thinking the same - but was not sure if this was the same in PROD?

nsabri1: meaning do they have same shared user or seperate?

sengi (approved):
As discussed out-of-band, just need to use a single secret for the basic auth since they're all the same secrets as it turns out. Otherwise LGTM - thanks for trawling through these!

PR #639, opened 2022-10-19
Use `$GITHUB_OUTPUT` file instead of deprecated `::set-output`.

sengi:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


nsabri1 (approved):
OK

PR #640, opened 2022-10-19
Use the right OAuth secrets for Authenticating Proxy.

sengi:
Since #574 / #596, the OAuth secrets that are synced automatically from Signon are named after the Signon (Doorkeeper) app names rather than the app repo names.

No changes to staging/prod values needed since we're only running authenticating-proxy in integration at the moment.

Rollout: I've deleted the old secrets in all three clusters so we won't trip over them in future.


nsabri1 (approved):
OK

PR #641, opened 2022-10-19
Adding new publisher apps.

nsabri1:
content-tagger
  link-checker-api
  local-links-manager
  search-admin
  support
  support-api
  transition

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -236,6 +236,39 @@ govukApplications:
           mongo-3.integration.govuk-internal.digital/content_store_production"
       - name: GOVUK_APP_NAME
         value: content-store
+- name: content-tagger
+  helmValues:
+    dbMigrationEnabled: true
+    workerEnabled: true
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -236,6 +236,39 @@ govukApplications:
           mongo-3.integration.govuk-internal.digital/content_store_production"
       - name: GOVUK_APP_NAME
         value: content-store
+- name: content-tagger
+  helmValues:
+    dbMigrationEnabled: true
+    workerEnabled: true
+    workerReplicaCount: 1
```
sengi, line 243: optional: you could set this just once at the top of the file (between lines 5 and 6) instead of for every app with workers.

sengi (approved):
Cool!

charts/app-config/values-integration.yaml
nsabri1: that's actually good idea - i will leave that for a seperate PR, as will require cleanup of other apps also which have that in play.

PR #642, opened 2022-10-19
Eliminate a duplicate secret for the basic auth creds.

sengi:
The same Basic Auth credentials are used in a bunch of other places and we now have a canonical Secrets Manager secret for them as of #638.


nsabri1 (approved):
Thanks Chris

PR #643, opened 2022-10-19
Don't require basic auth on draft-router.

sengi:
The draft flavour of Router sits behind `authenticating-proxy` and doesn't have its own Ingress. It's therefore not supposed to have Basic Auth.

This should fix another thing that was stopping `draft-origin` (content preview) from working.

Rollout: I'll restart draft-router to pick up the new configmap once this is merged.


nsabri1 (approved):
ok

PR #654, opened 2022-10-24
Add support for running arbitrary commands in cron jobs

samsimpson1:
Some apps need to run scripts instead of rake tasks, so we need to support this

https://trello.com/c/ppfM0Sw8/1025-ensure-scheduled-tasks-by-the-whenever-gem-can-run


nsabri1 (approved):
OK

PR #655, opened 2022-10-24
Add cron jobs for all apps that require them

samsimpson1:
https://trello.com/c/ppfM0Sw8/1025-ensure-scheduled-tasks-by-the-whenever-gem-can-run

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -196,6 +196,10 @@ govukApplications:
             key: SECRET_KEY_BASE
 - name: contacts-admin
   helmValues:
+    cronTasks:
+      - name: org-import
+        task: "organisations:import"
+        schedule: "0 3 * * *"
     extraEnv:
       - name: SECRET_KEY_BASE
         valueFrom:
@@ -204,6 +208,10 @@ govukApplications:
             key: SECRET_KEY_BASE
 - name: content-store
   helmValues: &content-store
+    cronTasks:
+      - name: report-delays
+        task: "publishing_delay_report:report_delays"
+        schedule: "15 2 * * *"
     uploadAssets:
       enabled: false
     extraEnv:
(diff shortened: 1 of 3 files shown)


sengi (commented):
Thanks for pulling these out - it's certainly easier to see what's going on.

Couple of thoughts:

- Do any of these jobs assume that they're writing to the same filesystem as the webserver/worker process like in EC2?

- Have you checked that they run successfully? (no need to add monitoring though as they're currently unmonitored)

- Do you have a plan to keep these cronjobs in sync with the applications in the meantime? i.e. if a developer adds/changes/removes a cronjob, causing the cronjobs here to differ from what the app is requesting via the `whenever` gem.

samsimpson1:
> * Do any of these jobs assume that they're writing to the same filesystem as the webserver/worker process like in EC2?

They all appear to use external caches or databases, except [this report creation task](https://github.com/alphagov/publisher/blob/main/lib/tasks/generate_reports.rake#L5). I'm not sure what this report is needed for, so I'll do some investigation.

> * Have you checked that they run successfully? (no need to add monitoring though as they're currently unmonitored)

Not all of them. I was planning on merging this in and letting them run at their usual times, then check them in the morning

> * Do you have a plan to keep these cronjobs in sync with the applications in the meantime? i.e. if a developer adds/changes/removes a cronjob, causing the cronjobs here to differ from what the app is requesting via the `whenever` gem.

Once EC2 is switched off, this repo should become the source of truth for these cron jobs. We shouldn't need to use the whenever gem anymore, since we won't be using it anywhere for crontab generation. Until then, we could definitely get into a situation where more jobs are added and we don't notice, but we can just do another check before switching EC2 off ([sourcegraph](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/alphagov/.*%24+file:%5Econfig/schedule%5C.rb&patternType=standard) makes this very easy)

sengi:
Cool, all SGTM 👍

sengi (approved):

PR #657, opened 2022-10-25
Fix RabbitMQ credentials for publishing-api.

sengi:
I've set the secretsmanager secret in all three environments to match what's in EC2.

Also fix the chart-testing pre-merge check.


nsabri1 (approved):
OK

PR #658, opened 2022-10-26
Adding external secrets for the following apps:

nsabri1:
Content-data-api
    Content-data-admin
    Immience
    Locations-api

### DIFF
--- a/charts/app-config/templates/external-secrets/content-data-admin/google.yaml
+++ b/charts/app-config/templates/external-secrets/content-data-admin/google.yaml
@@ -1,0 +1,21 @@
+{{- if not (has $.Values.govukEnvironment (list "staging" "production")) }}
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: content-data-admin-google
+  labels:
+    {{- include "app-config.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      Google secrets used by Content data admin.
+spec:
+  refreshInterval: {{ .Values.externalSecrets.refreshInterval }}
+  secretStoreRef:
+    name: aws-secretsmanager
+    kind: ClusterSecretStore
+  target:
+    name: content-data-admin-google
+  dataFrom:
+    - extract:
+        key: govuk/content-data-admin/google
+{{- end }}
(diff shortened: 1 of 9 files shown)


charts/app-config/templates/external-secrets/content-data-admin/google.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if not (has $.Values.govukEnvironment (list "staging" "production")) }}
```
sengi, line 1: Can we add them for all the envs so we don't trip over this later?

charts/app-config/templates/external-secrets/content-data-api/rabbitmq.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if not (has $.Values.govukEnvironment (list "staging" "production")) }}
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: content-data-api-rabbitmq
+  labels:
+    {{- include "app-config.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      Credentials used by Content data API to access RabbitiMQ.
```
sengi, line 10:
```suggestion
      RabbitMQ user and password for Content Data API.
```

sengi (approved):

charts/app-config/templates/external-secrets/content-data-admin/google.yaml
nsabri1: Copy and paste error - good spot - yes we can add for all and we should.

charts/app-config/templates/external-secrets/content-data-api/rabbitmq.yaml
nsabri1: ok

charts/app-config/templates/external-secrets/content-data-admin/google.yaml
nsabri1: but i would need to then create the relevant secret for all ENVS

sengi: Yeah that's true, though it's a good idea to do all the envs now because otherwise we'll trip over them later and spend ages debugging unnecessarily (this has been such a bugbear for me the last couple of days trying to troubleshooting the publishing-api/content-store stuff).

nsabri1: i will update them in a separate PR. where i will go through all the app secrets one by one. all ENVS

PR #660, opened 2022-10-26
Adding new apps:

nsabri1:
content-data-admin
content-data-api
locations-api
imminence

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -206,6 +206,147 @@ govukApplications:
           secretKeyRef:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
+- name: content-data-admin
+  helmValues:
+    workerEnabled: true
+    workerReplicaCount: 1
+    extraEnv:
+      - name: CONTENT_DATA_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-content-data-admin-content-data-api
+            key: bearer_token
+      - name: GOOGLE_TAG_MANAGER_ID
+        valueFrom:
+          secretKeyRef:
+            name: content-data-admin-google-tag-manager
+            key: manager-id
+      - name: GOOGLE_TAG_MANAGER_PREVIEW
+        valueFrom:
+          secretKeyRef:
+            name: content-data-admin-google-tag-manager
+            key: manager-preview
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -206,6 +206,147 @@ govukApplications:
...
+    extraEnv:
+      - name: CONTENT_DATA_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-content-data-admin-content-data-api
+            key: bearer_token
+      - name: GOOGLE_TAG_MANAGER_ID
+        valueFrom:
+          secretKeyRef:
+            name: content-data-admin-google
```
sengi, line 222: The naming of these seems a little odd / confusing to me. How about changing `google` to `google-tag-manager`, and removing the `manager-` prefix from the field names?

charts/app-config/values-integration.yaml
```diff
@@ -206,6 +206,147 @@ govukApplications:
...
+- name: content-data-api
+  helmValues:
+    dbMigrationEnabled: true
+    workerEnabled: true
+    workerReplicaCount: 1
+    extraEnv:
+      - name: GOOGLE_ANALYTICS_GOVUK_VIEW_ID
+        valueFrom:
+          secretKeyRef:
+            name: content-data-api-google
```
sengi, line 286: Similarly with this one — it would make sense for it to be `google-analytics` rather than just `google`. And probably just `view-id` instead of `govuk-view-id`, unless there's more than one view ID? 

(Also, should the view ID be part of the secret? I suspect not, as it's not itself a secret and it wouldn't be needed in order to rotate the key.)

charts/app-config/values-integration.yaml
```diff
@@ -206,6 +206,147 @@ govukApplications:
...
+        valueFrom:
+          secretKeyRef:
+            name: content-data-api-postgres
+            key: DATABASE_URL
+      - name: SUPPORT_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-content-data-api-support-api
+            key: bearer_token
+      - name: RABBITMQ_PASSWORD
```
sengi, line 330: Just FYI: these are changing to connection strings; see alphagov/govuk-helm-charts#659 and alphagov/publishing-api#2173.

sengi (approved):
Cool, thanks for trawling through all these!

PR #662, opened 2022-10-27
Remove update image tag logic from set automatic deployments workflow

theseanything:
This workflow should only be concerned with enable/disabling automatic deployments. This PR removes logic from modifying the image tag - that can be done with the update image tag workflow. This functionality was originally introduced to allow user to reset the image tag to main when re-enabling automatic deployments. However, this is an untested need and the maintenance of the extra complexity outweighs the time saved to trigger the update image tag workflow.


sengi (approved):
Cool, this is much nicer - thanks!

PR #669, opened 2022-10-28
Factor out repetitive `workerReplicaCount` value in integration.

sengi:
Also fix a wrong name in the template, which was preventing this Helm value from having any effect.

Tested: inspected output of `helm template ../generic-govuk-app --values <(helm template . --values values-integration.yaml |yq e '. |select(.metadata.name == "publishing-api").spec.source.helm.values') |yq e '.|select(.kind == "Deployment")'`


nsabri1 (approved):
OK

PR #675, opened 2022-11-01
Adding following app

nsabri1:
Content Data API


charts/app-config/templates/signon-secrets-sync-configmap.yaml
```diff
@@ -21,6 +22,7 @@ data:
       "Imminence",
       "Link Checker API",
       "Local Links Manager",
+      "Locations API",
```
sengi, line 25: There doesn't seem to be a `Locations API` app [in Signon](https://signon.publishing.service.gov.uk/doorkeeper_applications).

nsabri1: thanks - for the spot on this Chris

sengi (approved):

PR #676, opened 2022-11-01
Update Locations API Helm values.

nsabri1:


sengi (approved):

sengi:
Yup locations-api doesn't use either of those 👍

PR #680, opened 2022-11-02
Add Content Data Admin to the list off Signon APPS

nsabri1:


charts/app-config/templates/signon-secrets-sync-configmap.yaml
```diff
@@ -10,6 +10,7 @@ data:
       "Asset Manager",
       "Collections publisher",
       "Contacts",
+      "Content Data Admin",
```
sengi, line 13: There's no app by that name in Signon - is it [`Content Data (beta)`](https://signon.publishing.service.gov.uk/doorkeeper_applications/57/edit)?

nsabri1: hmm ok as the app does have a GDS OAUTH secret requirment. https://github.com/alphagov/govuk-puppet/blob/6d5ce8349718e8fdda01e7e5d0b8eec488238871/modules/govuk/manifests/apps/content_data_admin.pp#L156

nsabri1: i double checked the secret for the app in EC2 and https://signon.integration.publishing.service.gov.uk/doorkeeper_applications/63/edit - they match so it is Content Data (beta)

sengi (approved):
👍

PR #682, opened 2022-11-02
Adding AWS credentials for Content data admin.

nsabri1:
Adding external secret belonging to Content data admin.


nsabri1:
> We should be able to eliminate this and grant access via IAM (either to the node instance profile or IRSA if we wanna be fussy). But this is fine for now 👍

I feel like we tried this and it didnt work well previously due to the way the App calls AWS cli - i will look into it though

PR #684, opened 2022-11-03
Added external-secrets and integration values for new apps

marcpomfret:
added external secrets for email-alert-api and email-alert-secrets as well as integration app values


nsabri1 (approved):
OK

PR #687, opened 2022-11-03
Avoid duplicating jobs/resources for draft-government-frontend.

sengi:
Sentry ExternalSecrets and Rails assets are common to the draft and non-draft versions of an app.

Fixes 98f8724 / #685.


nsabri1 (approved):
ok

PR #689, opened 2022-11-04
government-frontend does use memcached after all.

sengi:
Despite there being [no config in govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/cddc3d/modules/govuk/manifests/apps/government_frontend.pp) for it, government-frontend expects to be able to connect to memcache on `localhost:11211` unless `MEMCACHE_SERVERS` is set to an alternative address.

This rather opaque reference is what configures its usage of memcache: https://github.com/alphagov/government-frontend/blob/e2acb52/config/environments/production.rb#L58

Fixes 98f87246.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -799,6 +799,8 @@ govukApplications:
 - name: government-frontend
   helmValues:
     extraEnv:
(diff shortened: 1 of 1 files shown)


dj-maisy (approved):
Ooh, nice find. The elusive Memcached server.... 👍

PR #690, opened 2022-11-04
Sync Static's env vars with Puppet in integration.

sengi:
- Add Google Tag Manager ID.    
- Add values for Google Tag Manager preview functionality.    
- Add bearer token for Static to talk to Publishing API. This is only used in a Rake task.
    
I double-checked with UX Measurement team about the GA/GTM values and they confirmed that they're definitely not supposed to be secrets.

Tested:

- `helm install chris-test-static ../generic-govuk-app --values <(helm template . --values values-integration.yaml |yq e '. |select(.metadata.name == "static").spec.source.helm.values') --set sentry.createSecret=false` (sentry.createSecret=false is just to avoid a name conflict).
- `k get deploy chris-test-static`
- `k logs deploy/chris-test-static`
- `helm uninstall chris-test-static`.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1775,15 +1775,28 @@ govukApplications:
 - name: static
   helmValues:
     extraEnv:
+      # These GA/GTM values are not secrets (not even the the gtm_auth one).
+      # https://github.com/alphagov/govuk-puppet/pull/8041
+      - name: GA_UNIVERSAL_ID
+        value: &ga-universal-id UA-26179049-22
+      - name: GOOGLE_TAG_MANAGER_ID
+        value: &static-gtm-id GTM-MG7HG5W
+      - name: GOOGLE_TAG_MANAGER_AUTH  # Non-prod only.
+        value: &static-gtm-auth C7iYdcsOlYgGmiUJjZKrHQ
+      - name: GOOGLE_TAG_MANAGER_PREVIEW  # Non-prod only.
+        value: &static-gtm-preview env-4
+      - name: PUBLISHING_API_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-static-publishing-api
+            key: bearer_token
+      - name: REDIS_URL
+        value: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
       - name: SECRET_KEY_BASE
         valueFrom:
           secretKeyRef:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
-      - name: REDIS_URL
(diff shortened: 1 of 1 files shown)


charts/app-config/templates/external-secrets/static/google-tag-manager-preview.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: static-google-tag-manager-preview
+  labels:
+    {{- include "app-config.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      gtm_auth and gtm_preview values to enable the Google Tag Manager preview
+      facility in non-production Static.
```
nsabri1, line 11: Is it the case that this secret will no exist in Production?

sengi: That's right — doesn't exist in current Production and won't need to in ours either.

sengi: It's for testing changes to Google Tag Manager configuration without applying them to the live website, hence it's not applicable to production (because there wouldn't be any point in "previewing" changes on the live website, because that's the same as just making the changes live).

charts/app-config/values-integration.yaml
```diff
@@ -1701,17 +1701,32 @@ govukApplications:
 - name: static
   helmValues:
     extraEnv:
+      - name: GA_UNIVERSAL_ID  # Not a secret.
+        value: &ga-universal-id UA-26179049-22
+      - name: GOOGLE_TAG_MANAGER_ID  # Not a secret.
+        value: &static-gtm-id GTM-MG7HG5W
+      - name: GOOGLE_TAG_MANAGER_AUTH  # Non-prod only.
```
sengi, line 1708: Sounds like perhaps this doesn't need to be a secret after all: https://github.com/alphagov/govuk-puppet/pull/8041

I couldn't find anything on the web about what knowing a `gtm_auth` value might allow someone to do, so I've asked UX measurement team.

sengi: Verdict from UX measurement team: not secret 👍

Summary of my understanding of this now: if someone wanted to push fake data to our GA account (which they can do anyway), knowing this value would let them have that data be processed with the GA/GTM configuration that's being previewed on the integration/staging site as opposed to the configuration for the live site. In other words, it's useless for nefarious purposes.

sengi:
RFAL

nsabri1 (approved):
LGTM - always feel weird when the label is relating to Auth that we have it as plain text. As you mentioned in standup you have discussed this with the relevant teams and they have given the green light.

PR #694, opened 2022-11-07
Make GTM values not secret.

sengi:
See comments on #690. Double-checked with UX Measurement team about this.

Rollout: the no-longer-needed Secrets Manager secrets are already scheduled for deletion.


nsabri1 (approved):
Thanks

PR #695, opened 2022-11-07
Improve Asset Manager's ClamAV configuration.

sengi:
- Use a single ConfigMap for both config files, since they belong in the same directory anyway.
- Set size limits to reflect the maximum size of objects that we allow uploading (500 MB). Fail if an object is bigger than that or if it requires more than an additional 700 MB RAM (plus the 500 MB for the object itself) to scan.
- Remove redundant options which weren't actually changing anything from the [default values](https://manpages.ubuntu.com/manpages/jammy/man5/clamd.conf.5.html).
- Don't try to log to syslog; there is no syslog.
- Set `Foreground` so that we never daemonise. We're setting this in the command line anyway, so this is really just to make troubleshooting slightly easier.
- Disable foreground reload (which doubles RAM usage during database updates, so effectively halves our RAM utilization) and freshclam's `TestDatabases` feature (same reason).
- Tune the resource requests/limits a bit, based on observations while testing the above.

https://trello.com/c/moEiT5BG

Tested: `helm install chris-test-assetmgr ../asset-manager --values <(helm template . --values values-integration.yaml |yq e '. |select(.metadata.name == "asset-manager").spec.source.helm.values') --set sentry.createSecret=false`

Rollout: I'll keep an eye out for OOMs etc. over the next few days.

### DIFF
--- a/charts/asset-manager/values.yaml
+++ b/charts/asset-manager/values.yaml
@@ -51,23 +51,23 @@ appResources:
     memory: 1Gi
   requests:
     cpu: 500m
-    memory: 1Gi
+    memory: 0.7Gi
 
 workerResources:
   limits:
     cpu: 1
-    memory: 2Gi
+    memory: 2.5Gi
   requests:
     cpu: 500m
     memory: 2Gi
 
 freshclamResources:
   limits:
-    cpu: 1
-    memory: 2Gi
+    cpu: 500m
+    memory: 0.2Gi
   requests:
     cpu: 500m
-    memory: 2Gi
+    memory: 0.2Gi
 
 clamdResources:
   limits:
@@ -75,7 +75,7 @@ clamdResources:
     memory: 2Gi
   requests:
     cpu: 500m
-    memory: 2Gi
+    memory: 1.5Gi
 
 # assetManagerNFS is the address of the NFSv4 (or Amazon EFS) server where uploaded
 assetManagerNFS: "asset-manager-efs.dev.gov.uk"


nsabri1 (approved):
OK

charts/asset-manager/templates/clamav-configmap.yaml
```diff
@@ -9,29 +9,27 @@ metadata:
...
 data:
   clamd.conf: |
+    # If it's too big to scan, fail it.
+    AlertExceedsMax yes
+    # Avoid RAM spikes on reload. Scanning tasks are async anyway.
+    ConcurrentDatabaseReload no
+    ExtendedDetectionInfo yes
+    Foreground yes
     LogTime yes
-    LogSyslog yes
```
ChrisBAshton: There are quite a few changes happening in the same commit, which would make it trickier to revert if any one of them is an issue. Could you split out each bulletpoint into its own commit?

charts/asset-manager/templates/worker-deployment.yaml
```diff
@@ -20,25 +20,30 @@ spec:
...
+          command: ["bundle"]
+          args: ["exec", "sidekiq", "-C", "config/sidekiq.yml"]
           ports:
             - name: metrics
               containerPort: {{ .Values.metricsPort }}
-          command: ["bundle"]
-          args: ["exec", "sidekiq", "-C", "config/sidekiq.yml"]
+          volumeMounts:
+            - name: asset-manager-efs
+              mountPath: &uploads-path /mnt/asset-manager
```
ChrisBAshton, line 38: Instead of doing both the moving of the YAML block _and_ the change to the mount directory in the same commit, can we split it out to make it easier to follow/review?

charts/asset-manager/values.yaml
```diff
@@ -51,31 +51,31 @@ appResources:
     memory: 1Gi
   requests:
     cpu: 500m
-    memory: 1Gi
```
ChrisBAshton, line 54: In the last commit message, please elaborate on _how_ you did the 'recent testing' and link to any graphs/resources that show the last 7 days' usage?

ChrisBAshton (commented):
Good work - just a few comments below ⭐ 

(thanks for letting us this as a guinea pig)

sengi:
No probs :D lmk if you want me to respond to these btw!

PR #698, opened 2022-11-08
Use `RABBITMQ_URL` throughout instead of separate vars.

sengi:
Using a connection string constructed by a template in the ExternalSecret makes the configuration a bit simpler and more readable.

Also makes it easier to automate rotation of the secrets, should we choose to.

Remove the environment conditions from the ExternalSecrets.

Rollout: secrets are updated to fit the expected format in all three accounts.


marcpomfret (approved):
ok

PR #701, opened 2022-11-08
Remove disused AWS creds secret for email-alert-api.

sengi:
As of alphagov/email-alert-api@fe6ce7a in Nov 2020, email-alert-api no longer uses the AWS API for anything. That commit removed the AWS client library from the build.

Also delete a redundant `workerReplicaCount: 1` while we're there. (It's specified at the top level for all of integration.)


nsabri1 (approved):
OK

PR #706, opened 2022-11-09
Added email-alert-api postgres connection string

marcpomfret:


martynipratt (approved):
Working through Email-Alert-Api think this value is missing as DB attempting local connection

PR #711, opened 2022-11-10
Make container image filesystems read-only.

sengi:
App code shouldn't need to write to the image filesystem.

We're already mounting an emptyDir volume on `/tmp`, so most cases where apps need to write to a local filesystem should already be covered by that. This change might flush out a few as-yet-unnoticed cases of apps writing to other locations, which is a good thing.

For background as to why this is good practice, see for example https://kubesec.io/basics/containers-securitycontext-readonlyrootfilesystem-true/


nsabri1 (approved):
OK - lets see what happens!!

PR #712, opened 2022-11-10
Disable service links env vars consistently everywhere.

sengi:
We're not using these generated `*_SERVICE_{HOST,PORT}` env vars and they just clutter up the environment with reconnaissance info :P


nsabri1 (approved):
ok

PR #714, opened 2022-11-10
Consistently disable API credential automounting for all app pods.

sengi:
None of these containers should need to talk to kube-apiserver, so not expecting any breakage. If I'm wrong then it'll be good to shake out any unexpected stuff here anyway.


nsabri1 (approved):
OK

PR #715, opened 2022-11-10
Fix pod spec template for freshclam jobs.

sengi:
I'd put automountServiceAccountToken and enableServiceLinks in the wrong place in #712, #714.


nsabri1 (approved):
Ok

PR #716, opened 2022-11-10
Add missing volumeMounts for /tmp.

sengi:
Turns out we'd been writing to /tmp inside the image filesystem and getting away with it until recently because we hadn't set readOnlyRootFilesystem.

This should fix most (maybe all?) of the fallout from #711 / d337216.

Also clean up the indentation on all the templates that we're touching. This should help to avoid a repeat of #702. (We need a linter/fixer for this, but that could be a bit tricky because we're dealing with YAML generated by Go templates, rather than just YAML.)

Tested: successfully deployed `asset-manager`, `publisher` and `static` using Helm, e.g. `helm install chris-static ../generic-govuk-app --values <(helm template . --values values-integration.yaml |yq e '. |select(.metadata.name == "static").spec.source.helm.values') --set sentry.createSecret=false`.

### DIFF
--- a/charts/asset-manager/templates/_freshclam_podspec.yaml
+++ b/charts/asset-manager/templates/_freshclam_podspec.yaml
@@ -1,29 +1,29 @@
 {{- define "asset-manager.freshclam.podspec" }}
-  automountServiceAccountToken: false
-  enableServiceLinks: false
-  securityContext:
-    seccompProfile:
-      type: RuntimeDefault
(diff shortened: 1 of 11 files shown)


nsabri1 (approved):
Ok

marcpomfret (approved):
LGTM - Thanks for doing this Chris looks like the indenting could have been a bit of a pain

PR #717, opened 2022-11-11
Add DatabaseMirror to freshclam config

theseanything:
This config is needed to allow clamav to update. Since this was missing the update job was unable to complete blocking Argo syncing the entire application.

### DIFF
--- a/charts/asset-manager/templates/clamav-configmap.yaml
+++ b/charts/asset-manager/templates/clamav-configmap.yaml
@@ -28,6 +28,7 @@ data:
     TCPAddr 127.0.0.1
     TCPSocket 3310
   freshclam.conf: |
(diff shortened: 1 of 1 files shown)


sengi:
Huh, interesting. The [manpage](https://manpages.ubuntu.com/manpages/jammy/man5/freshclam.conf.5.html) says that's already the default.

sengi (approved):

sengi:
Thanks for fixing!

sengi:
The log messages are pretty unequivocal so this is the right fix and the manpage is wrong 🤷 

```
ERROR: No DatabaseMirror or PrivateMirror servers set in freshclam config file.
ERROR: Unable to find DatabaseMirror or PrivateMirror option(s) that specify database server FQDNs.
```

PR #718, opened 2022-11-11
Rename asset ingress from asset-manager to assets-origin

theseanything:
This renames the dns hostname, ingress and able to assets-origin as they will be shared resources to route requests to both Whitehall (for CSV previews) and asset manager.

(This also enables ALB logging with the same conventions we use for www-origin ALB)


marcpomfret (approved):
LGTM

PR #719, opened 2022-11-11
Add missing Nginx config for asset domain

theseanything:
This migrates Nginx config relevant for the asset domain including:

- Setting X-Forwarded-Host header
- Proxy buffer sizes
- Google site verification path

### DIFF
--- a/charts/asset-manager/templates/nginx-configmap.yaml
+++ b/charts/asset-manager/templates/nginx-configmap.yaml
@@ -121,6 +121,10 @@ data:
         error_log /dev/stderr;
 
         location / {
+          proxy_busy_buffers_size 16k;
+          proxy_buffers 8 8k;
+          proxy_buffer_size 8k;
+
           try_files $uri/index.html $uri.html $uri @app;
         }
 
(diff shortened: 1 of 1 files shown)


nsabri1 (approved):
OK

charts/asset-manager/templates/nginx-configmap.yaml
```diff
@@ -119,6 +121,10 @@ data:
         error_log /dev/stderr;
 
         location / {
+          proxy_busy_buffers_size 16k;
```
sengi, line 124: You'll probably want to align this with https://github.com/alphagov/govuk-helm-charts/blob/1230472/charts/generic-govuk-app/templates/nginx-configmap.yaml#L31 otherwise nginx will be needlessly spooling a lot of requests to and from disk (yup, nginx is full of lovely surprises like that 😅)

sengi: (incidentally there should be no need to set `proxy_busy_buffers_size` if you're setting the other two options from that link)

PR #720, opened 2022-11-11
Work around Ruby `Dir.tmpdir` issue with `readOnlyRootFilesystem` and `emptyDir` on /tmp

sengi:
Set `fsGroup` for app containers so that the temp dir we're creating for Ruby ends up with the magic permissions that Ruby [deigns acceptable](https://github.com/ruby/ruby/blob/v2_7_6/lib/tmpdir.rb#L27). See https://github.com/alphagov/govuk-ruby-images/pull/18 and https://github.com/alphagov/govuk-ruby-images/pull/20 for background.

Also:
- Clean up the structure of the Helm templates to make them more consistent so that it's easier to spot inconsistencies and repeated patterns. This should make it easier to factor out repeated components into library charts in future.
- Set a bunch of missing labels which this brought to light.
- Allow kubelet to cache app images by default (`pullPolicy: IfNotPresent`), since they're supposed to be uniquely tagged now.

### DIFF
--- a/charts/app-config/templates/govuk-application.yaml
+++ b/charts/app-config/templates/govuk-application.yaml
@@ -34,6 +34,7 @@ spec:
         appImage:
           repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/{{ $reponame }}
           tag: {{ $imageTag }}
+          pullPolicy: {{ .appImagePullPolicy | default "IfNotPresent" }}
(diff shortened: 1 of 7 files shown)


marcpomfret (approved):
LGTM

sengi:
I've changed this a fair bit since https://github.com/alphagov/govuk-ruby-images/pull/20. RFAL.

PR #721, opened 2022-11-11
Install Argo webhooks into the cluster-services namespace.

sengi:
The webhook servers seem to be failing because they're trying to mount a Secret that's in (presumably correctly) the `cluster-services` namespace but the deployments/pods are being created in the `apps` namespace.

Very much open to alternative fixes here - I don't know whether my proposed fix here is optimal.

Tested: `helm template ../argo-services --values <(helm template . --values values-production.yaml |yq e '. |select(.metadata.name == "argo-services").spec.source.helm.values')` doesn't produce any errors and the output looks plausible.

### DIFF
--- a/charts/app-config/templates/govuk-application.yaml
+++ b/charts/app-config/templates/govuk-application.yaml
@@ -39,7 +39,7 @@ spec:
         {{- end }}
   destination:
     server: https://kubernetes.default.svc
-    namespace: {{ $.Values.appsNamespace }}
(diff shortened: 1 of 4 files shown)


nsabri1 (approved):
OK

sengi:
@theseanything I'm gonna merge this now since rollouts are already broken and this might get them going again, but if it's not the best fix then just shout and I'll happily revert it.

PR #722, opened 2022-11-14
Fix workflow namepsace reference in sensors

theseanything:
The sensors that triggering workflow were referring to the wrong namespace, after the workflows had been moved to the cluster-services namespace. This meant sensors were unable to trigger a new workflow and our deployment process was broken.


sengi (approved):
Aha cool, thanks!

PR #726, opened 2022-11-14
Change asset manager Nginx proxy buffer size

theseanything:
This is to make the proxy buffers consistent with the proxy buffer sizes used by other GOV.UK apps. As suggested by @sengi: https://github.com/alphagov/govuk-helm-charts/pull/719/files/dda86934dd46b83e34f1f9566568db8ca39a36b1#r1020280731


nsabri1 (approved):
Thanks

sengi (commented):
Thanks for sorting these out!

PR #727, opened 2022-11-14
Share ALB for asset domain between Whitehall, Asset Manager and Static

theseanything:
This adds ALB ingress annotations for Asset Manager, Whitehall Frontend and Static to create a IngressGroup so the AWS Load Balancer Controller uses a shared ALB. This allows request on the assets domain to routed to the correct application.


charts/app-config/values-integration.yaml
```diff
@@ -23,9 +23,14 @@ govukApplications:
       enabled: true
       annotations:
         alb.ingress.kubernetes.io/load-balancer-name: assets-origin
+        alb.ingress.kubernetes.io/group.name: assets-origin
+        alb.ingress.kubernetes.io/group.order: '1'
```
sengi, line 27: Might be worth making these `10`, `20`, `30` initially so that we have the option of making future changes without necessarily having to do lots of renumbering. (insert BASIC joke here)

The range is 1-1000 so we have plenty of headroom.

charts/app-config/values-integration.yaml
```diff
@@ -2179,6 +2194,21 @@ govukApplications:
...
     # the other way around like it is now).
+    ingress:
+      rules:
+        - host: assets-origin.eks.integration.govuk.digital
+          path: /government/placeholder/
+        - host: assets-origin.eks.integration.govuk.digital
+          path: /assets/whitehall/
+        - host: assets-origin.eks.integration.govuk.digital
+          path: /government/uploads/system/uploads/attachment_data/file/*/*/preview
+          pathType: ImplementationSpecific
```
sengi, line 2205: Is there a particular reason for specifying `ImplementationSpecific`? Are we overriding it back to the default from something else that's being set elsewhere?

sengi (approved):

charts/app-config/values-integration.yaml
sengi: If it's intended as self-documentation then that's cool btw - was just curious as to whether its also overriding something or not

theseanything: So the default pathType in the [Helm template](https://github.com/alphagov/govuk-helm-charts/blob/5b6c575fb924e329f4b8ada5b6d2da35be8a842d/charts/generic-govuk-app/templates/ingress.yaml#L29) is `Prefix`. This has specific behaviour to modify the path to prefix path rule in the ALB and prevents you using wildcards (i.e. *) in the path. Need to override it with ImplementationSpecific, so we can write an ALB path rule and use wildcards.

More info here: https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1772

charts/generic-govuk-app/templates/ingress.yaml
```diff
@@ -23,7 +23,8 @@ spec:
   {{- end }}
   rules:
     {{- range .Values.ingress.rules }}
-    - http:
+    - host: {{ .host }}
```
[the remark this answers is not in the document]
theseanything: I think so - I've tested the branch out on integration and the assets-origin ALB has the correct cert and so do the other application ALBs.

charts/app-config/values-integration.yaml
sengi: Ah ok, I did wonder if we were overriding the default somewhere. Thanks.

charts/generic-govuk-app/templates/ingress.yaml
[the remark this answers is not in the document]
sengi: Awesome!

PR #729, opened 2022-11-14
Re-enable auto deploys for {manuals,travel-advice}-publisher.

sengi:
These are the last two left out of sync, as far as I can tell.

licence-finder, router, signon and static are not currently meant to be CDed beyond integration, and publishing-api is pinned to a snowflake build in integration while we wait for content-schemas to move into publishing-api on mainline.


nsabri1 (approved):
OK

PR #731, opened 2022-11-15
Add missing $ which was preventing cron tasks from rendering correctly

samsimpson1:


sengi (approved):
Oh yikes, thanks for fixing!

PR #733, opened 2022-11-15
Fix undefined proxy_add_x_forwarded_host in asset-manager nginx.conf.

sengi:
A reference to this was added in f0a205f but without a definition. Added the same definition as in 8d1e20d.


sengi:
No worries!

PR #734, opened 2022-11-15
Run the Signon token sync cron nightly instead of hourly.

sengi:
I couldn't think of much of an advantage to running it hourly now that the configuration is fairly stable.

When it comes to rotating tokens, we'll still need a transition period where old and new token are both valid (which might as well be a day if it's an hour), and it'd be straightforward for any rotation automation to just start a sync job on demand.


nsabri1 (approved):
Looks good to me - just wondered why at this specific time? 01:05

sengi:
No reason - just needed to pick a time (and it's always best to avoid running on the hour). Unfortunately there's no randomisation/fuzz feature in k8s cronjob yet.

PR #736, opened 2022-11-15
Don't retry CronJobs automatically.

sengi:
This makes our behaviour equivalent to the existing system where these jobs are/were run by `crond`.

This helps to prevent things from blowing up when a cronjob starts failing for whatever reason.


nsabri1 (approved):
OK

PR #737, opened 2022-11-16
Sprinkle some deadlines, backoffLimits and common labels.

sengi:
Set deadlines and backoffLimits for pre- and post-sync jobs so that ArgoCD syncs (rollouts) don't get stuck indefinitely. This should help to avoid the situation where a broken pre/post-sync job has since been fixed but Argo is still stuck waiting indefinitely for the broken job to finish and requires someone to go into the UI and manually terminate the sync to unstick it.

Also add the common set of labels to a bunch of objects where they were missing.

Tested: installed asset-manager and collections (generic-govuk-app) using `helm install`, everything still validates, deployments come up healthy, successfully ran the freshclam cronjob using `k create job --from cronjob/...`.


nsabri1 (approved):
LGTM

PR #738, opened 2022-11-16
Set USE_TMPDIR_PAGE_CACHE on static in staging and prod.

sengi:
It needs this flag in order to work with `readOnlyRootFilesystem`.

Tested: installed via `helm install`


nsabri1 (approved):
OK

PR #740, opened 2022-11-16
Prevent content-store from changing Router backend addresses.

sengi:
See https://github.com/alphagov/content-store/pull/1015.

Tested: `helm install`, validates and deployments come up healthy.


nsabri1 (approved):
OK

PR #742, opened 2022-11-17
Remove the disused `govukStack` Helm value.

sengi:
`git grep govukStack` (at the root of this repo) returns nothing.


nsabri1 (approved):
OK

PR #743, opened 2022-11-17
Multiple workers in generic-govuk-app

marcpomfret:
Changed worker-deployment.yaml to allow for multiple workers to be run
Added flag to enable app deployment.

Tested with  

helm install marc-worker ../generic-govuk-app --values <(helm template . --values values-integration.yaml |yq e '. |select(.metadata.name == "email-alert-service" and .kind == "Application").spec.source.helm.values') --set sentry.createSecret=false

### DIFF
--- a/charts/generic-govuk-app/templates/worker-deployment.yaml
+++ b/charts/generic-govuk-app/templates/worker-deployment.yaml
@@ -38,41 +38,43 @@ spec:
           {{- . | toYaml | trim | nindent 8 }}
         {{- end }}
       containers:
-        - name: app
-          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
-          imagePullPolicy: {{ .Values.appImage.pullPolicy | default "Always" }}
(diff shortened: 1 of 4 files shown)


charts/generic-govuk-app/templates/worker-deployment.yaml
```diff
@@ -38,41 +38,42 @@ spec:
...
           ports:
             - name: metrics
-              containerPort: {{ .Values.metricsPort }}
+              containerPort: {{ $.Values.metricsPort }}
           envFrom:
             - configMapRef:
                 name: govuk-apps-env
           env:
-            {{- if .Values.sentry.enabled }}
+            {{- if $.Values.sentry.enabled }}
```
nsabri1, line 53: Sorry just wondering why there is a $ before calling the values from the values files - just have not seen it being used before?

nsabri1 (approved):
LGTM - have you tested it just incase?

sengi (approved):
LGTM 👍

PR #746, opened 2022-11-22
Add uri, query_string and server_protocol to ngnix logs

theseanything:
Previously we only logged the request header which contained the uri, query_string and protocol in a single value making it hard to parse and search. This add individual attributes the logs to make it easier to search on the specific components of the request.

Might be easier to look at the 2nd commit. This first commit just sorts the message structure.


nsabri1 (approved):
Looks good. i assume you tested it

PR #747, opened 2022-11-23
Allow requests from assets domain in assets ALB rules

theseanything:
This adds a condition the ALB rules for asset-origin to allow requests from the public facing asset domain. This is because request are being forwarded from Fastly, where the Host header is set as the original asset domain (not the domain of the alb). This is done by adding an annotation to the ingress resources that is used to AWS ALB ingress controller. T

See: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/ingress/annotations/#conditions

The annotation are put through the template function as the key and value require a Helm value to be rendered.

This has been tested with helm install.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -3,6 +3,7 @@
 govukEnvironment: integration
 externalDomainSuffix: eks.integration.govuk.digital
 publishingServiceDomainSuffix: integration.publishing.service.gov.uk
+assetsDomain: assets-eks.integration.publishing.service.gov.uk
 workerReplicaCount: 1
 appResources:
   limits:
@@ -26,6 +27,8 @@ govukApplications:
         alb.ingress.kubernetes.io/group.name: assets-origin
         alb.ingress.kubernetes.io/group.order: '10'
         alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-integration-aws-logging,access_logs.s3.prefix=elb/assets-origin-eks
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
+          [{"field":"host-header","hostHeaderConfig":{"values":["{{ .Values.assetsDomain }}"]}}]
       hosts:
       - name: assets-origin.eks.integration.govuk.digital
         path: /government/uploads/
@@ -1791,6 +1794,8 @@ govukApplications:
         alb.ingress.kubernetes.io/group.name: assets-origin
         alb.ingress.kubernetes.io/group.order: '30'
         alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-integration-aws-logging,access_logs.s3.prefix=elb/assets-origin-eks
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
+          [{"field":"host-header","hostHeaderConfig":{"values":["{{ .Values.assetsDomain }}"]}}]
     extraEnv:
       # These GA/GTM values are not secrets (not even the the gtm_auth one).
       # https://github.com/alphagov/govuk-puppet/pull/8041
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -26,6 +27,8 @@ govukApplications:
         alb.ingress.kubernetes.io/group.name: assets-origin
         alb.ingress.kubernetes.io/group.order: '10'
         alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-integration-aws-logging,access_logs.s3.prefix=elb/assets-origin-eks
+        alb.ingress.kubernetes.io/conditions.rule-{{ .Release.Name }}: >
+          [{"field":"host-header","hostHeaderConfig":{"values":["{{ .Values.assetsDomain }}"]}}]
```
sengi, line 31: Since this is going through `toYaml` already, might be cleaner to write it as YAML and use `toJson` (it's supposed to end up as JSON, right?)

```suggestion
        alb.ingress.kubernetes.io/conditions.rule-{{ .Release.Name }}: |
          - field: host-header
            hostHeaderConfig:
              values:
                - '{{ .Values.assetsDomain }}'
```

Alternatively could keep it as JSON here but prettify it, though I think it looks marginally clearer as YAML 🤷

```
[
  {
    "field": "host-header",
    "hostHeaderConfig": {
      "values": [
        "{{ .Values.assetsDomain }}"
      ]
    }
  }
]
```

charts/asset-manager/templates/ingress.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     {{- include "asset-manager.labels" . | nindent 4 }}
     app.kubernetes.io/component: web
   annotations:
-    {{- toYaml .Values.ingress.annotations | nindent 4 }}
+    {{- (tpl (toYaml .Values.ingress.annotations) .) | trim | nindent 4 }}
```
sengi, line 12: Does the controller actually accept YAML, or was this only working because toYaml just happens to preserve JSON syntax? More robust/clearer to use `toJson` explicitly?
```suggestion
    {{- (tpl (toJson .Values.ingress.annotations) .) | trim | nindent 4 }}
```

sengi (approved):
LGTM - just a minor clarity thing about JSON vs YAML

charts/asset-manager/templates/ingress.yaml
theseanything: Just tested and the controller doesn't accept YAML, so has to be annotation value for the condition has to be JSON.

Was using `toYaml` to preserve the YAML structure for the rest of the annotations.
If we use `toJson`, we get:
```yaml
  annotations:
    {"alb.ingress.kubernetes.io/conditions.release-name":"- field: host-header\n  hostHeaderConfig:\n    values:\n      - 'assets-eks.integration.publishing.service.gov.uk'\n","alb.ingress.kubernetes.io/group.name":"assets-origin","alb.ingress.kubernetes.io/group.order":"11","alb.ingress.kubernetes.io/healthcheck-path":"/healthcheck/ready","alb.ingress.kubernetes.io/listen-ports":"[{\"HTTP\": 80}, {\"HTTPS\": 443}]","alb.ingress.kubernetes.io/load-balancer-attributes":"access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-integration-aws-logging,access_logs.s3.prefix=elb/assets-origin-eks","alb.ingress.kubernetes.io/load-balancer-name":"assets-origin","alb.ingress.kubernetes.io/scheme":"internet-facing","alb.ingress.kubernetes.io/ssl-redirect":"443","alb.ingress.kubernetes.io/target-type":"ip"}
```
Which then get reformatted back to YAML when applied to the cluster anyways, and the annotation value is actually a string so stays as YAML (which doesn't work).

charts/app-config/values-integration.yaml
theseanything: As per the reply below - but the value has to be JSON. The `toJson` template method doesn't format this because its seen as just a multiline string.

charts/asset-manager/templates/ingress.yaml
sengi: Oops, yes of course - I didn't consider that it's on the whole set of annotations and not just the new one. My suggestion doesn't make sense, sorry.

PR #748, opened 2022-11-23
Update path to content schemas for Pub API

theseanything:
The content schemas have been merged into the same repository as the Publishing API, so should be accessible within the app.

https://github.com/alphagov/publishing-api/pull/2184


sengi (approved):
🎉

PR #750, opened 2022-11-24
Add Asset Manager to Staging and Production environments

theseanything:
This adds the config to deploy to tell Argo to deploy Asset Manager in the Staging and Production environments. Essentially copied from the values-integration.yaml file.

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -509,6 +509,18 @@ govukApplications:
   chartPath: charts/smokey
 - name: static
   helmValues:
+    ingress:
+      rules:
+        - host: assets-origin.eks.production.govuk.digital
+          path: /
+      enabled: true
+      annotations:
+        alb.ingress.kubernetes.io/load-balancer-name: assets-origin
+        alb.ingress.kubernetes.io/group.name: assets-origin
+        alb.ingress.kubernetes.io/group.order: '30'
+        alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-production-aws-logging,access_logs.s3.prefix=elb/assets-origin-eks
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
+          [{"field":"host-header","hostHeaderConfig":{"values":["{{ .Values.assetsDomain }}"]}}]
     extraEnv:
       # These GA/GTM values are not secrets.
       - name: GA_UNIVERSAL_ID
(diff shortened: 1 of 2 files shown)


charts/app-config/values-production.yaml
```diff
@@ -3,6 +3,7 @@
 govukEnvironment: production
 externalDomainSuffix: gov.uk
 publishingServiceDomainSuffix: publishing.service.gov.uk
+assetsDomain: assets-eks.publishing.service.gov.uk
```
sengi, line 6: For testing, this looks good. For the traffic split, we'd presumably need to change this from `assets-eks` to `assets`, right?

charts/app-config/values-production.yaml
```diff
@@ -15,6 +16,64 @@ replicaCount: 3
 # Individual apps to be deployed by ArgoCD, and any values specific to those
 # apps.
 govukApplications:
+- name: asset-manager
+  chartPath: charts/asset-manager
+  helmValues:
+    workerEnabled: true
```
sengi, line 22: This is ok as long as we're prepared to support it. Otherwise it might be better to turn off workers in prod until launch (but would need a checklist item so we don't forget!)

charts/app-config/values-production.yaml
```diff
@@ -15,6 +16,64 @@ replicaCount: 3
...
+          secretKeyRef:
+            name: signon-app-asset-manager
+            key: oauth_id
+      - name: GDS_SSO_OAUTH_SECRET
+        valueFrom:
+          secretKeyRef:
+            name: signon-app-asset-manager
+            key: oauth_secret
+      - name: GOVUK_ASSET_ROOT  # testing in isolation
+        value: https://assets-origin.eks.production.govuk.digital
```
sengi, line 55: I can't remember what this one does. Is it set right?

sengi (approved):
🚀

charts/app-config/values-production.yaml
theseanything: yup - we'll need to switch that over once we're done testing

charts/app-config/values-production.yaml
theseanything: Ah good shout - assume we only need the workers for publishing assets. Will turn them off for now and add a reminder.

charts/app-config/values-production.yaml
theseanything: I actually don't know what this does in the context of asset manager - it's used by plek, but asset manager doesn't generate links to static (I think ?🤷🏻‍♂️) 

https://github.com/alphagov/plek/blob/484196157989ba5cb385957538daf963f3bbf087/README.md?plain=1#L59

and it's not used within the Asset Manager codebase. So it can either get removed, or at least should be set to `assets.publishing.service.gov.uk`, instead of the internal domain.

PR #751, opened 2022-11-25
Add extra nginx conf for Static

theseanything:
This adds extra nginx configuration for static to set up a redirect for the crown logo and robots.txt. These rules are used by the asset domain. The configuration has been copied from [govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/templates/static_extra_nginx_config.conf.erb).

### DIFF
--- a/charts/generic-govuk-app/templates/nginx-configmap.yaml
+++ b/charts/generic-govuk-app/templates/nginx-configmap.yaml
@@ -168,6 +168,11 @@ data:
         location = /readyz {
           return 200 'ok\n';
         }
+
+
+        {{- if .Values.nginxConfigMap.extraConf }}
+        include       /etc/nginx/extra-nginx.conf;
+        {{- end }}
       }
     }
 {{- end }}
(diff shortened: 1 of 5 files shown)


charts/generic-govuk-app/templates/nginx-configmap.yaml
```diff
@@ -168,6 +168,11 @@ data:
         location = /readyz {
           return 200 'ok\n';
         }
+
+
+        {{- if .Values.nginxConfigMap.extraConf }}
```
sengi, line 173: Apologies if you've already considered this and it's doesn't fly for whatever reason, but wouldn't it be simpler/cleaner to just pass the actual extra config here instead of going to all the trouble of creating and including a separate file?

theseanything: Ahah yeah 😅 I think I was probably over thinking it bit. I was worried if the pattern got used other apps, which had much more config. I'll do another PR just passing it in via a Value.

theseanything:
Closed in favour of https://github.com/alphagov/govuk-helm-charts/pull/754

charts/generic-govuk-app/templates/nginx-configmap.yaml
sengi: Ah ok cool, no probs!

PR #754, opened 2022-11-25
Add extra nginx conf for Static

theseanything:
This adds extra nginx configuration for static to set up a redirect for the crown logo and robots.txt. These rules are used by the asset domain.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1823,6 +1823,13 @@ govukApplications:
             key: SECRET_KEY_BASE
       - name: USE_TMPDIR_PAGE_CACHE
         value: "true"
+    nginxConfigMap:
+      extraConf: |
+        # 1stline use this URL in their zendesk template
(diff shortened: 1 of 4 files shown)


samsimpson1 (approved):
LGTM

charts/app-config/values-integration.yaml
```diff
@@ -1823,6 +1823,17 @@ govukApplications:
...
+    nginxConfigMap:
+      extraConf: |
+        # 1stline use this URL in their zendesk template
+        location = /static/gov.uk_logotype_crown.png {
+          absolute_redirect off;
+          return 301 /media/5c810ef4ed915d43e50ce260/gov.uk_logotype_crown.png;
+        }
+
+        location /robots.txt {
+          expires 1w;
```
sengi, line 1835: Will this have any effect? It looks like we're currently serving robots.txt from Router's nginx.

theseanything: Removed.

PR #756, opened 2022-11-28
Add ALB host header condition for www

theseanything:
This adds a check that the host header of requests to the Router match GOV.UK domains. This is to add a layer of protection against cache poisoning and leaking internal domains.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1432,6 +1432,8 @@ govukApplications:
         alb.ingress.kubernetes.io/load-balancer-name: www-origin
         alb.ingress.kubernetes.io/healthcheck-path: /readyz
         alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-integration-aws-logging,access_logs.s3.prefix=elb/www-origin-eks
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1432,6 +1432,8 @@ govukApplications:
         alb.ingress.kubernetes.io/load-balancer-name: www-origin
         alb.ingress.kubernetes.io/healthcheck-path: /readyz
         alb.ingress.kubernetes.io/load-balancer-attributes: access_logs.s3.enabled=true,access_logs.s3.bucket=govuk-integration-aws-logging,access_logs.s3.prefix=elb/www-origin-eks
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
+          [{"field":"host-header","hostHeaderConfig":{"values":["www.{{ .Values.externalDomainSuffix }}", "www.{{ .Values.publishingServiceDomainSuffix }}"]}}]
```
sengi, line 1436: Do we not also need www-origin in these?

sengi (approved):

charts/app-config/values-integration.yaml
theseanything: The ALB adds the host specified in the rules to the list allowed host headers - so should be okay.

theseanything: (oh if you mean "www-origin.publishing.service.gov.uk"? Think that'd send traffic to the router ELB in the old infra anyway - we wouldn't be splitting that traffic)

sengi: Aha cool 👍

PR #760, opened 2022-11-29
Update memory limit for Finder-frontend from 1500Mi to 2000Mi

nsabri1:


charts/app-config/values-production.yaml
```diff
@@ -190,7 +190,7 @@ govukApplications:
     appResources:
       limits:
         cpu: 4
-        memory: 1500Mi
+        memory: 2000Mi
       requests:
         cpu: 2
         memory: 1Gi
```
sengi, line 196: Might wanna increase the request as well. Maybe 1500?

sengi (approved):

charts/app-config/values-production.yaml
nsabri1: thats sensible done

sengi (approved):

charts/app-config/values-production.yaml
```diff
@@ -190,10 +190,10 @@ govukApplications:
     appResources:
       limits:
         cpu: 4
-        memory: 1500Mi
+        memory: 2000Mi
       requests:
         cpu: 2
-        memory: 1Gi
+        memory: 1500mi
```
sengi, line 196: This needs to be big M otherwise you're requesting 1500 millibytes.
```suggestion
        memory: 1500Mi
```

sengi (changes requested):

sengi (approved):

PR #761, opened 2022-11-29
Fix url namespace for argo-events-service

theseanything:
Post sync notification were broken since we changed the namespace for the argo workflows. The service is now in the cluster-services namepsaces instead of apps.


nsabri1 (approved):
OK

sengi (approved):

PR #762, opened 2022-11-29
Change smokey namespace to cluster-services

theseanything:
Post sync workflow are currently failing to trigger smokey because they are in a different namespace. This make smokey and post sync workflows to be in the same namespace.


samsimpson1 (approved):

nsabri1 (approved):
OK

PR #767, opened 2022-11-30
Use SECRET_KEY_BASE which matches EC2 for frontend

samsimpson1:


charts/app-config/values-integration.yaml
```diff
@@ -833,7 +833,7 @@ govukApplications:
       - name: SECRET_KEY_BASE
         valueFrom:
           secretKeyRef:
-            name: rails-secret-key-base
+            name: frontend-rails-secret-key-base
```
sengi, line 836: Might as well use it for `draft-frontend` too.

sengi (approved):
Cool — good call on splitting out just this one for now. Thanks!

PR #768, opened 2022-11-30
Use SECRET_KEY_BASE which matches EC2 for email-alert-frontend

samsimpson1:


charts/app-config/templates/external-secrets/email-alert/rails-secret-key-base.yaml
```diff
@@ -0,0 +1,21 @@
...
+kind: ExternalSecret
+metadata:
+  name: email-alert-rails-secret-key-base
+  labels:
+    {{- include "app-config.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      Rails SECRET_KEY_BASE, primarily used for encrypting Rails session
+      cookies. For email-alert-frontend specifically, this matches the
+      EC2 secret so Sign On works correctly during a traffic split.
```
sengi, line 11:
```suggestion
      EC2 secret so Sign In works correctly during a traffic split.
```

sengi (approved):

PR #769, opened 2022-11-30
Move Argo Workflows back into the `apps` namespace

theseanything:
This PR moves the relevant components to run our workflows back into the "apps" namespace. This is needed because the post sync workflow cannot trigger the smoke tests because they run in different namespaces. (Smoke tests are configured in a separate helm chart and are dependent on resources in the "apps" namespace).

This was simpler than trying to move smokey test to the cluster-services namespace (due dependent resource) or trying to use a sensor to trigger a workflow across namespace (as they add complexity and potential for breakage).

To move the web hook resources to "apps", we need a copy of the argocd-notifications-secret in both the "apps" and "cluster-services" namespace - as it is also used by the core argo-notification service (which is apart of argocd installed in cluster-services).

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -78,7 +78,6 @@ govukApplications:
         value: govuk-assets-integration
 - name: argo-services
   chartPath: charts/argo-services
-  namespace: cluster-services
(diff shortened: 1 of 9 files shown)


sengi (approved):
LGTM

PR #772, opened 2022-12-01
Fix allowed host headers for www

theseanything:
The external domain prefix is the real domain in production vs the test domain in integration and staging. This adds an explicit value for the test domain and adds it to the allowed host headers. Also uses the publishing suffix for integration, as the external domain is currently the test domain.


nsabri1 (approved):
OK

PR #794, opened 2022-12-14
Fix template error in Ingress spec for Maslow.

sengi:
Took me ages to spot what was wrong here 😅 Missed this in #744.

Tested: `helm template ../generic-govuk-app --values <(helm template . --values values-integration.yaml |yq e '.|select(.metadata.name=="maslow").spec.source.helm.values')`


nsabri1 (approved):
LGTM

PR #799, opened 2022-12-16
Added additional worker process

marcpomfret:
Extra worker process added for search-api and content-data-api

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -280,6 +280,13 @@ govukApplications:
       enabled: false
     dbMigrationEnabled: true
     workerEnabled: true
+    workers:
+      - command: ["sidekiq", "-C", "config/sidekiq.yml"]
+        name: worker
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -280,6 +280,13 @@ govukApplications:
       enabled: false
     dbMigrationEnabled: true
     workerEnabled: true
+    workers:
+      - command: ["bundle", "exec", "sidekiq", "-C", "config/sidekiq.yml"]
```
sengi, line 284: optional: with the new base images, the `bundle exec` should now be optional (but I haven't actually tried it with sidekiq yet)

sengi (approved):

charts/app-config/values-integration.yaml
marcpomfret: cool, i'll take it out and give it a try with this then

PR #804, opened 2022-12-21
Adding ingress to remaining publisher apps

nsabri1:
manuals-publisher
service-manual-publisher
travel-advice-publisher


sengi (approved):
🚀

PR #805, opened 2022-12-21
Removing custom assets upload path from Manuals-publisher

nsabri1:


nsabri1:
> Just checking is this PR still valid - given the discussion in [alphagov/manuals-publisher#2003](https://github.com/alphagov/manuals-publisher/pull/2003)?

thanks for checking - this will still be valid

PR #825, opened 2023-01-04
Adding draft-service-manual-frontend

nsabri1:
While testing user journey for service-manual-publisher, discovered missing draft-service-manual-frontend.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1667,6 +1667,21 @@ govukApplications:
           secretKeyRef:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
+- name: draft-service-manual-frontend
+  repoName: service-manual-frontend
+  helmValues:
+    sentry:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -553,6 +553,21 @@ govukApplications:
         value: "https://licensify.integration.govuk-internal.digital"
       - name: BACKEND_URL_search-api
         value: "https://search-api.integration.govuk-internal.digital"
+- name: draft-service-manual-frontend
```
sengi, line 556: nit: could you move this underneath `service-manual-frontend` please so that it's easier to see the differences?

sengi (approved):

charts/app-config/values-integration.yaml
nsabri1: thought we was ordering alphabetically? - but yes that would be beneficial - happy to do that if it makes it easier

PR #827, opened 2023-01-05
Omit `bundle exec` from worker commands.

sengi:
`bundle exec` is no longer necessary with recent govuk-ruby-images.

Tested: inspected template output

```sh
helm template publisher ../generic-govuk-app --values
  <(helm template . --values values-integration.yaml \
    | yq e '.|select(.metadata.name=="publisher").spec.source.helm.values' \
  ) 
```

```sh
helm template asset-manager ../asset-manager --values
  <(helm template . --values values-integration.yaml \
    | yq e '.|select(.metadata.name=="asset-manager").spec.source.helm.values' \
  )
```


nsabri1 (approved):
LGTM

PR #831, opened 2023-01-05
Add missing Ingresses for some publisher apps.

sengi:
contacts-admin, content-data-admin, imminence and local-links-manager are all (content management) frontends for user requests.

This should hopefully get a bunch more Smokey probes passing.

Tested: no errors running `helm template . --values values-integration.yaml`


nsabri1 (approved):
Thanks

PR #834, opened 2023-01-06
Add app config for content-publisher in integration.

sengi:
Secrets Manager secrets are created in all three accounts.

Container image build: https://github.com/alphagov/content-publisher/pull/2705
IAM permissions for it to access its S3 bucket: https://github.com/alphagov/govuk-infrastructure/pull/815

Tested: deployed via `helm install`, pods come up healthy.


nsabri1 (approved):
LGTM

PR #836, opened 2023-01-06
Set max request body size for file uploads.

sengi:
This matches the 500 MB limit that we currently set for these via govuk-puppet. It fixes things like image uploads on content-publisher.

nginx's default is 1M, which should be plenty, so keep that for everything else.

In govuk-puppet this is set to 20M for all the other apps (i.e. apps that don't handle file uploads), but I intentionally haven't set that here because there should be no legit reason for those apps to receive such huge request bodies; it's just a (small but unnecessary) resource-exhaustion attack risk. I couldn't find any justification in the govuk-puppet commit history as to why it was set so large.

Similarly, I couldn't find any explanation as to why search-api had this set to 500 MB so I'm leaving that at the default 1M too.

### Testing

Inspected `helm template` output for an app with/without an override of the new value:

```sh
helm template content-publisher ../generic-govuk-app --values <(
    yq e '.|select(.metadata.name=="content-publisher").spec.source.helm.values'
)
```

```sh
helm template frontend ../generic-govuk-app --values <(
    yq e '.|select(.metadata.name=="frontend").spec.source.helm.values'
)
```

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1681,6 +1681,7 @@ govukApplications:
             key: DATABASE_URL
 - name: search-api
   helmValues:
+    nginxClientMaxBodySize: 20M
     uploadAssets:
       enabled: false
     workerEnabled: true
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -363,6 +363,7 @@ govukApplications:
 - name: content-publisher
   helmValues:
     dbMigrationEnabled: true
+    nginxClientMaxBodySize: &max-upload-size 500M
     workerEnabled: true
     ingress:
       enabled: true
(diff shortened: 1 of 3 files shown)


martynipratt (approved):
Looks good to me

sihugh:
I was thinking about the search-api thing. Whitehall still has [some content types](https://github.com/alphagov/whitehall/blob/039a31eeb2e491601bb11960e512b30dce608783/app/presenters/rummager_presenters.rb#L1) that are submitted directly to search, as opposed to the standard approach which is search-api finding out via a message on a queue and then retrieving what it needs.
I don't know what the largest thing that Whitehall still submits is, but wonder if the 50MB was a finger in the air from someone at some point to make this safe?

sengi:
> I was thinking about the search-api thing. Whitehall still has [some content types](https://github.com/alphagov/whitehall/blob/039a31eeb2e491601bb11960e512b30dce608783/app/presenters/rummager_presenters.rb#L1) that are submitted directly to search, as opposed to the standard approach which is search-api finding out via a message on a queue and then retrieving what it needs. I don't know what the largest thing that Whitehall still submits is, but wonder if the ~50MB~ 500MB was a finger in the air from someone at some point to make this safe?

Oh that's good to know, thanks! I'll loosen this a bit for search-api then so it won't fail if there's a large (> 1MB) content item.

PR #837, opened 2023-01-06
Add account-api

marcpomfret:
Add external secret for oauth and add account-api values to integration.


martynipratt (approved):
Looks good adding for new account api service in eks

PR #849, opened 2023-01-09
Configure Bouncer in integration.

sengi:
- Add ArgoCD app config for Bouncer.
- Allow passing arbitrary Ingress rules into generic-govuk-app. (It's not possible to specify `host: '*'`.)
- Fix a quoting issue in the govuk-generic-app Ingress template.

Related: https://github.com/alphagov/bouncer/pull/383

Tested: `helm install bouncer ../generic-govuk-app --values <(helm template . --values values-integration.yaml |yq e '.|select(.metadata.name=="bouncer").spec.source.helm.values')`, app comes up healthy.

### DIFF
--- a/charts/generic-govuk-app/values.yaml
+++ b/charts/generic-govuk-app/values.yaml
@@ -7,15 +7,20 @@ service:
 
 ingress:
   enabled: false
+  # ingress.annotations defines the base set of annotations for an app's Ingress.
+  #
+  # When deployed via ArgoCD using ../charts/app-config, these annotations are
+  # merged with the ones from the app's config in
+  # ../charts/app-config/values-$env.yaml.
+  #
+  # TODO: this feels a bit messy and unintuitive; move these into app-config
+  # and make it less surprising.
   annotations:
     alb.ingress.kubernetes.io/scheme: internet-facing
     alb.ingress.kubernetes.io/target-type: ip
     alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
     alb.ingress.kubernetes.io/ssl-redirect: "443"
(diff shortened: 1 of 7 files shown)


charts/generic-govuk-app/values.yaml
```diff
@@ -7,15 +7,6 @@ service:
 
 ingress:
   enabled: false
-  annotations:
```
sengi, line 10: These defaults weren't being used anywhere and they would have caused a problem with the addition to ingress.yaml on lines 22-24, hence getting rid of them now. It's also potentially confusing to have AWS-specific stuff being set in the chart defaults, especially when it's actually unused.

theseanything (approved):

charts/generic-govuk-app/values.yaml
nsabri1: if i am right in saying the Healthcheck path is used for assigning path in the target group of an App
example:
https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#TargetGroup:targetGroupArn=arn:aws:elasticloadbalancing:eu-west-1:210287912431:targetgroup/k8s-apps-servicem-3b7164aa5a/c03dc7568155497f

sengi: Oh that's a great point — it looks (to my surprise) that those annotations were actually being merged with the per-app ones from `../app-config/values-*.yaml` and we're currently relying on that.

I've put the annotations back in here. (Tried setting them at the top level in `../app-config/values-integration.yaml` but that doesn't work without extra template faff, so I've just stuck a TODO in for now.)

Fixed.

sengi: It was only really the `rules` that need to go for the purpose of this PR.

nsabri1 (approved):

PR #850, opened 2023-01-10
Adding ingress to search-admin and removing custom asset path

nsabri1:
Search-admin needs an ingress.

This will add ingres to search admin.

Also removing custom asset path:
https://github.com/alphagov/search-admin/pull/904


sengi:
Could you amend the commit title to say which app?

PR #863, opened 2023-01-17
Set GOVUK_NOTIFY_RECIPIENTS to prevent emails being set to "@example.…

martynipratt:
…org". 

It appears that email alert api is trying to send emails to @example.org email addresses. Setting GOVUK_NOTIFY_RECIPIENTS variable looks like it overrides this behaviour.


sengi (approved):
Nice find! 😅

PR #871, opened 2023-01-25
Remove separate rule Nginx for assets with digest

theseanything:
This separate rules isn't needed as all assets are (/should be) immutable and have unique paths for cache busting. We can make the cache-control header conistent for all requests to assets.


sengi (approved):
Cool, thanks for tidying this up!

theseanything:
@sengi - I left the Surrrogate-Key header in just in case someone is using it (and also was only an extra line). But agree we probably don't need it.

PR #872, opened 2023-01-25
Fix aws-lb-controller Ingress annotations for Bouncer.

sengi:
Bouncer can't inherit our default set of aws-load-balancer-controller Ingress annotations because it doesn't serve TLS. (TLS is terminated at Fastly for a few transitioned sites where it's been enabled.)

Tested: installed Bouncer and Publisher (as an example of an app whose ingress should stay the same) via `helm install` and checked that the ingresses were created and reconciled successfully.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -19,6 +19,13 @@ appResources:
     cpu: 0.1
     memory: 800Mi
 
+default-alb-ingress-annotations: &default-alb-ingress-annotations
+  alb.ingress.kubernetes.io/scheme: internet-facing
+  alb.ingress.kubernetes.io/target-type: ip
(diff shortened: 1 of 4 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1691,6 +1716,7 @@ govukApplications:
     ingress:
       enabled: true
       annotations:
+        <<: *default-alb-ingress-annotations
         alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
         alb.ingress.kubernetes.io/load-balancer-name: www-origin
         alb.ingress.kubernetes.io/healthcheck-path: /readyz
```
theseanything, line 1722: Can we remove this as it's now in `default-alb-ingress-annotations`?

sengi: Oh yeah well spotted, thanks! Done!

theseanything (approved):

PR #880, opened 2023-01-26
Add AWS_REGION to support-api

marcpomfret:


martynipratt (approved):
Brings inline with Support service

PR #881, opened 2023-01-26
Send deployment information to Release

theseanything:
This adds an Argo Workflow to send deployment information to the Release App. This is a very MVP version of getting deployment information to display in Release. This using the existing Deployment API to notify Release of deployments, however it might be better in future if Release could get deployment information directly from Kubernetes or Prometheus.

Another downside of this implementation, is that it records a deployment when the image tag is updated in the GitHub repo. This does not necessarily guarantee that the new image was deployed to K8s correctly. 

At least this give more visibility into deployment in EKS compared to the old infra.

Note: more work needs to be in Release to support showing deployments based on commit SHA instead of tags. (As we don't have git tag information in EKS)

### DIFF
--- a/charts/argo-services/templates/workflows/notify-release/workflow.yaml
+++ b/charts/argo-services/templates/workflows/notify-release/workflow.yaml
@@ -21,7 +21,7 @@ spec:
         command:
           - sh
         source: >
-          curl https://release.publishing.service.gov.uk/deployments \
+          curl http://release/deployments \
            -H "Authorization: Bearer ${RELEASE_API_TOKEN}" \
            --json '{
               "repo": "alphagov/{{"{{inputs.parameters.repositoryName}}"}}",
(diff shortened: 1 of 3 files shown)


charts/argo-services/templates/workflows/notify-release/workflow.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - name: environment
+  templates:
+    - name: notify-release
+      inputs:
+        parameters:
+        - name: repositoryName
+        - name: commitSha
+        - name: environment
+      script:
+        image: curlimages/curl
```
sengi, line 20: No need to worry about it in this PR, but we should fetch these through ECR (pull-through cache maybe?) so that we don't get flakiness from DockerHub's rate-limiting.

sengi (approved):

charts/argo-services/templates/workflows/notify-release/workflow.yaml
theseanything: Good shout - on the to do list to look at!

PR #886, opened 2023-01-27
Refactor Helm domain values

theseanything:
This refactors the Helm values being used to reference domains.

Changes are explained in commit messages - probably easier to review commit by commit.


sengi (approved):
This is so much clearer, thanks!

PR #888, opened 2023-01-27
Use `assets-eks.$env.publishing...` during testing.

sengi:
We'll have to revert this before launch, but until then I think we need this in order for Smokey probes to hit the k8s stack instead of getting redirected back to EC2.

### DIFF
--- a/charts/app-config/templates/env-configmap.yaml
+++ b/charts/app-config/templates/env-configmap.yaml
@@ -37,7 +37,8 @@ data:
   RAILS_LOG_TO_STDOUT: "true"
   SENTRY_CURRENT_ENV: {{ .Values.govukEnvironment }}-eks
 
-  PLEK_SERVICE_ASSETS_URI: https://assets.{{ .Values.publishingDomainSuffix }}
+  # TODO: change testAssetsDomain to assetsDomain before launch.
+  PLEK_SERVICE_ASSETS_URI: https://{{ .Values.testAssetsDomain }}
   PLEK_SERVICE_DRAFT_ASSETS_URI: https://draft-assets.{{ .Values.publishingDomainSuffix }}
 
   # Services which remain in EC2 for a while after "MVP launch".


charts/app-config/templates/env-configmap.yaml
```diff
@@ -37,7 +37,8 @@ data:
   RAILS_LOG_TO_STDOUT: "true"
   SENTRY_CURRENT_ENV: {{ .Values.govukEnvironment }}-eks
 
-  PLEK_SERVICE_ASSETS_URI: https://assets.{{ .Values.publishingServiceDomainSuffix }}
+  # TODO: revert assets-eks to assets before launch.
+  PLEK_SERVICE_ASSETS_URI: https://assets-eks.{{ .Values.publishingServiceDomainSuffix }}
   PLEK_SERVICE_DRAFT_ASSETS_URI: https://draft-assets.{{ .Values.publishingServiceDomainSuffix }}
```
theseanything, line 42: Should we also be pointing to the draft assets on the test domain?

theseanything (approved):
I did PR to tidy up the helm values https://github.com/alphagov/govuk-helm-charts/pull/886 - using a value that explicitly references a test domains (e.g. `testAssetsDomain`) might make it easier to clean up the test domains after launch?

sengi:
> I did PR to tidy up the helm values #886 - using a value that explicitly references a test domains (e.g. `testAssetsDomain`) might make it easier to clean up the test domains after launch?

Oh that's an excellent idea. Let's go with that — I'll rebase onto your PR.

charts/app-config/templates/env-configmap.yaml
sengi: I had the same thought but it looks like we don't have one (yet). Maybe we should? (I guess all it needs is a CNAME and corresponding hostname on the ingress?)

theseanything: Ah yeah - I guess we don't use Fastly for the draft domains. Anyways can sort that out later!

sengi:
RFAL (I can do the draft one after this if we need it.)

PR #893, opened 2023-02-03
Add GOVUK_DATA_SYNC_PERIOD env

theseanything:
This env var is used by GovukError module to ignore errors in the defined period that might be caused by the data sync process taking place.


sengi (approved):

sengi:
Thanks!

PR #895, opened 2023-02-06
Add missing app specific Nginx config

theseanything:
There specific bits of Nginx config relevant for certain apps - this adds:
- Rate limiting for email-alerts-api's public endpoints
- Increase timeouts and body size to handle large attachments in specialist publisher

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -782,6 +782,17 @@ govukApplications:
             key: SECRET_KEY_BASE
       - name: WEB_CONCURRENCY
         value: '10'
+    nginxConfigMap:
+      extraConf: |
+        location ~ ^/status-updates(/|$) {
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2377,6 +2388,15 @@ govukApplications:
         value: 759acac6-da53-4a19-b591-b7538c7c39de
       - name: REDIS_URL
         value: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
+    nginxConfigMap:
+      extraConf: |
+        client_max_body_size 500m;
```
sengi, line 2393: nit: best to write this as M rather than m for least surprise. While nginx treats both as mega, in k8s m is milli and M is mega.

sengi (approved):
Thanks!

PR #899, opened 2023-02-07
Add Google site verification to nginx

marcpomfret:
Removed config from asset-manager nginx and added to static only for all envs, tested and working in integration

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2238,6 +2238,10 @@ govukApplications:
           absolute_redirect off;
           return 301 /media/5c810ef4ed915d43e50ce260/gov.uk_logotype_crown.png;
         }
+        # Google site verification
+        location = /googlec908b3bc32386239.html {
+          return 200 'google-site-verification: googlec908b3bc32386239.html';
+        }
 
 - name: draft-static
   repoName: static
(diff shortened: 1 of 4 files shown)


sengi (commented):
Which domain does this need to be served on?

theseanything:
@marcpomfret is this going to enable to verification for all domains? I think it's only needed for the asset domain? i.e. https://www.gov.uk/googlec908b3bc32386239.html doesn't work

sengi:
If it's for the assets domain then it'll be the [asset-manager nginx config](https://github.com/alphagov/govuk-helm-charts/blob/main/charts/asset-manager/templates/nginx-configmap.yaml) that you want.

sengi:
Oh, sorry for the bad steer there. It is indeed going to static.

sengi:
I don't think it's going via Router, though. At least not the way we've got it set up at the moment.

sengi:
It's a little tricky to see from the config but static has an ingress that's sharing an ALB with asset-manager (and whitehall-frontend). The catch-all path goes to static, so I think this request will be going from assets-origin ALB straight to the static Service.

marcpomfret:
yeah got it, added an extra bit to static and the = in the other config, tested it out works now

sengi (approved):

PR #900, opened 2023-02-07
Fix missing limit_req_zone/status for email alert api rate limiting

theseanything:
We were missing Nginx config for email alert api to correctly configure rate limiting. The existing rate limiting rules required the shared memory zone to be declared and we also missed the response status. 

These directives needed to be added into the http context, so I've added a new `extraHttpConfig` value that can be used to add app specific http nginx config. Also renamed the existing value for extraConfig to make it clear that its for the server context. 

(Tested doing helm template)


martyni (approved):
This approach seems necessary to more finely customize the nginx config

martyni (approved):
^^

martyni (approved):
^^

PR #902, opened 2023-02-07
Fix post sync smokey env vars

theseanything:
This adds the missing PLEK_USE_HTTP_FOR_SINGLE_LABEL_DOMAINS and points to the eks test domains for assets.

It also:
- reorders the env vars to group similar ones
- uses the testAssetDomain helm value in the govuk-app-env config map (doesn't change the end result)


sengi (approved):
Thanks! I should have kept these in sync, sorry.

PR #905, opened 2023-02-08
Don't auto-refresh dashboards by default.

sengi:
It can be frustrating when dashboards refresh unexpectedly, plus it's quite wasteful of resources to leave on by default.

Meant to do this in ad3d030 but hadn't spotted the right field in the JSON.


marcpomfret (approved):
OK

PR #906, opened 2023-02-08
Disable cache for draft stack apps

theseanything:
Draft apps shouldn't need a cache as the request rate is usually much lower (as the pages aren't public). This also fixes a cache poisioning bug caused by the normal and draft version governement frontend sharing the same cache. The page template for both the draft and published pages were being cached with same cache key, which caused the wrong page template being shown draft envs.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1034,8 +1034,6 @@ govukApplications:
     uploadAssets:
       enabled: false
     extraEnv:
-      - name: MEMCACHE_SERVERS
-        value: frontend-memcached-govuk.eks.integration.govuk-internal.digital
       - name: PLEK_HOSTNAME_PREFIX
         value: draft-
       - name: SECRET_KEY_BASE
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2820,7 +2816,10 @@ govukApplications:
           [{"field":"host-header","hostHeaderConfig":{"values":["{{ .Values.assetsDomain }}", "{{ .Values.testAssetsDomain }}"]}}]
     sentry:
       createSecret: false  # Sentry DSNs are per repo.
-    extraEnv: *whitehall-envs
+    extraEnv:
+      - *whitehall-envs
```
sengi, line 2820: Does YAML merge lists like that?

sengi (approved):
Does the output for whitehall look ok?

theseanything:
Did the helm template - and the output showed the lists merged. (I was surprised, but seemed to work)

theseanything:
I'd removed the anchor tags and just duplicated the config.

PR #907, opened 2023-02-09
Remove draft Whitehall Frontend and unused env vars

theseanything:
This removes draft Whitehall Frontend app as we don't run a separate instance of Whitehall for the draft stack. Request from the origin domain are routed to the non-draft version of Whitehall Frontend.

This also removes environment variables from Whitehall Frontend aren't used, as they are only relevant for Whitehall Admin.


nsabri1 (approved):
OK

charts/app-config/values-integration.yaml
```diff
@@ -2880,79 +2844,10 @@ govukApplications:
           secretKeyRef:
             name: rails-secret-key-base
             key: SECRET_KEY_BASE
-      - name: MEMCACHE_SERVERS
```
sengi, line 2883: Does whitehall-frontend not use memcache?

sengi (approved):

charts/app-config/values-integration.yaml
sengi: Looks like it doesn't 🤷👍

```
$ sshp whitehall_frontend:1 -- govuk_setenv whitehall env |grep MEMC
...
Running command: ssh -J jumpbox.production.govuk.digital <...>.eu-west-1.compute.internal govuk_setenv whitehall env

$ 
```

sengi: Ohh wait, there is a memcache running locally on the `whitehall_frontend` machines. Is the app defaulting to using that local memcached?

sengi: Confirmed

```
irb(main):002:0> Rails.cache
=> 
#<ActiveSupport::Cache::MemCacheStore:0x00007f7581afe8a0                                            
 @coder=ActiveSupport::Cache::MemCacheStore::Coders::Rails70Coder,                                  
 @coder_supports_compression=true,                                                                  
 @data=                                                                                             
  #<Dalli::Client:0x00007f7581afdc20                                                                
   @key_manager=                                                                                    
    #<Dalli::KeyManager:0x00007f7581afd310                                                          
     @key_options={:digest_class=>Digest::MD5},                                                     
     @namespace=nil>,                                                                               
   @normalized_servers=["127.0.0.1:11211"],                                                         
```

theseanything: > ~Looks like it doesn't 🤷👍~
> ```
> $ sshp whitehall_frontend:1 -- govuk_setenv whitehall env |grep MEMC
> ```

Ah yeah, that's why I assumed no, but you're right should have checked for the local memcache. Will add back the env var.

PR #908, opened 2023-02-09
Remove namespace prefix from Prometheus job labels.

sengi:
- Make the `app.kubernetes.io/name` selector label consistent with the `app` selector label.
- Remove the redundant namespace prefix from the `job` Prometheus label.
- Fix existing dashboard queries to reflect the new job labels.

We don't want the `job` labels in Prometheus to be prefixed with the namespace, because this complicates queries in dashboards etc. There's already a `namespace` label.

In order to fix this, we first have to fix the inconsistency in our use of selector labels, specifically our `app` label vs the well-known (but ambiguously defined) `app.kubernetes.io/name`. We intended for these to be the same and their main use case is to select pods belonging to a given Deployment (e.g. all worker pods for a particular app, as opposed to all pods in all deployments of a particular app, which it's harder to see a case for).

Using `app.kubernetes.io/name` (as opposed to our `app` label) for the Prometheus job label has the advantage that it should work for monitoring third-party stuff as well as our own, without the need for special-case configuration via extra PodMonitors etc.

For now, we still have one or more PodMonitors per app, but this we we're able to simplify that in the (hopefully near) future.

Recommend reviewing commit-by-commit.

Output diff for asset-manager and publisher: https://gist.github.com/sengi/0e06ff62581961513bba6ea4d4c52df8


nsabri1 (approved):
LGTM

PR #910, opened 2023-02-09
Add draft-whitehall-frontend service

theseanything:
This adds a template to the govuk-generic-app chart to allow the creation of additional service for the same application pod. This is then used by Whitehall frontend to add a draft-whitehall-frontend service, to make the page available on the draft domain without running a separate instance of Whitehall (as pages rendered by Whitehall don't support draft versions).


theseanything:
There is no draft Whitehall-frontend.

PR #911, opened 2023-02-09
Route draft requests for whitehall frontend to live

theseanything:
This configures draft router to send requests for whitehall frontend to the live version of the app (rather than a draft version). This is becuase whitehall doesn't render pages that can be drafts, yet we still want those pages to be available on the draft domain.


sengi (approved):
Might be worth a comment here just so people don't wonder if it's a mistake 😅 (edit: meant to write this in-line rather than as a general comment)

PR #913, opened 2023-02-10
Enable memcache for local-links-manager and maslow.

sengi:
These two apps use memcached in the EC2 environments, but we'd missed this when configuring them in Kubernetes because they magically connect to `localhost:11211` instead of having `MEMCACHE_SERVERS` set explicitly.

Enable memcache so that we don't cause performance regressions for these apps.

Neither stores a lot of data in memcache or does a lot of throughput, and as of https://github.com/alphagov/maslow/pull/953 they both namespace their keys appropriately. So it's fine for them to use the shared memcache.

I suppose this makes the name "frontend-memcached" less than a perfect fit, but I think it's still a reasonable name because frontend is a suitably ambiguous term and these two apps are still frontends in all but GOV.UK's unusual usage of the term. (They're UIs and they're the first thing in the request path after the load balancer, so they're frontends.)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1272,6 +1272,8 @@ govukApplications:
           secretKeyRef:
             name: local-links-manager-link-checker-api-callback-token
             key: LINK_CHECKER_API_SECRET_TOKEN
+      - name: MEMCACHE_SERVERS
+        value: frontend-memcached-govuk.eks.integration.govuk-internal.digital
       - name: SECRET_KEY_BASE
         valueFrom:
           secretKeyRef:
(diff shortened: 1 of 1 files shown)


theseanything (approved):

theseanything:
I wonder what the additional risk is for the public frontend apps being effected by issues from our publishing applications? (e.g. publishing app exhausting memcached resources, or misconfiguration causing cache poisoning). Is this risk large enough to warrant having a separate shared memcached for the public facing frontends and our publishing frontends?

sengi:
> I wonder what the additional risk is for the public frontend apps being effected by issues from our publishing applications? (e.g. publishing app exhausting memcached resources, or misconfiguration causing cache poisoning). Is this risk large enough to warrant having a separate shared memcached for the public facing frontends and our publishing frontends?

So I looked into this quite a bit and both the size of the data and the throughput involved are small compared to the "frontend" frontends, so I think the chance of noisy neighbour issues is minimal and it's not worth the operational cost of running another Elasticache.

(If for whatever reason we really wanted apps to have separate memcaches anyway, then it'd be better to do that in-cluster rather than manage lots of Elasticaches. I don't think we need it though.)

sengi:
The risk of name clashes is essentially zero now that the last app (Maslow) that wasn't prefixing keys with its name has been fixed.

PR #914, opened 2023-02-10
Add draft asset hostname to asset manager

theseanything:
This enable the draft asset domain on EKS, these request go directly to the live version of asset manager (i.e there's no draft version of asset manager). Authentication for access to drafts is done by asset manager, rather than auth proxy.


sengi (approved):
Awesome!

PR #916, opened 2023-02-19
Switch authenticating-proxy from Mongo to Postgres.

sengi:
Reflects https://github.com/alphagov/authenticating-proxy/pull/350.

Rollout: need to add the corresponding Secrets Manager secret in all three accounts.


marcpomfret (approved):
ok

PR #918, opened 2023-02-22
Omit ingress hostnames from ALB conditions annotation.

sengi:
The controller adds the hostnames from `ing.spec.rules.host[]` itself and fails with `ValidationError: Condition values must be unique` if the same hostname is present in a `hostHeaderConfig` via `alb.ingress.kubernetes.io/conditions.*`.

This means `alb.ingress.kubernetes.io/conditions` now contains only those hostnames which we want ALB to recognise but for which we **don't** want external-dns controller to create DNS records. `ing.spec.rules.host[]` remains the same: it contains the hostnames for which we want external-dns to create records (as well as being recognised by ALB).

Fixes a bug I introduced in https://github.com/alphagov/govuk-helm-charts/pull/915.

Kudos to @nsabri1 for finding the issue.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -126,9 +126,7 @@ govukApplications:
               "{{ .Values.assetsDomain }}",
               "{{ .Values.testAssetsDomain }}",
               "assets-origin.{{ .Values.publishingDomainSuffix }}",
-              "draft-assets.{{ .Values.publishingDomainSuffix }}",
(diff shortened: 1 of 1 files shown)


sengi:
Hmm on further investigation I'm not so sure the controller authors intended `alb.ingress.kubernetes.io/conditions` to be used at the same time as `ing.spec.rules.host`. I'll do a bit more testing to check what the controller actually does with these specs.

theseanything (approved):

nsabri1 (approved):
OK

PR #919, opened 2023-02-23
Add custom location rule for businesslink.

marcpomfret:
This custom location rule for businesslink looks to be the only bit of the config from the bouncer vhost setup that's still in use and not just relying on the standard bouncer config.

Links to old config:
https://github.com/alphagov/govuk-puppet/blob/c71acac58d8bbc3c317ae1b035f021cf843cfa3a/modules/govuk/manifests/apps/bouncer.pp
https://github.com/alphagov/govuk-puppet/blob/c71acac58d8bbc3c317ae1b035f021cf843cfa3a/modules/govuk/templates/bouncer_nginx_vhost.conf.erb

https://trello.com/c/EBy5Fsxd/1123-check-bouncer-nginx-config-matches-old-infra

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -267,6 +267,12 @@ govukApplications:
           secretKeyRef:
             name: bouncer-postgres
             key: DATABASE_URL
+    nginxConfigMap:
+      extraServerConf: |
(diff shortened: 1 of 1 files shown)


sengi (approved):
👍

PR #920, opened 2023-02-24
Use same Rails secret key base as applications in EC2

theseanything:
This refactors the configuration for Rails secret key base for applications. 

For go live, Rails applications need to have the same secret key base as the existing applications. This is because the key base is used to encrypt session cookies to prevent session hijacking. Whilst traffic is shared between the application running in EC2 and EKS, they need the same key base to share session cookies.

This creates an external secret per app that looks up a the key base secret from a single secret stored in AWS Secret Manager. Draft version of application use the same secret key base as the live versions, hence don't need to create a separate external secret. This only applies to applications using Rails.

(Tested in integration - AWS Secret Manager updated to contain relevant secrets for all envs)


charts/app-config/values-production.yaml
```diff
@@ -87,11 +87,6 @@ govukApplications:
             key: MONGODB_URI
       - name: REDIS_URL
         value: redis://shared-redis-govuk.eks.production.govuk-internal.digital
-      - name: SECRET_KEY_BASE
```
sengi, line 90: Oh that's awesome that this gets rids of that bit of copypasta in the app values :D

charts/generic-govuk-app/templates/rails-secret-key-base-external-secret.yml
```diff
@@ -0,0 +1,24 @@
+{{- if (and .Values.rails.enabled .Values.rails.createKeyBaseSecret) }}
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: {{ .Values.repoName }}-rails-secret-key-base
```
sengi, line 5: I guess we need a default for repoName to satisfy kubeconform? I'm not sure where kubeconform is currently reading the Helm values from though. Presumably the default values.yaml for the generic-govuk-app chart?

sengi (approved):
Nice!

charts/generic-govuk-app/templates/rails-secret-key-base-external-secret.yml
theseanything: Yeah, so ended up adding that values file. I was worried about it overriding [the default where repoName = name](https://github.com/alphagov/govuk-helm-charts/blob/dfd5a9a53ad1bc938dd07a4e5bff49776ef44764/charts/app-config/templates/govuk-application.yaml#L2) - but then realised that is calculated separately.

PR #922, opened 2023-03-01
Run a single parent process for email-alert-service.

sengi:
This makes it much simpler to monitor. See
https://github.com/alphagov/email-alert-service/pull/621


dj-maisy (approved):
Good stuff. Let's give this a go. 👍

PR #924, opened 2023-03-01
Set Rails SECRET_KEY_BASE for workers too.

sengi:
content-publisher-worker was failing without SECRET_KEY_BASE.

We effectively stopped defining SECRET_KEY_BASE for workers in #920, but it's needed in at least some cases and it always would have been defined for workers in the old EC2/Puppet setup.

Tested:

```
helm install chris-content-pub ../generic-govuk-app --values <(
    yq e '.|select(.metadata.name=="content-publisher").spec.source.helm.values'
  ) --set sentry.enabled=false --set rails.createKeyBaseSecret=false
k get all -l'app in (content-publisher,content-publisher-worker)'
...
helm uninstall chris-content-pub
```

### DIFF
--- a/charts/generic-govuk-app/templates/worker-deployment.yaml
+++ b/charts/generic-govuk-app/templates/worker-deployment.yaml
@@ -54,6 +54,13 @@ spec:
             - configMapRef:
                 name: govuk-apps-env
           env:
+            {{- if $.Values.rails.enabled }}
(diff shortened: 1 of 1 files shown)


martynii (approved):
LGTM

martynipratt (approved):

PR #929, opened 2023-03-02
Remove obsolete mongo config for imminence

KludgeKML:
Imminence has been running on Postgres in both the old estate and EKS for a few months now, so remove redundant config.


sengi (approved):
Thank you!

PR #931, opened 2023-03-02
Clean up integration values before copying to staging.

sengi:
- We no longer need to override `GOVUK_ASSET_ROOT` for asset-manager.
- Factor out some repeated, opaque values for readability.
- Simplify an unnecessarily-repeated bit of nginx config.
- Update Content Store `DEFAULT_TTL` to match what's in EC2/Puppet.


nsabri1 (approved):
OK

PR #933, opened 2023-03-03
Set nginx timeouts to match EC2/Puppet + tune healthcheck probes.

sengi:
- Set nginx `proxy_read_timeout` the same as EC2/Puppet for a few apps where it was different.
- Apply the 30s liveness/readiness/startup probe timeout (up from 15s) to all apps rather than just Whitehall, since it's simpler and probably slightly safer. Saves Whitehall being a special-case.

Bit more detail in the commit messages.


samsimpson1 (approved):

sengi:
Thanks for the super-quick review! :D

PR #935, opened 2023-03-03
Remove long-lived AWS creds for support-api.

sengi:
We're using EC2 instance profile creds now (and we might switch to IRSA in future).

See also:
- https://github.com/alphagov/support-api/pull/760
- https://github.com/alphagov/govuk-infrastructure/pull/823


dj-maisy (approved):
🗑️

PR #936, opened 2023-03-03
Merge integration config into staging.

sengi:
This adds all the publisher and API apps into the staging config, as well as merging all the fixes for frontend apps' config that we've made since the production experiments.

Resource requirements for the frontend apps are taken from production, since we tuned them there during the live traffic experiments.

Some notable things that I looked into, which may be helpful when reviewing:

- Opaque values that vary by environment but are hard to find by search/replace, e.g. Notify `TEMPLATE_ID`s, `ELASTICSEARCH_URI`.
- Some email addresses (vars containing `REPLY` or `EMAIL` or `ADDRESS`) differ by more than just replacing the environment name
- We no longer need to override `GOVUK_ASSET_ROOT` in asset-manager
- Google Analytics / Tag Manager stuff (e.g. `GOOGLE_*`, `GA_*`)
- Vars with names like `OVERRIDE` tend to vary by environment
- Resources / quotas / limits:
    - `WEB_CONCURRENCY`
    - `appResources`
     - `replicaCount` (doesn't need to be the same as prod, but should be more than one so that there's at least a chance of shaking out concurrency bugs)
- Timeouts:
    - `{liveness,readiness,startup}Probe`
    - `nginxProxyReadTimeout`
- Checked TODOs for anything that's supposed to differ by environment.
- Checked security groups / ACLs are appropriate (e.g. signon ALB ingress security group, anything to do with basic-auth / `HTTP_USERNAME` etc.)
- "Non-prod only" stuff stays, but we'll need to look out for this when we come to production.
- `runAsUser`/`runAsGroup` for asset-manager and whitehall-admin: checked that the uid and gid are the same in all EC2/Puppet envs.
- `EXPOSE_GOVSPEAK` is supposed to be enabled in integration only (not staging or production).

### DIFF
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -13,6 +13,8 @@ testAssetsDomain: assets-eks.staging.publishing.service.gov.uk
 
 cspReportURI: https://csp-reporter.staging.publishing.service.gov.uk/report
 
+replicaCount: 2
+workerReplicaCount: 2
 appResources:
   limits:
     cpu: 1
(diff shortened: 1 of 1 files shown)


dj-maisy (approved):
It's a big diff and a little hard to review for this reason, but if there's anything wrong we should be able to spot it fairly quickly. 👍

PR #938, opened 2023-03-05
Remove some footcannons from Signon app and API user names.

sengi:
The inconsistencies in naming of Signon Doorkeeper apps and API users is troublesome when trying to reconcile config between integration and staging/production. The worst of these were cases where live apps were using tokens belonging to API users of retired apps, or where a live app was using tokens belonging to multiple different API users.               
                                                                  
I've fixed these by making changes on the Signon side. This commit reflects those changes.

The idea is to reduce the chance of future confusion as well as eliminating problematic renames in future (for example, getting rid of "beta" from names of secrets gets rid of some unnecessary brittleness).

Fix wrong Signon token name for content-data-api.


marcpomfret (approved):
Nice, should be less confusing

PR #940, opened 2023-03-05
Keep 2 previous replicasets rather than 5.

sengi:
We don't expect to use `kubectl rollout undo` (since we're using ArgoCD for almost everything and Helm for the rest) and even in the unlikely event that we did want to use it, rolling back more than two revisions is itself unlikely. Having lots of old revisions is therefore not very helpful — it just clutters and slows down the ArgoCD UI and makes unnecessary requests to apiserver and etcd.

This change means we'll see at most 3 ReplicaSets per Deployment rather than 6.


marcpomfret (approved):
Yeah, makes sense

PR #942, opened 2023-03-08
Fix conflict when applying ExternalSecrets templates via ArgoCD.

sengi:
Set [deletionPolicy=Delete](https://external-secrets.io/v0.7.2/guides/ownership-deletion-policy/) on ExternalSecrets, so that when Argo CD deletes an ExternalSecret and re-creates it with a different uid, the controller deletes Secret that was owned by the old ExternalSecret and is then able to create the new Secret without finding a conflict of ownership.

This hopefully fixes the problem where ExternalSecrets often shows as "degraded" in Argo CD when replacing an Application resource.

Also make refreshInterval configurable consistently. (I've intentionally kept the Helm value names the same throughout, so that things don't break if we move things between charts.)

All the changes except the ones to values.yaml were generated with:

    git grep -l ExternalSecret external-secrets asset-manager generic-govuk-app smokey |
      xargs gsed -i '/  target:/a \ \ \ \ deletionPolicy: {{ .Values.externalSecrets.deletionPolicy }}'

    git grep -l ExternalSecret asset-manager generic-govuk-app smokey |
      xargs gsed -i 's/refreshInterval: 1h/refreshInterval: {{ .Values.externalSecrets.refreshInterval }}/'

### DIFF
--- a/charts/asset-manager/templates/rails-secret-key-base-external-secret.yaml
+++ b/charts/asset-manager/templates/rails-secret-key-base-external-secret.yaml
@@ -10,11 +10,12 @@ metadata:
     kubernetes.io/description: >
       The Rails secret key base for the application.
 spec:
-  refreshInterval: 1h
+  refreshInterval: {{ .Values.externalSecrets.refreshInterval }}
   secretStoreRef:
     name: aws-secretsmanager
     kind: ClusterSecretStore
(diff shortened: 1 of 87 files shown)


nsabri1 (approved):
LGTM

PR #943, opened 2023-03-08
Rename `test*Domain*` to `k8s*Domain*` for clarity.

sengi:
Two of these three Helm values aren't going to go away after go-live. They're more than just for "testing"; they represent the intermediate service domain names that are Internet-facing but not end-user-facing. They're the equivalent of `*.$environment.govuk.digital.` in the old EC2 setup.

For example, we'll switch the publisher apps from EC2 to Kubernetes by changing the foo.publishing.service.gov.uk CNAMEs to point to `foo.k8sExternalDomainSuffix`.

All the changes here apart from the first diff hunk of values-*.yaml were generated by:

    git grep -l testExternalDomainSuffix |
      xargs gsed -i 's/testExternalDomainSuffix/k8sExternalDomainSuffix/'
    git grep -l testAssetsDomain |
      xargs gsed -i 's/testAssetsDomain/k8sAssetsDomain/'
    git grep -l testPublishingDomainSuffix |
      xargs gsed -i 's/testPublishingDomainSuffix/k8sPublishingDomainSuffix/'


marcpomfret (approved):
ok

PR #945, opened 2023-03-09
Increment argo-bootstrap version number.

sengi:
This is one of the few charts where the version number matters, because we're installing it using Terraform and Helm rather than Argo CD. Forgot this in #944. Ugh.


marcpomfret (approved):
ok

PR #946, opened 2023-03-10
Simplify monitoring-config chart.

sengi:
This should make it way easier to maintain alerts, dashboards and [recording rules](https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules).

We don't want to have to copy/paste boilerplate Helm templates just to add a dashboard file or a new set of alert rules.

The Helm templates in `monitoring-config/templates` are now simply for collating the static json files under `dashboards/` and `rules/` into the k8s objects expected by Prometheus Operator. They should rarely, if ever, need to change.

Dashboards go in `dashboards/*.json` and are literally just JSON saved from the Prometheus UI. No templating.

Alerts and recording rules go `rules/*.yaml`. Tests for `rules/foo.yaml` go in `rules/foo_tests.yaml`. No templating here either (at least for now — we can see how we go once we start adding recording rules etc.).

The ruletests still don't [run automatically](https://trello.com/c/9eBmgWhP), yet; that'll be another PR.

Using Prometheus Operator for this stuff probably isn't giving us any value in the first place, so it might make sense to switch everything to plain ConfigMaps, but that'd be a bigger change which could use a bit more thought.


martynipratt (approved):
Seems like the best approach. App config in source control never looks pretty but at least we have the state captured.

PR #947, opened 2023-03-10
Create external DNS name for Bouncer.

sengi:
For other services, we're using the hostnames on the Ingress to tell the external-dns controller to create/update the corresponding DNS names. Bouncer has to serve thousands of hostnames, so we have its Ingress rules match any hostname. We therefore configure external-dns for Bouncer explicitly via [an annotation](https://www.github.com/kubernetes-sigs/external-dns/blob/43d29f8/docs/initial-design.md#configuration).


martynipratt (approved):
LGTM

PR #949, opened 2023-03-10
Add ingress for account-api.

sengi:
Apparently Account API serves requests directly to the general public as well as being an internal API.

It'd be nice to have clear separation between the internal and public API surfaces, but there isn't any such separation in the existing EC2/Puppet/Terraform deployment so 🤷

### DIFF
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -49,6 +49,23 @@ govukApplications:
       enabled: false
     dbMigrationEnabled: true
     workerEnabled: true
+    ingress:
+      enabled: true
+      annotations:
+        <<: *default-alb-ingress-annotations
+        <<: *alb-ingress-www-waf-ruleset
+        alb.ingress.kubernetes.io/load-balancer-name: account-api
+        alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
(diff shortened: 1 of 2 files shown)


charts/app-config/values-staging.yaml
```diff
@@ -49,6 +49,23 @@ govukApplications:
...
+    ingress:
+      enabled: true
+      annotations:
+        <<: *default-alb-ingress-annotations
+        <<: *alb-ingress-www-waf-ruleset
+        alb.ingress.kubernetes.io/load-balancer-name: account-api
+        alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
+        alb.ingress.kubernetes.io/load-balancer-attributes: &assets-lb-attributes >
+          access_logs.s3.enabled=true,
+          access_logs.s3.bucket=govuk-integration-aws-logging,
```
martynipratt, line 61: integration bucket for staging env

sengi: Oof, good catch, thanks!

sengi: Fixed.

martynii (approved):
LGTM

martynipratt (approved):

PR #950, opened 2023-03-10
Remove Fastly dashboard.

sengi:
This code was copied from [mrnetops/fastly-dashboards](https://www.github.com/mrnetops/fastly-dashboards/blob/main/grafana/provisioning/dashboards/Fastly-Dashboard.json) in [#372](https://github.com/alphagov/govuk-helm-charts/pull/372/files#diff-a3b4ecb767e049b45ded337fc558544103cda7ee1c1bbed2f133b35cf4bbf077R8).

The source has no licence information, so we must presume that copyright subsists in the original work and therefore we must not include it (or a derivative of it) in our MIT-licensed repo without the author's permission.


martynipratt (approved):
LGTM

PR #951, opened 2023-03-10
Remove the fastly-exporter chart.

sengi:
We're no longer using it, and if/when we do want to use it again we would use https://www.github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-fastly-exporter rather than rolling our own.


martynipratt (approved):
LGTM

PR #952, opened 2023-03-11
Set PodDisruptionBudgets for interactive workloads.

sengi:
This should stop future node upgrades from tanking our serving availability.

I expect there'll be some room for optimisation once we have more experience running in production, but this should be fine to get us started. For the production settings I've leaned towards availability vs rollout progress/velocity for now.

We only want this for pods that are in the request path for interactive requests, not async workers or periodic/batch jobs etc.


nsabri1 (approved):
LGTM

PR #958, opened 2023-03-15
Create long-lived SA tokens for Argo Workflows.

sengi:
Before 1.22, Kubernetes used to create these automatically. They've been deprecated for ages but Argo Workflows [still relies on them](https://www.github.com/argoproj/argo-workflows/pull/9620/commits/8a69080cd58ea9f6a8ccc8c1372e5ee57487fbaf#diff-76961ec9c46eae52804982a6584c81e67658e9f9ae8eceb8525c9b9c085735b1). Not good.

We want to move away from Argo Workflows (i.e. just use Argo CD), but in the meantime, [add these long-lived serviceaccount tokens back in](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#manually-create-an-api-token-for-a-serviceaccount) so that Workflows continues to work when a cluster is installed from scratch.

### DIFF
--- a/charts/argo-bootstrap/Chart.yaml
+++ b/charts/argo-bootstrap/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v2
 name: argo-bootstrap
 description: Bootstraps ArgoCD with initial configuration
-version: 0.3.1
+version: 0.3.2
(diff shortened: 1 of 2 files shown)


nsabri1 (approved):
OK

PR #960, opened 2023-03-15
Use latest kubeconform for pre-merge checks.

sengi:
Use `latest` instead of specifying the version, because we don't have an automatic way to keep up to date and there's no security benefit either because image tags on GHCR are mutable.


nsabri1 (approved):
Sensible

PR #961, opened 2023-03-15
Refactor Router alert and add recording rules.

sengi:
- Add some recording rules for Router metrics and use these in the alert.
- Simplify the name of the alert.
- Add a minimum RPS threshold so it doesn't fire on statistically insignificant input. 
- Fix a failing alert test.
- Improve the messages on test failure and make the tests more self-documenting by adding some assertions on intermediate values.
- Clean up formatting.

Still to do in a later PR: actually run the tests pre-merge instead of them being manual 😅

### DIFF
--- a/charts/monitoring-config/rules/README.md
+++ b/charts/monitoring-config/rules/README.md
@@ -1,4 +1,9 @@
 ## Run Prometheus alert tests
 
 - Install `promtool` on your machine: `brew install prometheus`
-- Run `promtool test rules *_tests.yaml`.
+- Run the following command
+
+    ```sh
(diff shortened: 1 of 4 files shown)


sengi:
Hm looks like we're missing a kubeconform schema for PrometheusRule. Lemme fix that.

sengi:
Schema wasn't missing, but was borked somehow. Replaced it with the canonical one from [kube-prometheus](https://www.github.com/prometheus-operator/kube-prometheus) (`crdschemas/prometheusrule.json`, generated by running `make crdschemas`).

martynipratt (approved):
LGTM

PR #964, opened 2023-03-16
Use EKS backends for frontend apps in Staging and Prod

theseanything:
This removes Plek URI overrides to make frontend applications use backend services in the EC2 infrastructure. This left over from testing the frontend applications.

This also fixes an issue where draft stack is being configured to use the non draft backend services in EC2.


sengi (approved):
D'oh. Thanks! I should learn to grep my own TODOs before going on debugging adventures 😂

PR #965, opened 2023-03-16
1147 add missing cronjobs into staging and integration

martynipratt:
Adding cronjobs that are present in config/schedule.rb files in relevant app repositories into staging and integration.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2422,6 +2422,19 @@ govukApplications:
     uploadAssets:
       enabled: false
     workerEnabled: true
+    cronTasks:
+      - name: anonymous-feedback-deduplication-nightly
+        task: "anonymous_feedback_deduplication:nightly"
+        schedule: "10 0 * * *"
+      - name: import-organisations
+        task: "api_sync:import_organisations"
+        schedule: "20 0 * * *"
+      - name: service-feedback-aggregation
+        task: "service_feedback_aggregation:daily"
+        schedule: "30 0 * * *"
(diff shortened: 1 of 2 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2414,6 +2422,19 @@ govukApplications:
...
+        schedule: "10 0 * * *"
+      - name: import-organisations
+        task: "api_sync:import_organisations"
+        schedule: "20 0 * * *"
+      - name: service-feedback-aggregation
+        task: "service_feedback_aggregation:daily"
+        schedule: "30 0 * * *"
+      - name: anonymous-feedback-deduplication-recent
+        task: "anonymous_feedback_deduplication:recent"
+        schedule: "*/5 3 * * *"
```
sengi, line 2437: Just double checking this timespec as it seems a bit odd - is it supposed to be every 5 minutes from 03:00 through 03:55 and not run the rest of the time? (similarly for the staging one below)

sengi (approved):
Thanks for going through and finding these!

charts/app-config/values-integration.yaml
martynipratt: That was what I interpreted from this 
https://github.com/alphagov/support-api/blob/main/config/schedule.rb#L22

martynipratt: As per the comment the behaviour was actually to run on the 3rd minute every 5 minutes. Just going to make the cron run every 5 minutes. 
https://github.com/alphagov/support-api/commit/5e2151be4300c687d916ecf0dacbd972deafac09

PR #967, opened 2023-03-16
fixing workflows error as job names are too long

martynipratt:
Cronjob names must be no more than 52 characters. Appears to be a limitation in helm https://github.com/helm/monocular/issues/566


sengi (approved):
Pretty close to the 63-char limit on DNS name labels too 👍 

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names

PR #972, opened 2023-03-16
Adding logic to cron

nsabri1:
Adding statement to check if transition-config repo is already present, and if so run git pull if not then git clone.

### DIFF
--- a/charts/generic-govuk-app/templates/transition-config-cron.yaml
+++ b/charts/generic-govuk-app/templates/transition-config-cron.yaml
@@ -44,18 +44,16 @@ spec:
           initContainers:
             - name: clone-transition-config-repo
               image: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/github-cli:latest
-              command: ["git"]
-              args:
-                - clone
-                - --depth 1
-                - https://github.com/alphagov/transition-config 
-                - /transition-config/
+              command:
+                - sh
+                - -c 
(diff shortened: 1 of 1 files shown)


charts/generic-govuk-app/templates/transition-config-cron.yaml
```diff
@@ -44,18 +44,16 @@ spec:
...
-              command: ["git"]
-              args:
-                - clone
-                - --depth 1
-                - https://github.com/alphagov/transition-config 
-                - /transition-config/
+              command:
+                - sh
+                - -c 
+                - "[ -d .git ] && git pull || git clone --depth=1 https://github.com/alphagov/transition-config"
```
sengi, line 50: Should it be `transition-config/.git`?

sengi (approved):

charts/generic-govuk-app/templates/transition-config-cron.yaml
```diff
@@ -44,18 +44,16 @@ spec:
...
               securityContext:
                 allowPrivilegeEscalation: false
                 readOnlyRootFilesystem: true
               volumeMounts:
                 - name: transition-config-efs
-                  mountPath: /transition-config
+                  mountPath: /home/user
```
sengi, line 56: is this intentional?

nsabri1: yes

PR #981, opened 2023-03-17
Prevent govuk-replatform-test-app deploys to staging

theseanything:
This application is not present in our staging environment. This file will prevent the deployment system from trying to automatically deploy the app into the Staging environment, which just results in an error.


sengi (approved):
😅 thanks for fixing this!

PR #982, opened 2023-03-17
[RP-1135] Temp Disable Publisher Worker on Staging

dj-maisy:
## What?
This aims to disable the publisher worker in Staging temporarily, so we can load test the worker on legacy EC2 infra without the EKS worker interfering in the results.


sengi (approved):

sengi:
Don't forget to revert it 😅

PR #986, opened 2023-03-17
Merge staging Helm values into production.

sengi:
This adds the publisher apps and APIs in the production cluster.

Essentially a copy of the staging config with `s/staging/production/` and some other production-specific values changed. See commits for a bit more detail.

Worker deployments are disabled for now, since we don't want workers in k8s pulling work off queues over the weekend.

See #936 for a list of things to consider in review.

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -274,6 +274,7 @@ govukApplications:
 
 - name: collections
   helmValues:
+    replicaCount: 4
     appResources:
       limits:
         cpu: 2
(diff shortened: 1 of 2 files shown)


theseanything (approved):

charts/app-config/values-production.yaml
```diff
@@ -2,16 +2,16 @@
 # individual apps below.
 
 govukEnvironment: production
-externalDomainSuffix: gov.uk
-publishingDomainSuffix: publishing.service.gov.uk
-assetsDomain: assets.publishing.service.gov.uk
+externalDomainSuffix: production.publishing.service.gov.uk
+publishingDomainSuffix: production.publishing.service.gov.uk
+assetsDomain: assets.production.publishing.service.gov.uk
```
theseanything, line 7: Whilst I think these are valid domains - should we be using just `publishing.service.gov.uk`, without "production" subdomain?

sengi: Oof, good catch, thanks! I'd been lulled into a false sense of security from doing the integration->staging one earlier.

sengi: Think I've found and fixed 'em all. PTAL. Thanks!

PR #991, opened 2023-03-20
Fix Smokey Plek configuration for domain suffixes.

sengi:
This makes Smokey's Plek configuration consistent with the applications being probed (apart from the workaround for the content-data-admin vs. content-data alias, which we still need).

Tested on the (not-yet-live) prod cluster.

### DIFF
--- a/charts/smokey/templates/workflow-templates/smoke-test.yaml
+++ b/charts/smokey/templates/workflow-templates/smoke-test.yaml
@@ -31,15 +31,17 @@ spec:
           - name: GOVUK_APP_DOMAIN
             value: ""
           - name: GOVUK_APP_DOMAIN_EXTERNAL
-            value: eks.{{ .Values.govukEnvironment }}.govuk.digital
+            value: {{ .Values.k8sPublishingDomainSuffix }}
           - name: GOVUK_ASSET_ROOT
-            value: https://assets-eks.{{ .Values.govukEnvironment }}.publishing.service.gov.uk
+            value: https://{{ .Values.k8sAssetsDomain }}
           - name: GOVUK_WEBSITE_ROOT
-            value: https://www.eks.{{ .Values.govukEnvironment }}.govuk.digital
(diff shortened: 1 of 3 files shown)


charts/smokey/templates/workflow-templates/smoke-test.yaml
```diff
@@ -31,15 +31,17 @@ spec:
...
-            value: https://www.eks.{{ .Values.govukEnvironment }}.govuk.digital
+            value: https://www.{{ .Values.k8sExternalDomainSuffix }}
           - name: PLEK_SERVICE_ASSETS_URI
-            value: https://assets-eks.{{ .Values.govukEnvironment }}.publishing.service.gov.uk
+            value: https://{{ .Values.k8sAssetsDomain }}
           - name: PLEK_SERVICE_ASSETS_ORIGIN_URI
-            value: https://assets-origin.eks.{{ .Values.govukEnvironment }}.govuk.digital
+            value: https://assets-origin.{{ .Values.k8sExternalDomainSuffix }}
+          - name: PLEK_SERVICE_CONTENT_DATA_ADMIN_URI
+            value: https://content-data.{{ .Values.k8sExternalDomainSuffix }}
```
theseanything, line 44: Nothing about this PR specifically, not sure I've understood why this override for content data is needed. Is it because the URL is content-data rather than content-data-admin?

theseanything (approved):
Good catch!

theseanything:
Looks like you might need to add dummy helm values for the smokey chart for kubeconform.

charts/smokey/templates/workflow-templates/smoke-test.yaml
sengi: Yeah that's exactly it. In the old setup it redirects, but we don't have that set up at the moment so we need Smokey to use the canonical URL (which is ok because that's what real users would be using).

PR #996, opened 2023-03-20
Remove IP source address filter on prod Signon.

sengi:
When we launch the publisher apps on Kubernetes, other gov departments etc. will need to be able to make requests to Signon via its ingress.

Ideally we don't want to be serving Signon on the govuk.digital domain, but we can solve this after launch. Probably the way to solve it is to put Signon behind Fastly and re-apply the origin ACL to the ingress (i.e. revert this PR once Signon is behind Fastly). That'd also give us DoS protection.


nsabri1 (approved):
ok

PR #1001, opened 2023-03-20
Enable workers on some API apps.

sengi:
Some frontend features rely on Sidekiq jobs in the mid-tier APIs, like the feedback form that goes via support-api to the Zendesk API.


theseanything (commented):
Looks good - thanks for catching this @sengi!

PR #1004, opened 2023-03-21
Add preStop hook wait for ALB to drain connections

theseanything:
We have a race condition during a rollout between a pod being terminated and the ALB deregistering it. Currently these process happen independently and frequently the pods terminate before ALB deregistration which cause traffic to be sent to a terminating pod. This causes a few 502/504s to be served until deregistration is complete. Until we have a better fix, adding a delay until pod termination makes this senario less likely. However, it would be ideal if we had proper fix whereby pods are only terminated after it's known that they've be successfully deregistered from the ALB.

### DIFF
--- a/charts/generic-govuk-app/templates/deployment.yaml
+++ b/charts/generic-govuk-app/templates/deployment.yaml
@@ -126,6 +126,10 @@ spec:
             {{- with .Values.appExtraVolumeMounts }}
               {{- . | toYaml | trim | nindent 12 }}
             {{- end }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["sleep", "15"] # To allow time for ALB to deregister pod before termination
         - name: nginx
           image: "{{ .Values.nginxImage.repository }}:{{ .Values.nginxImage.tag }}"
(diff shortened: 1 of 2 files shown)


charts/generic-govuk-app/templates/deployment.yaml
```diff
@@ -126,6 +126,10 @@ spec:
             {{- with .Values.appExtraVolumeMounts }}
               {{- . | toYaml | trim | nindent 12 }}
             {{- end }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh","-c","sleep 15"]
```
sengi, line 132: simplifies to
```suggestion
                command: ["sleep", "15"]
```
(similarly below)

sengi (approved):
LGTM. Shame we need it but reasonable workaround for now.

Do we wanna do the same thing for the asset-manager chart?

theseanything:
> LGTM. Shame we need it but reasonable workaround for now.
Yah, sadly I known issue: https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2366 
 
> Do we wanna do the same thing for the asset-manager chart?
Yup, added!

PR #1006, opened 2023-03-21
Fix smokey domains for Integration

theseanything:
As we've switched over domains in Integration, smokey needs to use them instead of the internal domains as Signon will not authenticate causing test failures.


sengi (approved):

sengi:
Thanks!

PR #1007, opened 2023-03-21
Staging cutover: reference live domains for Smokey and Signon

theseanything:
- Remove the `s/staging.publishing.service.gov.uk/eks.staging.govuk.digital/` substitution.
- Remove the Plek override.
- Points smokey at non-internal domains, as Signon will fail otherwise.


sengi (approved):

martynipratt (approved):
LGTM

PR #1013, opened 2023-03-22
Create separate Whitehall db secrets

theseanything:
This separates the Whitehall frontend and admin database secrets, as frontend only has read-only permissions.


nsabri1 (approved):
ok

PR #1014, opened 2023-03-22
Set the max body size for Content Store request

theseanything:
Requests coming from the Publishing API can be larger than the default, this increase the max body size to prevent those request being rejected. Fixes an issue publishing organisation pages in Whitehall.


sengi (approved):
Good find!

PR #1026, opened 2023-03-23
Automatic deploys missing for some apps 

theseanything:
These applications had automatic deployments enabled in EC2 and should also have them enabled in EKS.


martynipratt (approved):
LGTM

PR #1027, opened 2023-03-23
Add ingress for email-alerts-api-public

samsimpson1:


sengi (approved):
Didn't realise there was already some nginx cruft that limits the paths that can be proxied through to the app. LGTM 👍

sengi:
Ah right that's just rate limiting and not rejecting requests - well spotted!

PR #1034, opened 2023-03-24
Corrected path for mount location of transition-config.

nsabri1:
It should look like:
/app/data/transition-config

Its currently looking like this:
/app/data/transition-config/transition-config


sengi (approved):
Good catch, thanks for fixing!

PR #1041, opened 2023-03-27
Set max upload size for Imminence.

sengi:
Looks like we missed this. Users definitely need to be able to upload large (at least > 5 MiB) objects to Imminence.

https://trello.com/c/vTe1fLR2
https://govuk.zendesk.com/agent/tickets/5271205


KludgeKML (approved):
Looks good to me.

PR #1043, opened 2023-03-29
Prevent post sync workflow from trigging deployment promotion

theseanything:
This adds a check to prevent the post sync workflow from trigging the deploy image workflow in the promotion environment. Previously this was normal behaviour as the deploy image workflow had a check prevent deployments that shouldn't be promoted. However, this mean that other steps (including notifying the release app of deployment) in the workflow continued to execute.

We now explicitly state whether a deployment should be promoted in the image tag file. This states is used by the post sync workflow to decide whether to trigger a promotion.

This simplifies the mental model, and remove the confusion around deploy image workflow always running even when a deployment isn't being promoted.

This check relies on workflow step outputs, which required the correct service account permissions - that had not been previously set up.

We continue to have the manual_deploy flag set - but will remove in future PRs, as should be no longer needed.

(Tested in integration)

### DIFF
--- a/charts/argo-services/templates/executor-default-rolebinding.yaml
+++ b/charts/argo-services/templates/executor-default-rolebinding.yaml
@@ -1,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: executor-default
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: executor
+subjects:
+  - kind: ServiceAccount
+    name: default
--- a/charts/argo-services/scripts/update-image-tag.sh
+++ b/charts/argo-services/scripts/update-image-tag.sh
@@ -10,6 +10,7 @@ change_image_tag() {
   git checkout -b "${BRANCH}"
 
   yq -i '.image_tag = env(IMAGE_TAG)' "${FILE}"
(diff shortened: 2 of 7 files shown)


charts/argo-services/templates/executor-default-rolebinding.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: RoleBinding
+metadata:
+  name: executor-default
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: executor
+subjects:
+  - kind: ServiceAccount
+    name: default
```
sengi, line 11: I guess it's ok to bind this to the default SA (maybe not 100% ideal since there's stuff other than Argo in this namespace but create/patch workflowtaskresults doesn't seem like a big deal so 👍)

sengi (approved):
Nice improvement!

charts/argo-services/templates/executor-default-rolebinding.yaml
theseanything: Yeah - probably not ideal to be using the default SA for workflows anyways. But didn't wanna tackle that now, particular if Argo Workflows might go anyways

sengi: 💯

PR #1048, opened 2023-03-31
Enable ingress for content-data-api.

sengi:
Missed this one! It's still serving from EC2 right now.

See https://github.com/alphagov/content-data-api/#content-data-api for example URLs that show that this API is indeed meant to be reachable from the Internet (with Signon auth). e.g. https://content-data-api.publishing.service.gov.uk/api/v1/metrics

Tested: applied the Ingress generated from the template in the integration cluster, verified that Signon works and URL above serves the expected response (with /etc/hosts patched to point at the Ingress ALB).

```
helm template content-data-api ../generic-govuk-app --values \
  <(helm template . --values values-integration.yaml |
    yq e '.|select(.metadata.name=="content-data-api").spec.source.helm.values'
   ) \
  --set sentry.enabled=false --set rails.createKeyBaseSecret=false \
  | yq e '.|select(.kind=="Ingress")' \
  | k apply -f -
```

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -500,10 +500,21 @@ govukApplications:
 
 - name: content-data-api
   helmValues:
+    ingress:
+      enabled: true
+      annotations:
+        <<: *default-alb-ingress-annotations
+        <<: *alb-ingress-backend-waf-ruleset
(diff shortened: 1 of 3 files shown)


nsabri1 (approved):
Nice

PR #1049, opened 2023-04-03
Add s3 bucket name for whitehall

BeckaL:
These variables were previously defined ([1](https://github.com/alphagov/govuk-puppet/blob/9c3dec9f7b01fbbae6aaf207fc8acd35eaa2cce8/hieradata_aws/production.yaml\#L252), [2](https://github.com/alphagov/govuk-puppet/blob/9c3dec9f7b01fbbae6aaf207fc8acd35eaa2cce8/hieradata_aws/staging.yaml\#L243), [3](https://github.com/alphagov/govuk-puppet/blob/9c3dec9f7b01fbbae6aaf207fc8acd35eaa2cce8/hieradata_aws/integration.yaml\#L109)) in govuk-puppet, and is needed for us to enable csv exports of whitehall data, which publishers use for a varierty of things, including accessibility reports and parliamentary questions. 

We've had several zendesk reports of this issue.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2754,6 +2754,8 @@ govukApplications:
           secretKeyRef:
             name: signon-token-whitehall-search-api
             key: bearer_token
+      - name: AWS_S3_BUCKET_NAME
+        value: govuk-integration-whitehall-csvs
       - name: GOVUK_NOTIFY_TEMPLATE_ID
         value: *publishing-notify-template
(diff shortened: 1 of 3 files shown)


marcpomfret (approved):
Ok, thanks

sengi (approved):
LGTM, looks like the nodes have some IAM [permissions](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf) on these buckets.

PR #1053, opened 2023-04-06
Enable 2sv mandating features in staging

BeckaL:
This forms part of a 'dry run' of 2sv mandating in staging. We've [set them in integration](https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-integration.yaml#L2139-L2142), so this change is equivalent for staging.

Once in place, they will enable features that allow admins to exempt individuals from 2 step-verification (2sv), and mandate 2sv at an organisation level.

[Trello](https://trello.com/c/bnFeNsvd/532-dry-run-enable-all-new-2fa-features-on-integration-staging-before-prod-rollout-which-will-happen-in-week-commencing-10-04-2023)


marcpomfret (approved):
ok

PR #1057, opened 2023-04-11
Simplify promotion checks

theseanything:
This moves the checks for whether an application deployment should be promoted into the post-sync workflow. Previously checks were done in the update-image-tag script, which was confusing as it looked like a deployment, however the workflow was just checking whether a deployment should occur. This improves the separation of concerns between the two workflows.

This also fixes a bug whereby we'd notify the release application of deployment, even though the update-image-tag didn't actually make a change (because the deployment wasn't promoted).

(Tested in integration)

### DIFF
--- a/charts/argo-services/scripts/check-for-promotion.sh
+++ b/charts/argo-services/scripts/check-for-promotion.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Fetch the YAML configuration
+CONFIG_URL="https://raw.githubusercontent.com/alphagov/govuk-helm-charts/main/charts/app-config/image-tags/${ENVIRONMENT}/${REPO_NAME}"
+CONFIG_CONTENT=$(curl -Ls "${CONFIG_URL}")
+
+# Extract the values of automatic_deploys_enabled and promote_deployment
+automatic_deploys_enabled=$(echo "${CONFIG_CONTENT}" | yq '.automatic_deploys_enabled' -)
+promote_deployment=$(echo "${CONFIG_CONTENT}" | yq '.promote_deployment' -)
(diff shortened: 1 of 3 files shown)


charts/argo-services/scripts/check-for-promotion.sh
```diff
@@ -0,0 +1,17 @@
...
+# Fetch the YAML configuration
+CONFIG_URL="https://raw.githubusercontent.com/alphagov/govuk-helm-charts/main/charts/app-config/image-tags/${ENVIRONMENT}/${REPO_NAME}"
+CONFIG_CONTENT=$(curl -Ls "${CONFIG_URL}")
+
+# Extract the values of automatic_deploys_enabled and promote_deployment
+automatic_deploys_enabled=$(echo "${CONFIG_CONTENT}" | yq '.automatic_deploys_enabled' -)
+promote_deployment=$(echo "${CONFIG_CONTENT}" | yq '.promote_deployment' -)
+
+# Check if both values are true
+if [ "${automatic_deploys_enabled,,}" == "true" ] && [ "${promote_deployment,,}" == "true" ]; then
```
sengi, line 13: TIL bash4 param expansions 😅 https://wiki.bash-hackers.org/syntax/pe#case_modification 👍

sengi (approved):
Nice!

charts/argo-services/scripts/check-for-promotion.sh
sengi, line 13: Oh these should really be single `=` rather than double (though bash lets you get away with it iirc) e.g.
```suggestion
if [[ "${automatic_deploys_enabled,,}" = "true" -a "${promote_deployment,,}" = "true" ]]; then
```

PR #1060, opened 2023-04-12
Fix the default for automatic_deploys_enabled.

sengi:
automatic_deploys_enabled used to default to true but this subtly changed to false in #1057, which broke automatic promotion (push-on-green) from staging to production for apps which didn't have automatic_deploys_enabled explicitly set to true in their image-tags YAML files.


theseanything (approved):
Approving as this was our existing behaviour anyways.

PR #1061, opened 2023-04-12
Set Sentry environment names back to normal

nsabri1:
removal of the eks suffix from env

https://trello.com/c/65hnxqmM/1169-set-sentry-environment-names-back-to-normal-lose-the-eks-suffix-so-that-teams-sentry-alert-rules-work-again


sengi (approved):
Pleasingly straightforward! :D

nsabri1:
> Pleasingly straightforward! :D

it took more time working out if i needed to do something to sentry as this seems to straightforward !!!

PR #1062, opened 2023-04-12
Explicitly set automatic deploy enabled values

theseanything:
This reverts the fix in https://github.com/alphagov/govuk-helm-charts/pull/1060 in favour of explicitly setting the `automatic_deploy_enabled` values in the image tag file.

This should minimise implicit behaviour and be less risky for apps that shouldn't be automatically deployed.

### DIFF
--- a/charts/app-config/image-tags/staging/account-api
+++ b/charts/app-config/image-tags/staging/account-api
@@ -1,1 +1,2 @@
 image_tag: release-84acfbec78e9a637e287609fd1c0c55d32176ea6
+automatic_deploys_enabled: true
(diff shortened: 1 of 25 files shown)


sengi (approved):
Some of these apps weren't previously push-on-green, like signon for instance. We probably wanna leave those off for now?

sengi:
Actually maybe Signon's the only one here where that's the case.

sengi:
Oh no, scratch that, sorry, looks like Signon is push-on-green these days.

theseanything:
> Some of these apps weren't previously push-on-green, like signon for instance. We probably wanna leave those off for now?

No good catch still tho - Signon should be push on green, but there were a few other apps that shouldn't have been e.g. service-manuals-frontend. I've gone back over them and made them match what's marked as "manually deployed" in the release app.

PR #1068, opened 2023-04-13
Add content-data-api ETL cronjob.

sengi:
This used to be run [from Jenkins](https://github.com/alphagov/govuk-puppet/blob/3327d34/modules/govuk_jenkins/templates/jobs/content_data_api.yaml.erb).

The variations in the minutes field are intentional, for load spreading (since k8s CronJob still doesn't [support fuzzing the start time](https://www.github.com/kubernetes/kubernetes/issues/91652) 😢).

Tested: ran successfully on the integration cluster.

https://trello.com/c/v9XYF5EB

Related: https://github.com/alphagov/content-data-api/pull/1867

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -535,6 +535,10 @@ govukApplications:
         name: publishing-api-consumer
       - command: ['rake', 'publishing_api:bulk_import_consumer']
         name: bulk-import-publishing-api-consumer
+    cronTasks:
+      - name: etl
+        task: "etl:main"
+        schedule: "22 13 * * *"
     extraEnv:
       - name: GOOGLE_ANALYTICS_GOVUK_VIEW_ID
         valueFrom:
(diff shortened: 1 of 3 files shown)


sengi:
Oh, thanks I hadn't spotted that it was running at completely different times in non-prod. I'll fix that before merging.

deborahchua (approved):

sengi:
I only just spotted that the "re-run" job also runs nightly on a schedule. Is that definitely still needed?

If so, I think you'd need to fix up the Rake task so that it can run without having to compute the dates to pass into it :)

sengi:
I'm fairly sure the nightly re-run job became unnecessary with GA4: https://support.google.com/analytics/answer/11198161?hl=en

Shall we see how it goes without it?

deborahchua:
Yeh it sounds like we may not need it anymore! Happy to leave it out and see how it goes 👍

PR #1070, opened 2023-04-14
Remove "eks" prefix from Sentry environments for Router

theseanything:
Router uses a different ENV VAR to set Sentry environment names - these were missed with when dropped the prefix in 1a6c38070e05087099a477ac171939e3cd505b92 - https://github.com/alphagov/govuk-helm-charts/pull/1061


nsabri1 (approved):
Ok

PR #1075, opened 2023-04-17
Set env var for bucket name for publisher reports

theseanything:
This env var is used by Publisher to set the bucket name which it uses to upload reports.

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1592,6 +1592,8 @@ govukApplications:
         value: e739d23a-b761-44af-9a99-a1eba0b75c7e
       - name: GOVUK_NOTIFY_TEMPLATE_ID
         value: *publishing-notify-template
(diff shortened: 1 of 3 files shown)


charts/app-config/values-production.yaml
```diff
@@ -1592,6 +1592,8 @@ govukApplications:
         value: e739d23a-b761-44af-9a99-a1eba0b75c7e
       - name: GOVUK_NOTIFY_TEMPLATE_ID
         value: *publishing-notify-template
+      - name: REPORTS_S3_BUCKET_NAME
+        value: govuk-integration-publisher-csvs
```
sengi, line 1596: copy-pasto :) (similarly for staging)

sengi (approved):

charts/app-config/values-production.yaml
theseanything: ahah thank you!!

PR #1080, opened 2023-04-18
Remove no-op PDBs for non-latency-sensitive apps.

sengi:
Neither cache-clearing-service nor email-alert-service has a web server component; they just pick up work from a shared queue. There's no problem if there are no running replicas of them momentarily, because the work isn't latency-sensitive.

The existing PDBs created by the template weren't matching anything, so we're just removing them to reduce the potential for confusion at seeing a PDB which isn't matching anything.

Tested: chart validates and installs ok with [`helm install`](https://github.com/alphagov/govuk-helm-charts#installing-an-application-chart-without-argo-cd).

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -282,6 +282,7 @@ govukApplications:
     rails:
       enabled: false
     appEnabled: false
+    podDisruptionBudget: {}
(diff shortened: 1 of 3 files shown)


marcpomfret (approved):
Ok

PR #1085, opened 2023-04-18
Remove old replicaset clutter from Argo WF sensors.

sengi:
The default of 10 really clutters up the Argo CD UI.

We don't really need to keep any previous replicasets for these, but let's make it 1 rather than 0 just on the off chance that it comes in handy when debugging one day.

[Schema reference for the field we're setting](https://www.github.com/argoproj/argo-events/blob/c2e63ea/api/jsonschema/schema.json#L3850)


dj-maisy (approved):
👍

PR #1087, opened 2023-04-19
Improve names of Signon cronjobs + dither timings.

sengi:
- Make it clearer that the "sync" jobs are syncing secrets from Signon into k8s.
- The CronJob names no longer match the Rake task names as closely, but that's fine; the contexts are different so e.g. the `kubernetes-` prefix makes sense in the Rake task but not so much in the name of the k8s cronjob.
- Also dither the timings a bit to smooth out the resource usage (i.e. reduce the chance of simultaneous cronjobs causing cluster scale-up).


nsabri1 (approved):
ok

PR #1091, opened 2023-04-19
Updated the command

nsabri1:
This ensure the git pull and cloning are done correctly.

This has been tested, and working as expected.

### DIFF
--- a/charts/generic-govuk-app/templates/transition-config-cron.yaml
+++ b/charts/generic-govuk-app/templates/transition-config-cron.yaml
@@ -47,7 +47,10 @@ spec:
               command:
                 - sh
                 - -c
-                - "[ -d .git ] && git pull || git clone --depth=1 https://github.com/alphagov/transition-config"
+                - "[ -d transition-config ] && \
+                   cd transition-config && \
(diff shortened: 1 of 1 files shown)


charts/generic-govuk-app/templates/transition-config-cron.yaml
```diff
@@ -47,7 +47,9 @@ spec:
               command:
                 - sh
                 - -c
-                - "[ -d .git ] && git pull || git clone --depth=1 https://github.com/alphagov/transition-config"
+                - "[ -d transition-config ] && cd transition-config && \
+                   [ -d .git ] && git pull || \
+                   git clone --depth=1 https://github.com/alphagov/transition-config transition-config"
```
sengi, line 52: optional: could reformat this a bit for readability, maybe something like
```suggestion
                - >
                  [ -d transition-config ] &&
                    cd transition-config &&
                    [ -d .git ] &&
                    git pull ||
                  git clone --depth=1 https://github.com/alphagov/transition-config transition-config
```

sengi (approved):

PR #1108, opened 2023-05-05
Bump max client body size for asset manager and travel advice publisher

BeckaL:
We've had a [zendesk ticket](https://govuk.zendesk.com/agent/tickets/5315094) from someone trying to upload a 2.1MB PDF via travel advice publisher, so bumping these limits to allow for this. We'll need both a bump to travel advice and to asset manager since this is an asset.

Questions I have:
* Is arbitrary 3MB OK? Any reason to pick any other number?
* I'm somewhat surprised asset manager didn't already have a higher limit than the default, although maybe mostly what we've been contending with has been large content items rather than large assets. But wanted to double check that there's definitely nowhere else this could be set for asset manager that I'm not overriding

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -123,6 +123,7 @@ govukApplications:
 - name: asset-manager
   chartPath: charts/asset-manager
   helmValues:
+    nginxClientMaxBodySize: 3M
     workerEnabled: true
     ingress:
       enabled: true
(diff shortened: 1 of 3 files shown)


marcpomfret (approved):
LGTM

marcpomfret:
3MB sounds fine
the default is set at 1MB in charts/generic-govuk-app/values.yaml so the only place that it gets overridden from is in the values files that you've changed

PR #1109, opened 2023-05-10
Enable Rails database migrations for imminence.

sengi:
We weren't running Rails database migrations (aka schema changes) on rollout for this app. Looks like we just forgot to set the flag in the Helm values. Probably a copy-pasto.

Thanks @kludgekml for spotting this!


KludgeKML (approved):
LGTM

PR #1110, opened 2023-05-10
Fix config for upload size limits on publisher apps.

sengi:
- Remove a [no-op Helm value](https://github.com/alphagov/govuk-helm-charts/pull/1108) for asset-manager.
- Apply the 500 MB upload size limit to travel-advice-publisher (same as the others).

Bit more detail in the commit messages.


BeckaL (approved):
Eek, sorry about that! Thanks for commit explanations

sengi:
No worries at all!

PR #1122, opened 2023-05-23
Actually set worker replicas to 1 for Release.

sengi:
Whoops, sorry for the extra noise. Fixes typo in 682df2f7.


theseanything (approved):
Ah shoot noticed that - but thought the intention was to just scale down staging. Should have asked sorry!

PR #1123, opened 2023-05-24
Let Release app access the Slack webhook url

AgaDufrat:
This adds SLACK_WEBHOOK_URL as extra variable to the Release app config.

The Slack webhook url is stored in AWS Secrets Manager and already used by Argo:
https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=govuk%2Fslack-webhook-url&region=eu-west-1


sengi (approved):
👍

PR #1126, opened 2023-05-24
Add Asset Manager API key to Frontend

callumknights:
Done as part of the work to migrate CSV Previews from Whitehall to Frontend. 

Trello: https://trello.com/c/7IA3mJD8/544-set-up-communication-between-frontend-and-asset-manager-for-csv-previews

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1016,6 +1016,11 @@ govukApplications:
           secretKeyRef:
             name: signon-token-frontend-account-api
             key: bearer_token
+      - name: ASSET_MANAGER_BEARER_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: signon-token-frontend-asset-manager
+            key: bearer-token
       - name: ELECTIONS_API_KEY
         valueFrom:
           secretKeyRef:
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1043,6 +1043,11 @@ govukApplications:
           secretKeyRef:
             name: signon-token-frontend-account-api
             key: bearer_token
+      - name: ASSET_MANAGER_BEARER_TOKEN
+        valueFrom:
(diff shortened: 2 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1016,6 +1016,11 @@ govukApplications:
           secretKeyRef:
             name: signon-token-frontend-account-api
             key: bearer_token
+      - name: ASSET_MANAGER_API_KEY
+        valueFrom:
+          secretKeyRef:
+            name: frontend-asset-manager
+            key: ASSET_MANAGER_API_KEY
```
sengi, line 1023: This API key is [from Signon](https://signon.integration.publishing.service.gov.uk/api_users/423/edit), right?

If so, you don't need ExternalSecret or Secrets Manager, because Signon creates those secrets automatically.

Signon creates secrets for you named `signon-token-<client app>-<server app>` containing a key named `bearer_token` whose value is the API key. For an example, have a look at the `secretKeyRef` for the Account API one that's directly above on line 1017.

You just need to change `frontend-asset-manager` to `signon-token-frontend-asset-manager` and `ASSET_MANAGER_API_KEY` to `bearer_token`. (You might want to consider calling your environment variable `ASSET_MANAGER_BEARER_TOKEN` for the sake of following the precedent of all the other ones.)

So you'd end up with something like this:

```suggestion
      - name: ASSET_MANAGER_BEARER_TOKEN
        valueFrom:
          secretKeyRef:
            name: signon-token-frontend-asset-manager
            key: bearer_token
```

charts/external-secrets/templates/frontend/asset-manager.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: external-secrets.io/v1beta1
```
sengi, line 1: You don't need this whole file.

sengi (approved):
LGTM

PR #1127, opened 2023-05-24
Add email alert monitoring

nsabri1:
Currently runs on jenkins and is being migrated over to EKS - this only needs to be run in Production.
https://deploy.blue.production.govuk.digital/job/medical-safety-email-alert-check/configure

Jenkins job currently runs to Rake commands on a cron:
bundle exec rake run_medical_alerts @ 21:51 daily.
bundle exec rake run_travel_alerts @ every 15th minute.

Purposely not added to Integration and Staging, though i have added Image tags in all environments incase we may want to deploy in other environments if testing.

https://trello.com/c/qBfopr21/1184-medicalsafetyemailalertcheck-traveladviceemailalertcheck-jenkins-job-move-to-eks

### DIFF
--- a/charts/app-config/image-tags/integration/email-alert-monitoring
+++ b/charts/app-config/image-tags/integration/email-alert-monitoring
@@ -1,3 +1,1 @@
-image_tag: release-2532625527a6ba455e543a56cefcd82d814f9cd8
 automatic_deploys_enabled: true
-promote_deployment: true
--- a/charts/app-config/image-tags/integration/email-alert-monitoring
+++ b/charts/app-config/image-tags/integration/email-alert-monitoring
@@ -1,0 +1,3 @@
+image_tag: release-2532625527a6ba455e543a56cefcd82d814f9cd8
+automatic_deploys_enabled: true
+promote_deployment: true
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -917,6 +917,38 @@ govukApplications:
             name: email-alert-auth
             key: token
 
+- name: email-alert-monitoring
+  helmValues:
+    appEnabled: false
+    uploadAssets:
+      enabled: false
+    rails:
+      enabled: false
+    cronTasks:
+      - name: run_medical_alerts
+        task: "run_medical_alerts"
+        schedule: "21,51 * * * *"
+      - name: run_travel_alerts
+        task: "run_travel_alerts"
+        schedule: "*/15 * * * *"
+    workerEnabled: false
+    extraEnv:
+      - name: GOOGLE_OAUTH_CREDENTIALS
+        valueFrom:
+          secretKeyRef:
+            name: email-alert-monitoring-google
+            key: oauth_credentials
+      - name: GOOGLE_CLIENT_ID
+        valueFrom:
+          secretKeyRef:
+            name: email-alert-monitoring-google
+            key: client_id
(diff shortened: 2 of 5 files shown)


charts/app-config/image-tags/integration/email-alert-monitoring
theseanything, line 1: I think you'll want to add:
```
automatic_deploys_enabled: true
```
to this file and the other environment image-tag files.

charts/app-config/values-production.yaml
```diff
@@ -917,6 +917,38 @@ govukApplications:
...
+  helmValues:
+    appEnabled: false
+    uploadAssets:
+      enabled: false
+    rails:
+      enabled: false
+    cronTasks:
+      - name: run_medical_alerts
+        task: "run_medical_alerts"
+        schedule: "21 51 * * *"
```
theseanything, line 930: This cron is invalid, think its meant to be:
```suggestion
        schedule: "21,51 * * * *"
```

theseanything (approved):

charts/external-secrets/templates/email-alert-monitoring/google.yaml
```diff
@@ -0,0 +1,20 @@
...
+  refreshInterval: {{ .Values.externalSecrets.refreshInterval }}
+  secretStoreRef:
+    name: aws-secretsmanager
+    kind: ClusterSecretStore
+  target:
+    deletionPolicy: {{ .Values.externalSecrets.deletionPolicy }}
+    name: email-alert-monitoring-google
+  dataFrom:
+    - extract:
+        key: govuk/email-alert-montioring/google
```
sengi, line 20: typo
```suggestion
        key: govuk/email-alert-monitoring/google
```

sengi (approved):
This just runs the probes but nothing's actually alerting on the results yet, right? (Just double-checking my understanding here.)

charts/app-config/values-production.yaml
nsabri1: Good spot @theseanything

charts/external-secrets/templates/email-alert-monitoring/google.yaml
nsabri1: Opss - thanks @sengi

nsabri1:
> This just runs the probes but nothing's actually alerting on the results yet, right? (Just double-checking my understanding here. I think it's a great way to approach it btw.)
That is correct - perhaps i can have a think on ths and have a chat with @KludgeKML on this.

sengi:
> > This just runs the probes but nothing's actually alerting on the results yet, right? (Just double-checking my understanding here. I think it's a great way to approach it btw.)
> > That is correct - perhaps i can have a think on ths and have a chat with @KludgeKML on this.

Cool. When you're ready to tackle that part, you'll just want to have the job push a metric to pushgateway (we have one now) so that you can have define a Prometheus alert on it.

nsabri1:
> > > This just runs the probes but nothing's actually alerting on the results yet, right? (Just double-checking my understanding here. I think it's a great way to approach it btw.)
> > > That is correct - perhaps i can have a think on ths and have a chat with @KludgeKML on this.
> 
> Cool. When you're ready to tackle that part, you'll just want to have the job push a metric to pushgateway (we have one now) so that you can define a Prometheus alert on it.

Thats good to know - ok i will look into this

PR #1129, opened 2023-05-26
Add SSO_PUSH_USER_EMAIL env var to signon

theseanything:
This is used to sync new user permissions with applications. This env var was previously set in the old infrastructure, but not copied over.


sengi:
Thanks! Good find!

sengi:
The app should probably just default to this. I don't think it even needs to be configurable(?)

PR #1131, opened 2023-05-30
Fix broken reference to a secret in Finder app

ollietreend:
The finder app was failing to start because it couldn't find the referenced key `bearer-token` in the secret store.

It turns out this was a typo in PR #1126 / commit 94f254ca5f527b6e9f7505ed5199c5885cf7909a and the correct key name should be `bearer_token`.


BeckaL (approved):
LGTM

PR #1135, opened 2023-05-30
Add Search API learn to rank CronJob

theseanything:
This replaces the CronJob in Jenkins. Enabling only for integration for testing purposes. This job occurs in two parts; firstly a rake task needs to be called within the search-api container to generate datasets, then a python container is used to train and deploy the Sagemaker model in AWS. Unfortunately AWS only has made the Sagemaker SDK available in Python, so forces to maintain two separate runtimes (Ruby and Python).

### DIFF
--- a/charts/app-config/image-tags/integration/search-api-learn-to-rank
+++ b/charts/app-config/image-tags/integration/search-api-learn-to-rank
@@ -1,0 +1,3 @@
+image_tag: latest
+automatic_deploys_enabled: true
+promote_deployment: true


charts/govuk-jobs/templates/search-api-ltr-cronjob.yaml
sengi, line 1: optional / for later: if the idea is to have somewhere for batch jobs that don't fit well in the generic-govuk-app chart, we could probably improve on the chart name a bit. (Maybe `batch-jobs` or something like that? since the `govuk` part is implied by it being in this repo at all.) I don't wanna hold you up with bikeshedding over naming though, so honestly feel free to merge as-is and we can easily rename in the near future.

sengi (approved):

sengi:
Alternative idea: what if we had the `learn_to_rank:generate_data` Rake task exec the Python thing itself when it's done? Would that avoid the whole need for a separate chart without introducing too much ugliness?

charts/govuk-jobs/templates/search-api-ltr-cronjob.yaml
theseanything: yup, completely agree on the naming. We added this chart with expectation that we'd probably put more into it (other Jenkins jobs that needed migrating), but since there's only two I feel like there must be better place for these workloads.

theseanything:
> Alternative idea: what if we had the `learn_to_rank:generate_data` Rake task exec the Python thing itself when it's done? Would that avoid the whole need for a separate chart without introducing too much ugliness? (don't feel obliged to do that or anything, I'm just spitballing here)

Thanks for the suggestion, do you mean exec within the same container or create a new container? Wasn't entirely sure what the best practise was in terms of co-ordinating workloads in separate containers, my thought was having the everything defined in the K8s resources (cronjob) might make it easier to understand sequence (rather than obfuscated in application code).

Make sense if we exec within the same container, but would need to add the python runtime and dependencies to search-api image. However, that felt like a lot of bloat for the application containers maybe?

PR #1136, opened 2023-05-31
Permit ingress to asset metadata routes for Asset Manager in integration

brucebolt:
We would like to use the integration asset manager to allow development of new features locally, without needing to replicate the entirety of Asset Manager's database and assets.

This will permit ingress to the relevant routes and allow us to obtain that metadata.

The Asset Manager application [requires a user to be authenticated](https://github.com/alphagov/asset-manager/blob/7cf2e0c88f130285a9beb7fed02d0cc648ccda03/app/controllers/application_controller.rb#L8) to access this path, so it will only be available to those with a relevant Signon account or bearer token.

[Trello card](https://trello.com/c/hn6IpttC/545-create-route-controller-basic-view-and-presenter-for-csv-preview-in-frontend)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -143,16 +143,20 @@ govukApplications:
       hosts:
         - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
           paths:
+            - path: /assets/  # Integration only
             - path: /government/uploads/
             - path: /government/assets/
             - path: /media/
             - path: /auth/gds  # Viewing draft assets requires user auth.
+            - path: /whitehall_assets/  # Integration only
         - name: draft-assets.{{ .Values.k8sExternalDomainSuffix }}
           paths:
+            - path: /assets/  # Integration only
             - path: /government/uploads/
             - path: /government/assets/
             - path: /media/
             - path: /auth/gds  # Viewing draft assets requires user auth.
+            - path: /whitehall_assets/  # Integration only
     assetManagerNFS: &assets-nfs assets.blue.integration.govuk-internal.digital
     nginxConfigMap:
       create: false


charts/app-config/values-integration.yaml
```diff
@@ -143,16 +143,20 @@ govukApplications:
       hosts:
         - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
           paths:
+            - path: /assets/
```
sengi, line 146: Might be worth commenting these so it's clear that it's not drift from prod/staging, like:
```suggestion
            - path: /assets/  # Integration only.
```

sengi (approved):

charts/app-config/values-integration.yaml
theseanything, line 146: I think this might take precedence over:

https://github.com/alphagov/govuk-helm-charts/blob/bef5709e4cfa23d37f2259a74cb8e12e638d49a2/charts/app-config/values-integration.yaml#L2892

theseanything: We use the same ALB to route to requests from the asset domain to Asset Manager, Whitehall and Static depending on path - need to make sure this rule isn't going capture requests currently going to Whitehall

sengi, line 146: I think this is ok even though Whitehall has an Ingress in the same ALB group (`assets-origin`) with a path rule for `/assets/whitehall`, because the `whitehall-frontend` Ingress has a higher priority (`group.order: 10` vs our `group.order: 20` here).

I'd recommend checking thoroughly that it's doing what you want after you've rolled this out! :)

sengi: Ah snap hehe :) Let's find out who guessed right 😂

brucebolt: Cool, thanks both.  I'll check the behaviour is correct after it's been rolled out.

brucebolt: Thanks.  I've added this comment to all 4 lines.

sengi:
Ugh sorry about the linter that should obviously just be a fixer 😅 (will get around to that one day!)

charts/app-config/values-integration.yaml
sengi: Just in case you end up having to dive into this: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/ingress/annotations/#group.order (in theory should be ok because lower number => its listener rules come first => should work like you want, but entirely possible I've missed something else)

brucebolt: > Just in case you end up having to dive into this: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.5/guide/ingress/annotations/#group.order (in theory should be ok because whitehall-frontend has the lower number => its listener rules come first => should work like you want, but entirely possible I've missed something else)

Thanks, that doc is a really interesting read.

This appears to have worked as planned, without (as far as I can see) any unintended consequences.

PR #1137, opened 2023-06-01
Change Smokey schedule in integration and staging

MuriloDalRi:
We don't need Smokey to run 24/7 in these environments.

The nightly data sync means a lot Smokey jobs fail and there's a lot of noise in Sentry.

### DIFF
--- a/charts/smokey/templates/cronjob.yaml
+++ b/charts/smokey/templates/cronjob.yaml
@@ -6,7 +6,7 @@ spec:
   concurrencyPolicy: Allow
   failedJobsHistoryLimit: 3
   successfulJobsHistoryLimit: 3
-  schedule: "*/10 * * * *"
+  schedule: {{ .Values.cronSchedule | quote }}
   jobTemplate:
     metadata:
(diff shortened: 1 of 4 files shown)


charts/smokey/templates/cronjob.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   concurrencyPolicy: Allow
   failedJobsHistoryLimit: 3
   successfulJobsHistoryLimit: 3
-  schedule: "*/10 * * * *"
+  schedule: {{ .Values.cronSchedule }}
```
sengi, line 9: I think the validator is complaining because of the spaces in the substituted value here. Enclosing in quotes should fix it, e.g.:
```suggestion
  schedule: {{ .Values.cronSchedule | quote }}
```

sengi (approved):

charts/smokey/templates/cronjob.yaml
sengi: Thanks for fixing this btw!

PR #1138, opened 2023-06-02
Update MaxScanTime on Clamd

nsabri1:
While scanning PDF file of 50mb Clamd reported: `Heuristics.Limits.Exceeded.MaxScanTime`

This can be repeated via direct scan on Clamd container or uploading via Whitehall.
`https://whitehall-admin.integration.publishing.service.gov.uk/government/admin`

Increasing the MaxScanTime from 2 minutes(120000ms) to 4 minutes(240000ms).
https://manpages.debian.org/unstable/clamav-daemon/clamd.conf.5.en.html#MaxScanTime 

Will be doing further testing on this to see the impact this change will have on the resources and larger files upto 500mb.

Tested by running locally on clamd container in integration.
`clamdscan --verbose 25900.pdf
/mnt/asset-manager/tmp/clam-test/25900.pdf: Heuristics.Limits.Exceeded.MaxScanTime FOUND``


charts/asset-manager/templates/clamav-configmap.yaml
```diff
@@ -25,6 +25,7 @@ data:
     # (as opposed to UNIX socket).
     MaxFileSize 500M
     MaxScanSize 1200M
+    maxScanTime 240000
```
sengi, line 28: tiny nit: might as well capitalise it the same as the others 🤷 
```suggestion
    MaxScanTime 240000
```

sengi (approved):
Cool, thanks for tracking down the issue here!

I wonder if on the old system it was just giving up and letting some of these files through when they timed out 🤔

charts/asset-manager/templates/clamav-configmap.yaml
nsabri1: dooooh Thank you Chris saved me pulling my hair out when i would have been troublshooting this :)

PR #1141, opened 2023-06-02
Disable time limit for Clamd scans for testing.

nsabri1:
Applying to Integration only for now.

Removing limit to test actual time it takes to scan certain files.
Will use the data from this exercise to set an somewhat accurate value here.


sengi (approved):
SGTM (esp as you have a ticket open on it already, so shouldn't get forgotten!)

PR #1142, opened 2023-06-02
Add cron task to post a list of apps with undeployed changes to Slack

AgaDufrat:
## What
Add a cron task to post a list of apps with undeployed changes to Slack.
It'll run at 09:00am on every day-of-week from Monday through Friday in production only.
Production Release shows all the relevant environments so there's no point running on Integration, since that only ever shows yesterday's data.

[Trello](https://trello.com/b/16mTLUnP/platform-security-reliability-doing)

## Why 

Following the replatforming, this work intends to replace the retired https://github.com/alphagov/govuk-deploy-lag-badger which only reported on EC2 deployed statuses.

Currently, there’s a risk that people may merge changes to non-continuously-deployed apps and not follow through by deploying to Production. We no longer have the Deploy Lag Badger to notify teams of this, and fixing the badger would have been as much work.

We also get an additional benefit with this work, for apps that are continuously deployed. If a team deploys a branch of a continuously deployed app to Integration, this work will highlight if that branch gets forgotten about!

## Depends on
- https://github.com/alphagov/release/pull/1198

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1736,7 +1736,7 @@ govukApplications:
           ]}}]
       hosts:
         - name: release.{{ .Values.k8sExternalDomainSuffix }}
-    workerEnabled: false
+    workerEnabled: true
     replicaCount: 2
     workerReplicaCount: 1
     podDisruptionBudget:
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1755,6 +1755,10 @@ govukApplications:
       requests:
         cpu: 100m
         memory: 400Mi
+    cronTasks:
+      - name: post-out-of-sync-deploys
+        task: "post_out_of_sync_deploys"
+        schedule: "0 9 * * 1-5"
     extraEnv:
       - name: GDS_SSO_OAUTH_ID
         valueFrom:


ChrisBAshton (approved):
Perhaps worth clarifying in the PR description that despite running "in production only", production Release shows all the relevant environments. And that it would be useless running on Integration, since that only ever shows yesterday's data.

charts/app-config/values-integration.yaml
```diff
@@ -1721,7 +1721,7 @@ govukApplications:
           ]}}]
       hosts:
         - name: release.{{ .Values.k8sExternalDomainSuffix }}
-    workerEnabled: false
+    workerEnabled: true
```
sengi, line 1724: optional: do you need the worker in integration and staging, if it's not posting to Slack from there?

sengi (approved):

charts/app-config/values-integration.yaml
AgaDufrat: yeah, good point. Will remove. Added it for consistency across environments so we have an option to test or debug but may be better to saves the resources.

sengi: Cool, SGTM. Consistency can sometimes be helpful but sounds like probably not in this case 👍 Thanks for being efficient with resources!

PR #1154, opened 2023-06-13
Use separate dedicated DB credential secret for draft-content-store-postgresql-branch

aldavidson:
In the old system, with mongo credentials managed by puppet, it made sense to minimise toil by sharing credentials between draft and live content stores. 

In the new platform, (and in this [migration project](https://trello.com/c/BIQ9GmY5/601-deploy-content-store-proxy-and-content-store-on-postgresql-in-integration-for-the-draft-content-store)) where DB credentials are secrets are managed by AWS, and the databases will be RDS PostgreSQL, it makes more sense to have separate credentials for each.  

This PR changes the name of the secret for the `draft-content-store-postgresql-branch` app to draft-content-store-postgres, in line with other postgresql-related secrets in the charts

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -783,7 +783,7 @@ govukApplications:
       - name: DATABASE_URL
         valueFrom:
           secretKeyRef:
-            name: content-store-postgres
+            name: draft-content-store-postgres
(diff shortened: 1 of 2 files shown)


sengi (approved):
Thanks for catching this!

PR #1157, opened 2023-06-13
Add new app router-api-postgres on integration

catalinailie:
This new app is a fork of Router API but it uses Postgres instead of Mongo.

Trello card: https://trello.com/c/dOXADTQP/1802-spin-up-a-separate-stack-of-router-api-l

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1800,6 +1800,31 @@ govukApplications:
           router-backend-2,\
           router-backend-3/draft_router"
 
+- name: draft-router-api-postgres
+  repoName: router-api-postgres
+  helmValues:
+    <<: *router-api
+    rails:
+      createKeyBaseSecret: false
+    sentry:
+      createSecret: false  # Sentry DSNs are per repo.
+    extraEnv:
+      - name: GDS_SSO_OAUTH_ID
+        valueFrom:
+          secretKeyRef:
+            name: signon-app-draft-router-api
+            key: oauth_id
+      - name: GDS_SSO_OAUTH_SECRET
+        valueFrom:
+          secretKeyRef:
+            name: signon-app-draft-router-api
+            key: oauth_secret
+      - name: DATABASE_URL
+        valueFrom:
+          secretKeyRef:
+            name: router-api-postgres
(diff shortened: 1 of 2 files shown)


sengi (approved):
Is there a design doc or something you can point us at for context on this Postgres migration?

The config itself looks fine, but running a fork of Router API seems a surprising choice to me, so it'd be remiss of me not to try to understand the motivation behind that and what alternatives were considered 🙂

In other words, I'm asking because I'm wondering if perhaps you've missed a much quicker/simpler/safer way to achieve your migration to Postgres.

(I found [this doc](https://docs.google.com/document/d/12IZ-wLJWZ8LqLuu9uRlOJRmuv3_yyaFdHYaz5enegfA/edit#heading=h.7009f2ch7zpw) but that's only about why Postgres rather than how you plan to do the switchover.)

Approving anyway cos I don't want to hold you up.

catalinailie:
> Is there a design doc or something you can point us at for context on this Postgres migration?
> 
> The config itself looks fine, but running a fork of Router API seems a surprising choice to me, so it'd be remiss of me not to try to understand the motivation behind that and what alternatives were considered 🙂
> 
> In other words, I'm asking because I'm wondering if perhaps you've missed a much quicker/simpler/safer way to achieve your migration to Postgres.

Here is the plan we are following for this migration https://github.com/alphagov/router/blob/main/docs/ADR/001-porting-to-postgresql.md

sengi:
Thanks, that's very helpful.

It sounds like doing an ordinary dual-write migration controlled by a feature flag in Router API wasn't considered. If it's not too late, I'd strongly advise considering switching to that approach cos if it fits then it'd be way less work for you and likely eliminates a bunch of risks and dependencies in your project. (background: I've done a bunch of migrations on multi-petabyte, >million-RPS distributed systems in previous roles so I have some experience in this area!)

sengi:
Hope I haven't thrown a spanner in the works here! My only intent here is to help you get your project done quicker and with less hassle, and thus minimise the risk of having to delay the Puppet switchoff. Certainly don't want to hold you up!

More than happy to chat if you have any concerns/questions or just want to bounce ideas around as to how to get your migration done safely and efficiently.

PR #1158, opened 2023-06-13
Remove redundant SSO_PUSH_USER_EMAIL env var for Signon

floehopper:
This effectively reverts #1129 & #1130 now that the value has been hard-coded in alphagov/signon#2200.


sengi (approved):

floehopper:
@sengi Thanks for reviewing. Do I need to do anything else now that I've merged it for it to take effect?

sengi:
Nope, all good! (There are a few dusty corners that need `k rollout restart` and that sort of thing, but plain environment variables from Helm values are the happy path where it's all shiny push-on-green GitOps goodness 🪄✨)

Thanks again for fixing all this so nicely!

PR #1165, opened 2023-06-15
Add chart for Elasticsearch index copy cronjobs.

sengi:
This replaces the Puppet-managed cronjobs for copying Elasticsearch indices from production to staging (and staging to integration).

(Neither this nor its predecessor is in involved in production backups/DR; backups are handled entirely within AWS Managed Elasticsearch.)

For now, this is its own separate Helm chart. It might become a subchart of a wider "environment sync" chart later, or fit into some other structure; we'll see what fits.

I've tried to keep this chart close enough to the standard `helm create` starter chart (minus the templates that we definitely won't use in the foreseeable future and some that can be trivially back added in if we want them, like serviceaccount) that it shouldn't be too difficult to fit it into whatever structure we end up with re what jobs go in what charts. I.e. it's intended to be composable while avoiding unnecessary complexity or esotericism.

[Puppet-managed cronjobs]:
https://github.com/alphagov/govuk-puppet/blob/603dc08/hieradata_aws/class/staging/search.yaml#L4

https://trello.com/c/Nq4oZAOE

Tested:

* installed chart via plain old `helm install $USER-test-search-index-env-sync .` in staging, successfully ran some test snapshot/restore jobs via `k create job --from cronjob/...`
* installs cleanly in integration with `env=integration chart=search-index-env-sync && helm install $USER-$chart ../$chart --values <(
  helm template . --values values-$env.yaml |
  yq e '.|select(.metadata.name=="'$chart'").spec.source.helm.values'
)` (i.e. with the values that ArgoCD will pass in from the app-config chart).

### DIFF
--- a/charts/search-index-env-sync/Chart.yaml
+++ b/charts/search-index-env-sync/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v2
+name: search-index-env-sync
+description: Cronjobs for copying Elasticsearch indices from prod to non-prod
+type: application
+version: 1.0.0
--- a/charts/search-index-env-sync/templates/_helpers.tpl
+++ b/charts/search-index-env-sync/templates/_helpers.tpl
@@ -1,0 +1,52 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "search-index-env-sync.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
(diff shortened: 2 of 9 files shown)


charts/search-index-env-sync/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: search-index-env-sync
+description: A Helm chart for Kubernetes
```
sengi, line 3: Oh flibble I forgot to update this auto-generated file. One sec.

charts/search-index-env-sync/templates/_helpers.tpl
sengi, line 1: This is literally just the auto-generated copypasta from `helm create`, with a couple of unused sections deleted.

charts/search-index-env-sync/Chart.yaml
sengi: Fixed

theseanything (approved):

PR #1166, opened 2023-06-16
Add CronJob to export content-store data from MongoDB to PostgreSQL

aldavidson:
Add a CronJob to  import JSON files to the PostgreSQL (draft-)content-store
Uses an `initContainer` to do the initial export of JSON from the MongoDB (draft-)content-store 

The job is templated so that it can be scheduled against the draft- and live content-stores without unnecessary duplication, but this PR only schedules it against the draft-content-store. Once we've seen this working successfully in integration, I'll add a subsequent PR to schedule it againt the live content-store.

[Trello card](https://trello.com/c/GMoOthyj/671-add-step-to-overnight-environment-sync-job-to-export-mongodb-to-json-and-import-to-postgresql), part of rolling out the [content-store Mongo -> PostgreSQL migration](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb) on [integration](https://trello.com/c/BIQ9GmY5/601-deploy-content-store-proxy-and-content-store-postgresql-branch-in-integration-for-the-draft-content-store)

### DIFF
--- a/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
+++ b/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
@@ -38,7 +38,8 @@ spec:
               image: "content-store:release"
               imagePullPolicy: "Always"
               command: ["rake"]
-              args: ["mongo:export:all[/mongo-export/]"]
+              args:
+                - "mongo:export:all[/mongo-export/]"
               env:
                 - name: MONGODB_URI
                   value: "{{ .Values.contentStoreMongoPostgresCron.mongoExport.mongoDbUri }}"
@@ -52,7 +53,7 @@ spec:
             - name: import-mongo-data-to-postgresql
               image: "content-store-postgresql-branch:release"
               imagePullPolicy: "Always"
-              command: ["bundle exec rake"]
+              command: ["rake"]
               args:
                 - "import:all[/mongo-export/]"
               env:
--- a/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
+++ b/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
@@ -1,0 +1,66 @@
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: "content-store-mongo-to-postgres-cron"
+  labels:
+    app: "content-store-mongo-to-postgres-cron"
+    app.kubernetes.io/component: "content-store-mongo-to-postgres-cron"
+spec:
+  schedule: "15 5 * * *"  
+  jobTemplate:
+    metadata:
+      name: "content-store-mongo-to-postgres-cron"
+      labels:
+        app: "content-store-mongo-to-postgres-cron"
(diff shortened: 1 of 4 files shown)


charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+            runAsGroup: {{ .Values.securityContext.runAsGroup }}
+          restartPolicy: Never
+          volumes:
+            - name: mongo-export
+              emptyDir: {}
+          initContainers:
+            - name: export-mongo-data
+              image: "content-store:release"
+              imagePullPolicy: "Always"
+              command: ["bundle exec rake"]
```
sengi, line 40: `["bundle exec rake"]` means "[exec](https://manpages.ubuntu.com/manpages/jammy/en/man3/exec.3.html) a binary called `bundle exec rake` with no arguments", i.e. it's equivalent to running `bundle\ exec\ rake` or `'bundle exec rake'` in the shell. To run the command you intended you'd need to write `["bundle", "exec", "rake"]` (quotes optional).

You don't need the `bundle exec` though (because we're in a container, so there's only ever going to be one Ruby app there by design) so this is preferred:

```suggestion
              command: ["rake"]
```

Similarly below.

charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+                readOnlyRootFilesystem: true
+              volumeMounts:
+                - name: mongo-export
+                  mountPath: /mongo-export
+          containers:
+            - name: import-mongo-data-to-postgresql
+              image: "content-store-postgresql-branch:release"
+              imagePullPolicy: "Always"
+              command: ["bundle exec rake"]
+              args: ["import:all[/mongo-export/]"]
```
sengi, line 56: YAML list syntax is probably more readable than JSON list syntax for this one.
```suggestion
              args:
                - "import:all[/mongo-export/]"
```

charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -0,0 +1,65 @@
...
+              emptyDir: {}
+          initContainers:
+            - name: export-mongo-data
+              image: "content-store:release"
+              imagePullPolicy: "Always"
+              command: ["bundle exec rake"]
+              args: ["mongo:export:all[/mongo-export/]"]
+              env:
+                - name: MONGODB_URI
+                  value: "{{.Values.mongoExport.mongoDbUri }}"
```
sengi, line 44: You'll need to provide defaults for these new values in `../values.yaml` in order for the chart tests to pass.

aldavidson: That [worked](https://github.com/alphagov/govuk-helm-charts/pull/1166/commits/69e84282168e5a7331f9987d38fc4c7a52825c50)! Thanks!

sengi (approved):
👍

PR #1167, opened 2023-06-18
Fix Init:ImagePullBackoff in content-store-mongo-to-postgres-cronjob

aldavidson:
The CronJob `content-store-mongo-to-postgres-cronjob` added in #1166 was scheduled and tried to execute for the first time this morning, but failed with an `Init:ImagePullBackoff` error.

This usually implies an incorrect container image name/tag (or repo), so I've parameterized the previously hard-coded repo names/tags.

This probably won't fix the issue by itself, but makes them more consistent with the other CronJobs and means I can adjust the values passed in without affecting the job definition. I've also added the `resource` lines from the other jobs while I was at it.

[Trello card](https://trello.com/c/GMoOthyj/671-add-step-to-overnight-environment-sync-job-to-export-mongodb-to-json-and-import-to-postgresql) part of the overall [content-store migration epic](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb) on [integration](https://trello.com/c/BIQ9GmY5/601-deploy-content-store-proxy-and-content-store-postgresql-branch-in-integration-for-the-draft-content-store)

### Update

On further investigation, I _think_ the issue(s) might have been that 

1. I was passing the parameters in the wrong place (should have been in the `govuk-jobs` block, and
2. I also needed to add the container-names into the `imageValues` block

I've made those changes in #[d5a3563](https://github.com/alphagov/govuk-helm-charts/pull/1167/commits/d5a3563c684979ac8803f9c0e0abaaeac90d2d79)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1162,7 +1162,7 @@ govukApplications:
       sageMakerImage: 210287912431.dkr.ecr.eu-west-1.amazonaws.com/search
       sagemakerIamRole: arn:aws:iam::210287912431:role/learn-to-rank-sagemaker
       serviceAccountIamRole: arn:aws:iam::210287912431:role/search-api-learn-to-rank-govuk
-    contentStoreMongoPostgresCron:
+    draftContentStoreMongoPostgresCron:
       mongoExport:
         mongoDbUri: "mongodb://\
           mongo-1.integration.govuk-internal.digital,\
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1164,7 +1164,10 @@ govukApplications:
       serviceAccountIamRole: arn:aws:iam::210287912431:role/search-api-learn-to-rank-govuk
     contentStoreMongoPostgresCron:
       mongoExport:
-        mongoDbUri: "{{ .Values.images.DraftContentStore.extraEnv.MONGODB_URI }}"
+        mongoDbUri: "mongodb://\
+          mongo-1.integration.govuk-internal.digital,\
+          mongo-2.integration.govuk-internal.digital,\
+          mongo-3.integration.govuk-internal.digital/draft_content_store_production"
       postgresImport:
         databaseUrlSecretName: "draft-content-store-postgres"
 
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1170,6 +1162,11 @@ govukApplications:
       sageMakerImage: 210287912431.dkr.ecr.eu-west-1.amazonaws.com/search
       sagemakerIamRole: arn:aws:iam::210287912431:role/learn-to-rank-sagemaker
       serviceAccountIamRole: arn:aws:iam::210287912431:role/search-api-learn-to-rank-govuk
+    contentStoreMongoPostgresCron:
+      mongoExport:
+        mongoDbUri: "{{ .Values.images.DraftContentStore.extraEnv.MONGODB_URI }}"
```
sengi, line 1167: This line seems odd to me on several levels :) Might be easiest to figure out in real-time with a brief chat if you have a minute?

charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -35,14 +35,18 @@ spec:
               emptyDir: {}
           initContainers:
             - name: export-mongo-data
-              image: "content-store:release"
+              image: "{{ .Values.images.DraftContentStore.repository }}:{{ .Values.images.DraftContentStore.tag }}"
```
sengi, line 38: These values will be coming from [here](https://github.com/alphagov/govuk-helm-charts/blob/c55a034/charts/app-config/templates/govuk-application.yaml#L36) and won't be varying by draft/non-draft (since draft and non-draft are rolled out together as one Argo CD `Application`).

Similarly below, `DraftContentStorePostgresqlBranch` becomes `ContentStorePostgresqlBranch` because the template is taking the string from `imageValues` (from values-integration.yaml L1149 above, camelcasing it and filling in the image tag etc. from the correspondingly named file in ../charts/app-config/image-tags/...).
```suggestion
              image: "{{ .Values.images.ContentStore.repository }}:{{ .Values.images.ContentStore.tag }}"
```

(Apologies for the way this oddball `govuk-jobs` chart has ended up — it's become a bit of an ugly example to extend already.)

sengi (approved):

PR #1169, opened 2023-06-20
Change rendering application of CSV previews in integration

brucebolt:
CSV Previews are currently being rendered by Whitehall Frontend. This is being changed to Frontend.

This makes that change in integration only (we will switch the remaining environments in a later PR).

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1007,6 +1007,21 @@ govukApplications:
 
 - name: frontend
   helmValues:
+    ingress:
+      enabled: true
+      annotations:
+        <<: *default-alb-ingress-annotations
+        <<: *alb-ingress-www-waf-ruleset
+        alb.ingress.kubernetes.io/load-balancer-name: assets-origin
+        alb.ingress.kubernetes.io/group.name: assets-origin
+        alb.ingress.kubernetes.io/group.order: '15'
+        alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
+        alb.ingress.kubernetes.io/load-balancer-attributes: *assets-lb-attributes
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: *assets-lb-conditions
+      hosts:
+        - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
+          path: /government/uploads/system/uploads/attachment_data/file/*/*/preview
+          pathType: ImplementationSpecific
     extraEnv:
       - name: GOVUK_NOTIFY_TEMPLATE_ID
         value: *frontend-notify-template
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1007,6 +1007,21 @@ govukApplications:
...
 - name: frontend
   helmValues:
+    ingress:
+      enabled: true
+      annotations:
+        <<: *default-alb-ingress-annotations
+        <<: *alb-ingress-www-waf-ruleset
+        alb.ingress.kubernetes.io/load-balancer-name: assets-origin
+        alb.ingress.kubernetes.io/group.name: assets-origin
+        alb.ingress.kubernetes.io/group.order: '10'
```
sengi, line 1017: The `whitehall-frontend` ingress already has priority 10, so we need a different number here in order to define an unambiguous ordering of rules for the load balancer. How do we choose this number?

- Besides our new `frontend` ingress and `whitehall-frontend`, there are two other ingresses in the same group (`group.name: assets-origin`) i.e. sharing the same load balancer. These are `asset-manager` and `static`.
- The path that we want to handle here at `frontend` (`/government/uploads/system/uploads/attachment_data/file/*/*/preview`) is a strict subset of one of `asset-manager`'s rules (`/government/uploads/`).
- `asset-manager` has priority 20.
- We therefore need to have a higher priority (lower number) than asset-manager.

```suggestion
        alb.ingress.kubernetes.io/group.order: '15'
```

In case you're curious and want all the gory details: https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/guide/ingress/annotations/

sengi (approved):
You've drawn the short straw here in that you've had to touch probably the gnarliest bit of configuration in this whole thing :) Only spotted one small detail out of place, otherwise LGTM! Good job 👍

There are a bunch of limitations on ALB rules that you could potentially run into here, so be sure to check carefully when you roll this out. You'll want to:

- check that your new ingress successfully reconciled with its spec (easiest way is to find your new ingress in the Argo CD web UI under the `frontend` app and look under Events)
- check for errors in the load balancer controller logs (e.g. `k -n cluster-services logs deploy/aws-load-balancer-controller | grep assets-origin`) (you're hoping for `successfully deployed model` and not something about failing to build the model; `jq` is your friend if you need to read any big blobs of JSON that it might spit out)
- check that the [listener rules](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#ELBListenerV2:loadBalancerArn=arn:aws:elasticloadbalancing:eu-west-1:210287912431:loadbalancer/app/assets-origin/e090029f47c2e54d;listenerPort=443;action=None;tab=Rules) on the load balancer make sense to you

charts/app-config/values-integration.yaml
brucebolt: Thanks.  I'll set this to 15 now.

`whitehall-frontend` will be going away within months, so we can remove that (and re-adjust the orders) at that point.

PR #1171, opened 2023-06-20
Permit access to Frontend for draft CSV previews in integration

brucebolt:
This adds ingress to Frontend at the draft asset host for CSV previews and the assets needed to render the page.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1024,6 +1024,10 @@ govukApplications:
           pathType: ImplementationSpecific
         - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
           path: /assets/frontend/
+        - name: draft-assets.{{ .Values.k8sExternalDomainSuffix }}
+          path: /government/uploads/system/uploads/attachment_data/file/*/*/preview
+        - name: draft-assets.{{ .Values.k8sExternalDomainSuffix }}
+          path: /assets/frontend/
     extraEnv:
       - name: GOVUK_NOTIFY_TEMPLATE_ID
         value: *frontend-notify-template


sengi (approved):

sengi:
Controller doesn't look happy this time :(

`k -n cluster-services logs deploy/aws-load-balancer-controller | grep assets-origin`

```json
{
  "level": "error",
  "ts": 1687266587.2025476,
  "logger": "controller-runtime.manager.controller.ingress",
  "msg": "Reconciler error",
  "name": "assets-origin",
  "namespace": "",
  "error": "ingress: apps/frontend: prefix path shouldn't contain wildcards: /government/uploads/system/uploads/attachment_data/file/*/*/preview"
}
```

brucebolt:
Yeah, just looking at this in Argo's event log.

I've missed off `pathType: ImplementationSpecific` compared to the other containing wildcards.  I can't see any mention of wildcards in the docs, but maybe this is what I need toadd?

sengi:
Ah snap, was just about to post this :) Yeah adding `pathType: ImplementationSpecific` should fix it.

brucebolt:
Thanks @sengi.  Have raised https://github.com/alphagov/govuk-helm-charts/pull/1172 to fix this.

sengi:
Yeah I can see how it'd be a tricky one to find out about. The wildcard pattern matching isn't a feature of aws-lb-controller, nor a feature of Ingress. It's a [feature of the ELBv2 API](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#path-conditions).

brucebolt:
Ah, that makes sense.  I was starting to wonder if it was something custom that we'd written, but this explains why adding that `pathType` works here.

PR #1173, opened 2023-06-21
Change `draft-assets` routing to use draft-frontend for CSV Previews

brucebolt:
We run two instances of the Frontend application: live and draft.

CSV Previews are currently being routed to the live version for both live and draft assets. This is not correct as the live version of Frontend does not have access to the Draft Content Store.

Therefore updating the routing so draft requests come to the draft frontend instance.


sengi (approved):
Cool 👍

PR #1176, opened 2023-06-21
Remove alb.ingress.kubernetes.io/security-groups from frontend and draft frontend

davidgisbey:
We've got an ongoing issue where draft asset manger is not working in production. We've received many Zendesk tickets stating people can't see draft PDFs

We were pointed to a similar incident which was resolves via these PRs:

https://github.com/alphagov/govuk-helm-charts/pull/1024
https://github.com/alphagov/govuk-helm-charts/pull/1025

So we're going to remove the same lines from the recent [PR](https://github.com/alphagov/govuk-helm-charts/pull/1174) that was merged for CSV rendering that includes the same.


KludgeKML (approved):
Looks good to me.

sengi (approved):

brucebolt (approved):

PR #1177, opened 2023-06-22
Update hostnames and ordering for draft CSV previews

brucebolt:
We are currently routing all requests for `draft-assets` to the live instance of Frontend first.

These requests should go to the draft Frontend instance instead.

Therefore updating the ordering so draft-frontend has higher priority and specifying a specific hostname for the draft Frontend instance so we match only draft requests.

[Trello card](https://trello.com/c/y2vagYMl)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1101,7 +1101,7 @@ govukApplications:
         <<: *alb-ingress-www-waf-ruleset
         alb.ingress.kubernetes.io/load-balancer-name: assets-origin
         alb.ingress.kubernetes.io/group.name: assets-origin
-        alb.ingress.kubernetes.io/group.order: '15'
+        alb.ingress.kubernetes.io/group.order: '16'
         alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
         alb.ingress.kubernetes.io/load-balancer-attributes: *assets-lb-attributes
         alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: *assets-lb-conditions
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1053,7 +1053,7 @@ govukApplications:
         <<: *alb-ingress-www-waf-ruleset
         alb.ingress.kubernetes.io/load-balancer-name: assets-origin
         alb.ingress.kubernetes.io/group.name: assets-origin
-        alb.ingress.kubernetes.io/group.order: '15'
+        alb.ingress.kubernetes.io/group.order: '16'
         alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
```
KludgeKML, line 1057: Is this going to cause problems? It was removing a similar line yesterday that fixed the draft assets problem.

brucebolt: Hmm, I _think_ we do want integration to be access limited.

KludgeKML: Ah, it's just integration - yes, sorry ignore me!

sengi (commented):
Hold up a minute. The reason we had this complex and brittle setup with the `alb.*group` annotations was because some of the draft stuff was being served by whitehall-frontend, which doesn't have a draft version. If we now have draft-frontend for that, there's no sense in tying ourselves into a pretzel trying to keep serving draft and non-draft assets from the same load balancer.

charts/app-config/values-integration.yaml
sengi: I dunno — staging definitely is supposed to be GDS-only but don't forget we have this whacky use of integration for training content people in other departments.

brucebolt:
I started looking into the decoupling of these into two load balancers, but we don't have a draft instance of Asset Manager (it's the same app serving both live and draft).

sengi:
> I started looking into the decoupling of these into two load balancers, but we don't have a draft instance of Asset Manager (it's the same app serving both live and draft).

Ohh, yeah of course — sorry my bad.

charts/app-config/values-integration.yaml
```diff
@@ -1109,10 +1109,13 @@ govukApplications:
         <<: *alb-ingress-www-waf-ruleset
         alb.ingress.kubernetes.io/load-balancer-name: assets-origin
         alb.ingress.kubernetes.io/group.name: assets-origin
-        alb.ingress.kubernetes.io/group.order: '16'
+        alb.ingress.kubernetes.io/group.order: '15'
         alb.ingress.kubernetes.io/security-groups: eks_ingress_www_origin
         alb.ingress.kubernetes.io/load-balancer-attributes: *assets-lb-attributes
-        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: *assets-lb-conditions
+        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: &draft-assets-lb-conditions >
```
sengi, line 1115: Remove unused anchor
```suggestion
        alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
```

sengi (approved):

charts/app-config/values-integration.yaml
brucebolt: Thanks.  Will remove this.

PR #1180, opened 2023-06-24
Export, compress & upload the draft-content-store-postgresql-branch DB to S3

aldavidson:
Export, compress, and upload-to-s3, the newly-import PostgreSQL database at the end of the mongo-to-postgres cronjob for `draft-content-store-postgresql-branch` in integration. This is needed so that the downstream consumers of the mongo database (Data Services, content-data app etc) can update their import scripts to import PostgreSQL rather than Mongo. 

[Trello card](https://trello.com/c/RxgagJzD/737-add-overnight-pgdump-of-content-store-postgresql-rds-instance-to-s3), part of the [overall epic](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb)

I don't know how well this is going to work until I try it, this solution is quite verbose and could probably be DRY-ed up quite a lot. There are also a couple of questions outstanding:

- would we want to do this as part of _this_ job (makes sense to do it as soon as the import is finished), or as part of a separate port of the `govuk_env_sync` job [from govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/782ae60e32118bdd50e3d4856d6bfd65ef010eaa/modules/govuk_env_sync/files/govuk_env_sync.sh)? 
- It also doesn't _need_ to all be run on the app images, there are probably simpler trusted images we could use (a postgres one for the first, maybe just a generic ubuntu one for all the other steps except the last s3 one ?)

### DIFF
--- a/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
+++ b/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
@@ -142,7 +142,6 @@ spec:
                   mountPath: /mongo-export
                 - name: app-tmp
                   mountPath: /tmp
-          containers:
             - name: import-mongo-data-to-postgresql
               image: "{{ .Values.images.ContentStorePostgresqlBranch.repository }}:{{ .Values.images.ContentStorePostgresqlBranch.tag }}"
               imagePullPolicy: "Always"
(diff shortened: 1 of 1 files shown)


sengi (approved):
This approach could potentially take a lot of local disk space on the kubelet — would be safer to stream it directly to `aws s3 cp` but happy to see how this goes for now since there's a lot of value in moving quickly on this.

charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -167,3 +166,103 @@ spec:
...
+                readOnlyRootFilesystem: true
+              volumeMounts:
+                - name: app-tmp
+                  mountPath: /tmp
+                - name: pg-dump
+                  mountPath: /pg-dump
+            - name: compress-pg-dump
+              image: "{{ .Values.images.ContentStorePostgresqlBranch.repository }}:{{ .Values.images.ContentStorePostgresqlBranch.tag }}"
+              imagePullPolicy: "Always"
+              command: ["tar"]
```
sengi, line 200: I'd probably at least do the tar/gzip in a shell pipeline so as to reduce the impact on local disk.

sengi:
I recently added the Postgres client tools to that `github-cli` image of ours btw, in case that helps.

aldavidson:
Just ran out of time - I'm off on holiday now, back August 14th. I'll have to park this and come back to it when I return

aldavidson:
> Just ran out of time - I'm off on holiday now, back August 14th. I'll have to park this and come back to it when I return

I am now back, and having come back to this with refreshed eyes & brain, I realised that I could simplify it quite a bit as the `-F c` flag to pg_dump produces a file that's [already compressed](https://www.postgresql.org/docs/current/app-pgdump.html) :
> The most flexible output file formats are the “custom” format (-Fc) and the “directory” format (-Fd). They (...)  are compressed by default. 

So we're now down to two steps - 1) dump, 2) upload to s3. Both steps also now use the `github-cli` image as you suggested.
I suggest we go with this, and once it's working, I'll then see if I can stream it directly to s3 in a single step as a separate PR.

PR #1181, opened 2023-06-26
Turn CSP report only off for Feedback, Finder-Frontend and Smart Answers

kevindew:
This is a draft because I don't want this applied until Tuesday morning.

This PR turns off Content Security Policy (CSP) report only mode for the applications owned by the User Experience Measurement (UXM) team. Turning report only mode off makes the CSP actually apply restrictions, whereas previously it would report and not block.

This is the first time I've created a PR for this repo so I've got a couple of questions:

1. Is the approach I've used to configuring the variables going to work, I'm assuming a variable set in`charts/app-config/values-integration.yaml` overrides something set in `charts/app-config/values.yaml`?
2. Have I got the conditional correct?
3. I'm assuming that once merged this just applies automatically to each environment and doesn't need any extra deploy steps, is that correct?

### DIFF
--- a/charts/app-config/templates/env-configmap.yaml
+++ b/charts/app-config/templates/env-configmap.yaml
@@ -17,7 +17,9 @@ data:
   GOVUK_APP_DOMAIN: ""
   GOVUK_APP_DOMAIN_EXTERNAL: {{ .Values.publishingDomainSuffix }}
   GOVUK_ASSET_ROOT: https://{{ .Values.assetsDomain }}
-  GOVUK_CSP_REPORT_ONLY: {{ .Values.cspReportOnly | quote }}
+  {{- if .Values.cspReportOnly }}
+  GOVUK_CSP_REPORT_ONLY: "true"
+  {{- end }}
   GOVUK_CSP_REPORT_URI: {{ .Values.cspReportURI | quote }}
   GOVUK_ENVIRONMENT: {{ .Values.govukEnvironment }}
   GOVUK_ENVIRONMENT_NAME: {{ .Values.govukEnvironment }}
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -980,6 +980,7 @@ govukApplications:
 
 - name: feedback
   helmValues:
+    cspReportOnly: false
     extraEnv:
       - name: GOVUK_NOTIFY_ACCESSIBLE_FORMAT_REQUEST_REPLY_TO_ID
         value: 93109fea-34d9-4c38-ac7e-1ebc75e7416b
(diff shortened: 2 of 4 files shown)


charts/app-config/templates/env-configmap.yaml
```diff
@@ -17,7 +17,9 @@ data:
   GOVUK_APP_DOMAIN: ""
   GOVUK_APP_DOMAIN_EXTERNAL: {{ .Values.publishingDomainSuffix }}
   GOVUK_ASSET_ROOT: https://{{ .Values.assetsDomain }}
-  GOVUK_CSP_REPORT_ONLY: {{ .Values.cspReportOnly | quote }}
+  {{- if .Values.cspReportOnly }}
+  GOVUK_CSP_REPORT_ONLY: "true"
+  {{- end }}
```
sengi, line 22:
```suggestion
```

charts/app-config/values-integration.yaml
```diff
@@ -980,6 +980,7 @@ govukApplications:
 
 - name: feedback
   helmValues:
+    cspReportOnly: false
     extraEnv:
```
sengi, line 984: GitHub still doesn't support suggestions on lines that aren't already modified, so I'll use `feedback` to demonstrate the change you want, even though `feedback` isn't one of the apps whose environment you want to change.

You're gonna need to specify this on every app that cares about the flag (except the three where you're removing the flag, naturally).

I know it's a bit ugly, but it's the best we can do given that this flag is inverted in its sense (i.e. it means "disable this new feature" rather than "enable this new feature"). Alternatively you could flip the sense of the flag in its implementation so that you can just enable the feature app-by-app instead of having to disable it in most apps and then gradually un-disable it 🙂

```suggestion
    extraEnv:
      - name: GOVUK_CSP_REPORT_ONLY
        value: "true"
```

sengi (commented):
First lemme just check I've properly understood your goal. Currently all apps have `GOVUK_CSP_REPORT_ONLY=true` in their environment. You want to remove this from `finder-frontend`, `feedback` and `smart-answers`. Is that right?

If so, you'll need to remove it from that configmap so that you can set it for everything apart from `finder-frontend`, `feedback` and `smart-answers`.

kevindew:
> First lemme just check I've properly understood your goal. Currently all apps have GOVUK_CSP_REPORT_ONLY=true in their environment. You want to remove this from finder-frontend, feedback and smart-answers. Is that right?

Yup, that is correct. 

Ah sure so I can't use another app to disable the configmap?
Ok sure, I'll set up the env var everywhere. That seems easier than re-releasing the gem.

sengi:
Realised I didn't answer your question about rollout. Most changes don't need any extra steps, but environment variables that come from a configmap are an exception to this, at least for now. See [the instructions at the top of the configmap](https://github.com/alphagov/govuk-helm-charts/blob/individual-env-vars-for-csp/charts/app-config/templates/env-configmap.yaml#L1-L2).

kevindew:
Thanks Chris - that's helpful. I'll close this PR and create a new one since it'll be quite different. I'll tag you for a review.

PR #1186, opened 2023-06-27
Revert "Permit ingress to asset metadata routes for Asset Manager in integration"

brucebolt:
Reverts alphagov/govuk-helm-charts#1136.

The development of CSV Previews in the Frontend application is complete, so we no longer need to permit ingress to these endpoints from local developer machines.

This change had a side effect of blocking the loading of all CSS and JavaScript on the preview pages (as they existed under the `/assets` path), so should be reverted to allow these pages to render correctly.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -139,20 +139,16 @@ govukApplications:
       hosts:
         - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
           paths:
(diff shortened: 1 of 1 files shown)


CristinaRO (approved):
This looks good to me, in that it does exactly what it says and reverts the original changes!

I don't know if it needs more approvals from a different team, because this repo is not listed on the [team-app ownership page](https://docs.publishing.service.gov.uk/apps.html#apps-by-team).

PR #1191, opened 2023-07-03
Swap in draft content store proxy via service app selector

aldavidson:
Second attempt at swapping-in `draft-content-store-proxy` as per the previous attempt in #1188  - this time with an additional Service that should route http://draft-content-store/ to the draft-content-store-proxy pods. 

[Trello card](https://trello.com/c/xeJcwf6w/707-deploy-content-store-proxy-in-integration-for-the-draft-content-store), part of rolling out the [content-store mongo->postgresql migration](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb)  on [integration](https://trello.com/c/BIQ9GmY5/601-deploy-content-store-proxy-and-content-store-postgresql-branch-in-integration-for-the-draft-content-store)

### DIFF
--- a/charts/app-config/templates/draft-content-store-proxy-service.yaml
+++ b/charts/app-config/templates/draft-content-store-proxy-service.yaml
@@ -5,10 +5,6 @@ kind: Service
 metadata:
   name: draft-content-store
 spec:
-  ports:
-    - name: app
-      port: 80
-      targetPort: 8080
-  selector:
-    app: draft-content-store-proxy
+  type: ExternalName
+  externalName: draft-content-store-proxy
 {{- end }}
--- a/charts/app-config/templates/draft-content-store-proxy-service.yaml
+++ b/charts/app-config/templates/draft-content-store-proxy-service.yaml
@@ -6,5 +6,5 @@ metadata:
   name: draft-content-store
 spec:
   type: ExternalName
-  externalName: draft-content-store-proxy
+  externalName: draft-content-store-proxy.apps.svc.cluster.local
 {{- end }}
--- a/charts/app-config/templates/draft-content-store-proxy-service.yaml
+++ b/charts/app-config/templates/draft-content-store-proxy-service.yaml
@@ -1,0 +1,14 @@
+{{ if .Values.enableDraftContentStoreProxy }}
+# route all requests for draft-content-store via draft-content-store-proxy
+apiVersion: v1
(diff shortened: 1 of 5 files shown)


charts/app-config/templates/draft-content-store-proxy-service.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Service
+metadata:
+  name: draft-content-store
+spec:
+  ports:
+    - name: app
+      port: 80
+      targetPort: 8080
+  selector:
+    app: draft-content-store-proxy
```
sengi, line 13: You probably want a service of type [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname), rather than duplicating all the endpoints and so on. It'd also be clearer that this is an alias that way.

(Does naming the Argo app for the proxy `content-store` not quite do what you want?)

aldavidson: I've tried with an `ExternalName` service type (it does look simpler that way)

aldavidson: > (Does naming the Argo app for the proxy content-store not quite do what you want?)

I thought that might be more confusing in Argo and so on, so I wanted to try this way first. If this doesn't work, that's the next thing to try.

charts/app-config/templates/draft-content-store-proxy-service.yaml
```diff
@@ -0,0 +1,10 @@
+{{ if .Values.enableDraftContentStoreProxy }}
+# route all requests for draft-content-store via draft-content-store-proxy
+apiVersion: v1
+kind: Service
+metadata:
+  name: draft-content-store
+spec:
+  type: ExternalName
+  externalName: draft-content-store-proxy
```
sengi, line 9: This needs to be the fully-qualified name unfortunately (but that's ok because this doesn't need to work with any other namespace).
```suggestion
  externalName: draft-content-store-proxy.apps.svc.cluster.local
```

sengi (approved):

charts/app-config/templates/draft-content-store-proxy-service.yaml
sengi: Ah I see, fair enough :)

PR #1193, opened 2023-07-03
Remove Cache Clearing Service

AgaDufrat:
## Context

Cache Clearing Service was build in August 2018. At the time GOV.UK pages would
typically return 30 minute cached version of a page, which was an unacceptably
long delay for publishers. The service listens for publishing events and purges
the registered routes from the Varnish and Fastly caches. This purging however
doesn’t remove all variants of the content and thus cannot be relied upon for
all circumstances. Consequently:
- any pages that are accessed with query strings or are within prefix routes
  won’t be purged from the cache by Cache Clearing Service.
- users could experience inconsistent experiences where one user can see a
  changed page and the other not
- GOV.UK HTTP responses incorrectly tell a client they can cache a page for the
  cache period when this isn’t strictly true

Content folks have been notified in [#govuk-2ndline-content](https://gds.slack.com/archives/CMW1476KZ/p1688395861315089) and [#content-community](https://gds.slack.com/archives/CACV3TACU/p1688398664580089)

## Why

Currently, the cache period significantly reduced to 5 minutes. 
We also have a Jenkins task equivalent to cache clearing service
https://github.com/alphagov/govuk-puppet/pull/11663 and documented 
how to [Purge a page from cache](https://docs.publishing.service.gov.uk/manual/purge-cache)

For those reasons the service is deemed as redundant. Removing it will
reduce maintenance burden for developers and reduce cost of hosting the
service.

[GIFT week summer 2023 card](https://trello.com/c/ENPwB2vT/41-retire-cache-clearing-service)
[Tech debt card](https://trello.com/c/TX9wLihh/101-cache-clearing-app-exists-and-we-dont-think-its-necessary-replacing-cache-clearing-app-only-clears-path-not-query-strings)

## Review questions

Since the process of retiring apps hosted on Kubernetes is not documented I'm wondering if I should first mark the workers as disabled `workerEnabled: true`. In an event of rolling back it would still be a matter of reverting this commit.

## Follow up tasks

- [ ] Remove `cache-clearing-service-rabbitmq` from AWS Secrets Manager
- [ ] https://github.com/alphagov/govuk-aws/pull/1740
- [ ] https://github.com/alphagov/govuk-puppet/pull/12104
- [ ] https://github.com/alphagov/govuk-puppet/pull/12105

Princess Bride Dead GIF


sengi (approved):
Woohoo!

PR #1195, opened 2023-07-03
Add Open Telemetry exporter endpoint

theseanything:
This adds a env var that can be used be Open Telemetry SDKs to export data to Jaeger (distributive tracing tooling). This env var is only enabled in integration as part of a demo of Jaeger and shouldn't affect other environments.

---
This work is being done as part of GIFT week, and should be removed if not implemented fully.


nsabri1 (approved):
LGTM

sengi (approved):

PR #1200, opened 2023-07-04
Remove unnecessary perl script that sets govuk_request_id

tunylund:
The same functionality can now be achieved with an nginx `map` directive.

I'm not sure if anything else needs the PERL module?

### DIFF
--- a/charts/app-config/templates/router-nginx-config.tpl
+++ b/charts/app-config/templates/router-nginx-config.tpl
@@ -36,23 +36,10 @@ http {
   }';
 
   # Set GOVUK-Request-Id if not set
-  # See http://nginx.org/en/docs/http/ngx_http_perl_module.html
-  perl_modules perl/lib;
-  perl_set $govuk_request_id '
-    sub {
-      my $r = shift;
-      my $current_header = $r->header_in("GOVUK-Request-Id");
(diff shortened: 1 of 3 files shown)


sengi (approved):
Awesome, thanks for this!

I think there's one other usage of Perl left, in charts/app-config/templates/router-nginx-config.tpl, but we can trivially move that into Router itself and then we can drop Perl altogether.

tunylund:
I changed the router-nginx-confg.tpl as well.

Is there a cute way to test these changes so that I won't break everything in case i did something stupid?

sengi:
> I changed the router-nginx-confg.tpl as well.

Thanks! 🌟 

> Is there a cute way to test these changes so that I won't break everything in case i did something stupid?

Great question - yeah hopefully [this recipe](https://github.com/alphagov/govuk-helm-charts#installing-an-application-chart-without-argo-cd) should do the trick. It's a bit hacky but should just about do it, I think.

sengi:
Anything I can do to help with this? (no problem at all if you're just busy with other stuff, only asking since it's been a while and I'd love to see the back of the Perl module!)

tunylund:
It would be great if you (or someone from platform engineering) could help test this pr. I haven't used helm-charts in GovUk before so it would be best to pair on this.

PR #1203, opened 2023-07-04
Perhaps raw numbers are invalid?

KludgeKML:
The changes to add SITE_IMPROVE env vars appear not to be syncing in Argo, and the current main suspect is the raw number, so wrap it in speechmarks.


theseanything (approved):

sengi (approved):
Ah yeah this is a common gotcha; env vars are always strings and it'll just enforce this via the JSON schema rather than coercing other types to string.

sengi:
Not sure why chart-testing doesn't catch it pre-merge though 🤔

KludgeKML:
Thanks both! I can see it's working now in Argo. :)

PR #1205, opened 2023-07-05
[DO NOT MERGE] Add cron job for sending review reminders on Whitehall

davidgisbey:
**Do not merge until rake task is merged in on Whitehall https://github.com/alphagov/whitehall/pull/7895/commits/4c1400b8fad7b71166e468aeee7f50c037d295d2**
## Description

We're adding functionality for publishers to set a review date for documents. The send_review_reminders rake task sends emails to these publishers on their selected date.

This job needs to run at 10.30am daily.


charts/app-config/values-integration.yaml
```diff
@@ -2958,6 +2958,9 @@ govukApplications:
       - name: send-notifications
         task: "publisher_notifications:send"
         schedule: "18 1 * * *"
+      - name: send-review-reminders
+        task: "send_review_reminders"
+        schedule: "30 10 * * *"
```
ollietreend, line 2963: Do we know which time would be best for publishers to receive these emails?

10:30am seems as good as any other time 🤷🏻 I just wondered if it was chosen randomly or if we've seen research to say this time is best.

davidgisbey: The thought process from Sally was that it wouldn't get lost in the early morning email spam. I'm sort of of the opinion that it would be better to send it at 5 or something so that people can work it into their day however they want to. So i think it could change based on feedback

PR #1211, opened 2023-07-07
Add ExternalSecret for content-store-postgresql-branch RDS access

aldavidson:
so that we can [rollout PostgreSQL on integration for the live (non-draft) content-store](https://trello.com/c/HMibQBll/735-roll-out-content-store-proxy-for-live-content-store-in-integration)


charts/external-secrets/templates/content-store-postgresql-branch/postgresql.yaml
sengi, line 1: I'd put this under `templates/content-store` like the existing one, as it's slightly clearer and saves moving it later.

sengi (approved):

charts/external-secrets/templates/content-store-postgresql-branch/postgresql.yaml
aldavidson: ooh, good call - done

PR #1212, opened 2023-07-07
Add CronJob to export live (integration) content-store from mongo to postgres

aldavidson:
essentially duplicating the existing job which does the same for draft-content-store.

There's very probably a much more elegant way of doing this which avoids the duplication, but my helm-fu is not yet up to the task, sorry.

### DIFF
--- a/charts/govuk-jobs/values.yaml
+++ b/charts/govuk-jobs/values.yaml
@@ -45,3 +45,10 @@ draftContentStoreMongoPostgresCron:
     mongoDbUri: ""
   postgresImport:
     databaseUrlSecretName: "draft-content-store-postgres"
+
+contentStoreMongoPostgresCron:
+  schedule: "15 5 * * *"
+  mongoExport:
+    mongoDbUri: ""
+  postgresImport:
+    databaseUrlSecretName: "content-store-postgres"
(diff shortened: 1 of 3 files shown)


charts/govuk-jobs/values.yaml
```diff
@@ -45,3 +45,10 @@ draftContentStoreMongoPostgresCron:
     mongoDbUri: ""
   postgresImport:
     databaseUrlSecretName: "draft-content-store-postgres"
+
+contentStoreMongoPostgresCron:
+  schedule: "15 5 * * *"
```
sengi, line 50: It'd be cool to run these at slightly different times so as to reduce peak resource usage. I'm guessing they don't take hours to run?

sengi (approved):
Cool, no worries about the duplication cos this is just temporary.

charts/govuk-jobs/values.yaml
aldavidson: Sure. 

They need to run after the overnight env_sync job finishes, which IIRC is sometime around 4.15, and the mongo-export part of this completes in a couple of minutes, but the postgres-import part takes about 1h 15m. 

So how about I keep the live one at 5:15, and start the draft one at 6:30 ?

sengi: Thanks!

PR #1213, opened 2023-07-08
Swap-in the (live) content-store-proxy on integration

aldavidson:
using the same method as for the draft-content-store in #1196 

[Trello card](https://trello.com/c/HMibQBll/735-roll-out-content-store-proxy-postgresql-app-for-live-content-store-in-integration), overall [epic](https://trello.com/c/C1BQDFTG/502-plan-for-migrating-content-store-off-mongodb)

PR #1219, opened 2023-07-10
[Do not merge - only in case of emergency] rollback the integration live content-store to just the mongo version

aldavidson:
As I'm away until August 14th, and the rollout of dual-running on mongo & postgres for the integration _live_ content-store had to be split across several PRs (#1213 , #1215 and #1216), this is a single draft PR to use in case of emergency, which will revert those 3 in one go


aldavidson:
closing un-merged - we're extremely unlikely to revert this now, it's well settled-in

PR #1220, opened 2023-07-10
Update versions of 3rd-party GitHub Actions.

sengi:
Also:
- Avoid creating tags and releases for the unversioned charts (the ones we deploy via Argo CD), since having those releases/tags lying around is kinda confusing as they don't represent anything useful.
- Update the k8s version that we validate against with kubeconform.

Tested: regression-tested the chart-releaser workflow by pushing a chart version update (with the workflow temporarily patched to trigger from the PR branch) and verified that the release worked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,6 @@ name: linting
 on: [push]
 jobs:
   chart-testing:
-    name: Chart testing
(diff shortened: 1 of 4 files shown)


theseanything (approved):

sengi:
(Fixed a link that I broke in the README.)

PR #1222, opened 2023-07-10
Add DIGITIAL_IDENTITY_ENVIRONMENT vars to relevant apps

KludgeKML:
- Frontend, Email Alert API and Static all use the GovukPersonalisation::URLs calls which return DI login links
- Currently these all go to live, update to use the env var put in place for this.

https://github.com/alphagov/govuk_personalisation/blob/main/lib/govuk_personalisation/urls.rb#L97-L104

Have confirmed with OneLogin that these are the right hostnames for these environments.

Part of this work:
https://trello.com/c/SiAQ3dZU/2086-redirect-govuk-account-to-homeaccountgovuk

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -931,6 +931,8 @@ govukApplications:
             secretKeyRef:
               name: email-alert-api-postgres
               key: DATABASE_URL
+        - name: DIGITAL_IDENTITY_ENVIRONMENT
+          value: integration
         - name: EMAIL_ALERT_AUTH_TOKEN
           valueFrom:
             secretKeyRef:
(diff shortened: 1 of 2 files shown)


theseanything (approved):

KludgeKML:
Could be... just to check, is GOVUK_ENVIRONMENT set in production? (we'd have to ignore that, but that's fine to do within the gem)

sengi:
Ah I see, is the DI production environment not called production then?

KludgeKML:
Yeah, the domain we want in production is just home.account.gov.uk - but that's easy enough. I'll revert this PR, and we can replace with: https://github.com/alphagov/govuk_personalisation/pull/53

sengi:
Ah that's brilliant, thanks! Usually could make a sound argument for either way (explicit config and no special logic in the app, or zero/minimal config and all the logic in the app) but since there's already some logic in the app related to this config item anyway, I think this is the way to go 👍

PR #1224, opened 2023-07-11
Turn on GOV.UK CSP for CIOP owned rendering apps

kevindew:
This turns on the GOV.UK Content Security Policy (CSP) in Email Alert Frontend and Licence Finder, the two apps on www.gov.uk that the Content Interactions on Platform (CIOP) team own.

The Licence Finder application is due to be retired very soon, but we don't anticipate any problems - should we get any we'll just turn this back on until the app is retired.


KludgeKML (approved):
LGTM

PR #1225, opened 2023-07-11
Enable disributed tracing in staging

theseanything:
This enables the Open Telemetry instrumentation for our Rails applications in Staging. Also sets the head sampling rate to 0.3 so we aren't recording everything.

Tested in integration.

### DIFF
--- a/charts/app-config/templates/env-configmap.yaml
+++ b/charts/app-config/templates/env-configmap.yaml
@@ -45,10 +45,12 @@ data:
   # Endpoint used to export Open Telemetry data. Only set for integration as
   # being used to demo Jaeger as part of GIFT week. Should be removed if not
   # Jaeger not implemented.
-  {{- if eq .Values.govukEnvironment "integration" }}
-  OTEL_EXPORTER_OTLP_ENDPOINT: "http://tempo-distributor.monitoring.svc.cluster.local:4318"
-  OTEL_RUBY_INSTRUMENTATION_RACK_CONFIG_OPTS: "untraced_endpoints=/healthcheck/live"
+  {{- if ne .Values.govukEnvironment "production" }}
   ENABLE_OPEN_TELEMETRY: "true"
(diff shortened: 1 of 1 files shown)


charts/app-config/templates/env-configmap.yaml
```diff
@@ -45,10 +45,12 @@ data:
   # Endpoint used to export Open Telemetry data. Only set for integration as
   # being used to demo Jaeger as part of GIFT week. Should be removed if not
   # Jaeger not implemented.
-  {{- if eq .Values.govukEnvironment "integration" }}
+  {{- if ne .Values.govukEnvironment "production" }}
+  ENABLE_OPEN_TELEMETRY: "true"
   OTEL_EXPORTER_OTLP_ENDPOINT: "http://tempo-distributor.monitoring.svc.cluster.local:4318"
+  OTEL_TRACES_SAMPLER: "traceidratio"
+  OTEL_TRACES_SAMPLER_ARG: '0.3'
```
sengi, line 52: 30% seems like a lot, but OTOH there's normally very little traffic in staging so I guess might be a pain to test otherwise. No problem for now but I'd recommend revising this down at some point later. (I'm thinking about use cases like load testing and that sort of thing.)

sengi (approved):

charts/app-config/templates/env-configmap.yaml
theseanything: Yeah, that's a good point - I was unsure what to go for, but in integration/staging it seemed fine due to low level of traffic. Will have a look tuning this number for staging and prod.

sengi: For prod we probably want either a very small proportion like < 1/1000 or (perhaps better) default to on-demand only so we can trigger it via a header or something but not have to worry about any impact on serving capacity / reliability when we're not actively debugging something.

PR #1231, opened 2023-07-18
Change rendering app of /government/placeholder in integration

CristinaRO:
**Depends on** https://github.com/alphagov/frontend/pull/3696

Step 2 in removing the rendering of /government/placeholder from whitehall: route the path to frontend instead of whitehall-frontend.

[Trello](https://trello.com/c/81j2sbeq/710-move-rendering-of-https-wwwgovuk-government-placeholder-out-of-whitehall)


charts/app-config/values-integration.yaml
```diff
@@ -3085,8 +3087,6 @@ govukApplications:
           alb.ingress.kubernetes.io/load-balancer-attributes: *assets-lb-attributes
           alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: *assets-lb-conditions
         hosts:
-          - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
-            path: /government/placeholder/
           - name: assets-origin.{{ .Values.k8sExternalDomainSuffix }}
             path: /assets/whitehall/
```
brucebolt, line 3091: I _think_ the entirety of the `ingress` section can now be deleted, since this is the last thing to be rendered by Whitehall on the assets domai.  The other path (`/assets/whitehall`) only serves the CSS and JavaScript for rendered pages, so that shouldn't be needed anymore either.

CristinaRO: So, to make sure: 
- removing all ingress rules in this PR will only affect integration anyway
- will we know just by manual smoke-testing if it turns out that something was still relying on those rules? (versus potentially not finding out for a while)

brucebolt: > * removing all ingress rules in this PR will only affect integration anyway
Yes

> * will we know just by manual smoke-testing if it turns out that something was still relying on those rules? (versus potentially not finding out for a while)
As far as I'm aware, there's nothing else being served on `assets.publishing.service.gov.uk` (which these rules relate to) by Whitehall Frontend.  Maybe we could leave this change in place on integration for a few hours, then check Logit to see if any requests have been made to that host?

brucebolt (approved):

PR #1235, opened 2023-07-19
Remove govuk-dependency-check in non-prod.

sengi:
This doesn't run in non-prod (there's a condition in charts/govuk-jobs/dependabot-metrics-cronjob.yaml), so remove the references to it.


nsabri1 (approved):
ok

PR #1236, opened 2023-07-24
Retire LicenceFinder app

KludgeKML:
Licence finder has been replaced by the Find a Licence specialist finder, so remove from K8s configuration.

https://trello.com/c/1VXQW5I1/1620-epic-retire-original-licence-finder-application


sengi (approved):
Awesome! Thanks for cleaning up all the references to it!

I'll leave you to check it's not receiving any real traffic in prod, if you haven't already.

sengi:
I double-checked anyway and it looks good, has received zero requests in prod since 2023-07-19 14:45 👍

PR #1238, opened 2023-07-27
Integration assets ingress needs to be public.

sengi:
Same as production already is. It's only staging that's supposed to be GDS-only.

The reason it needs to be publicly reachable is the long-standing oddball usage of the integration environment for training users of the content management system (which doesn't necessarily take place at GDS).

This is acceptable for now (at least until we can run the end-to-end stack in minikube/kind/k3s), because draft assets are still behind auth-proxy and there should never be anything sensitive in integration anyway (unlike staging, which is allowed to contain pre-publication content from production). Also because this is just restoring the same access that people had before we moved to Kubernetes, when we mistakenly gave the assets LB the wrong firewall ruleset.


ollietreend (approved):
Looks good to me – thanks for fixing this @sengi 👍🏻

sengi:
(force-push was just updating the commit message)

PR #1239, opened 2023-08-02
Email Alert Frontend no longer needs Publishing API access

KludgeKML:
Email Alert Frontend no longer has a publishing_rake task which needs publishing API access, so remove the key.

https://trello.com/c/hJVUjqbs/31-frontend-apps-publish-content-items-to-publishing-api


sengi (approved):
Nice one!

PR #1241, opened 2023-08-03
Let whitehall-admin use memcache.

sengi:
I suspect this will help with the 0.1% 504 ratio on whitehall-admin. At the very least, it'll clear up a bunch of errors about `ECONNREFUSED: localhost:11211`. Might also make whitehall-admin a bit snappier.

https://govuk.zendesk.com/agent/tickets/5434688

This memcache has < 2% peak CPU utilisation, < 25% peak network utilisation and > 14 GiB of available RAM, so adding whitehall-admin should be small beer.

The name is somewhat misleading: it's frontend in the normal sense rather than the oddball GOV.UK sense, i.e. it's a memcache for apps that have UIs, not just apps that serve pages to the general public. Other GOV.UK "backend" apps (i.e. frontends for content management tools) use this memcache already.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -3025,6 +3025,8 @@ govukApplications:
           value: *publishing-notify-template
         - name: EMAIL_ADDRESS_OVERRIDE
           value: whitehall-emails-integration@digital.cabinet-office.gov.uk
+        - name: MEMCACHE_SERVERS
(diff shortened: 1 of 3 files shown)


sengi:
surprise update: whitehall-admin is still a slowbro (but it's no longer like "dude where's my memcache" so we've got that going for us, which is nice)

PR #1242, opened 2023-08-03
Remove config item for disused Signon admin API.

sengi:
All references to this env var in Signon were removed in https://github.com/alphagov/signon/pull/2294


floehopper (approved):
Good catch - thanks for clearing this up! 👍

sengi:
No probs!

sengi:
Found an obscure doc which referred to this too.

I'll merge this once Signon's finished rolling out to prod.

PR #1243, opened 2023-08-08
Add Google Tag Manager env variable for Whitehall

dnkrj:
In advance of GA4 we're adding Tag Manager to whitehall: https://github.com/alphagov/whitehall/pull/8046

In keeping with Content Publisher the application code looks for an environment variable for the id - which this PR provides.

https://trello.com/c/EDrVRCsv/1515-set-up-ga4-property-and-gtm-container

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2877,6 +2877,8 @@ govukApplications:
             secretKeyRef:
               name: signon-app-whitehall
               key: oauth_secret
+        - name: GOOGLE_TAG_MANAGER_ID
+          value: GTM-P93SHJ4Z
         - name: GOVUK_UPLOADS_ROOT
           value: &whitehall-uploads-path /uploads
         - name: ASSET_MANAGER_BEARER_TOKEN


sengi (approved):

charts/app-config/values-production.yaml
```diff
@@ -2877,6 +2877,8 @@ govukApplications:
             secretKeyRef:
               name: signon-app-whitehall
               key: oauth_secret
+        - name: GOOGLE_TAG_MANAGER_ID
+          value: GTM-NZG8SF2
```
ollietreend, line 2881: This is the wrong GTM ID. According to the [Trello card](https://trello.com/c/EDrVRCsv/1515-set-up-ga4-property-and-gtm-container) and the [original commit in this Whitehall PR](https://github.com/alphagov/whitehall/commit/c9f8e13ce428b624c2175e417d0d731ffb557813), it should be `GTM-P93SHJ4Z`.

It looks like this one has been copied from the [config for content-data-admin](https://github.com/alphagov/govuk-helm-charts/blob/b36c636e740205dac070909df83465e158baaa99/charts/app-config/values-production.yaml#L457).

ollietreend (changes requested):

charts/app-config/values-production.yaml
ollietreend: I've just amended the commit and force pushed. Resolved.

ollietreend (approved):

sengi:
You might wanna add the same env var (maybe with a different web property ID) on staging, to make it easier for you to test configuration changes.

ollietreend:
Thanks @sengi 🙂 we'll add another Google Tag Manager ID if/when needed. I'm not sure what our workflow for testing GTM changes is yet, but for now I'm going to assume [GTM's built in preview functionality](https://support.google.com/tagmanager/answer/6107056?hl=en) is our performance analyst's preferred approach.

PR #1246, opened 2023-08-09
Add shield advanced protection to EKS ALBs

sihugh:
We had this enabled on the load balancers for our EC2 instances, and it's worth bringing across. There's a few benefits, including proactive engagement from the AWS SRT team and better monitoring options in CloudWatch.

We already pay for the subscription as part of the org, so it's free to us at this point.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -35,6 +35,7 @@ _alb-ingress-defaults: &alb-ingress-defaults
     access_logs.s3.enabled=true,
     access_logs.s3.bucket=govuk-{{ .Values.govukEnvironment }}-aws-logging,
     access_logs.s3.prefix=elb/{{ .Release.Name }}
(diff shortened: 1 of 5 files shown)


sengi (approved):
Oh wow, none of us had clocked that this requires opt-in for each individual load balancer! Good catch, thanks!

sengi:
We can (and probably should) use a [Firewall Manager policy](https://docs.aws.amazon.com/waf/latest/developerguide/shield-policies.html) to configure these account-wide instead of doing it per-LB, so that we have a single point of control and don't need to rely on adding that annotation everywhere in future.

theseanything:
@sihugh let us know if you need any help getting this in

sengi:
I totally missed that it was just a (data)type issue, not anything fancy with the schemas.

sengi:
Thanks for the PR! 🙇

PR #1250, opened 2023-08-16
Try a different format for specifying arguments, to fix "pg_dump invalid output format -F c specified" errors

aldavidson:
The `pg_dump` section of the overnight export-content-store-mongo-to-postresql CronJob is failing with a strange error:

`pg_dump error invalid output format c specified`

This is odd, because the github-cli image installs pg_dump v15 (latest), and the -F c flag is [explicitly supported](https://www.postgresql.org/docs/15/app-pgdump.html)`

This PR tries a different format for the job command arguments, to try and fix the error.
  
[Trello card](https://trello.com/c/lRyopges/801-fix-invalid-output-format-error-in-the-overnight-dump-of-content-store-postgresql-to-s3)

### DIFF
--- a/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
+++ b/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
@@ -174,9 +174,9 @@ spec:
               image: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/github-cli:latest
               command: ["pg_dump"]
               args:
-                - "-F c"
-                - "-f /pg-dump/$(date +%Y-%m-%dT%H:%M:%S)-content_store_production"
-                - "${DATABASE_URL}"
+                - --format=custom
+                - --file=/pg-dump/$(date +%Y-%m-%dT%H:%M:%S)-content_store_production
+                - '${DATABASE_URL}'
               env:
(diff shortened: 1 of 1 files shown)


charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -174,9 +174,9 @@ spec:
               image: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/github-cli:latest
               command: ["pg_dump"]
               args:
-                - "-F c"
```
richardTowers, line 177: I suspect it wanted:

```
- "-F"
-  c
- "-f"
... etc
```

I also don't think the `$(date ...)` expansion is going to happen, because this array-style argument format makes me think there's not going to be a shell involved.

You might need to consider doing:

```
command: ["bash"]
args:
  - "-c"
  - "pgdump -F c -f /pg-dump/.... ...." 
  ```

aldavidson: Thanks, I'll try that next (assuming this one doesn't work :) )

aldavidson: Tried it in #1253 1253

PR #1253, opened 2023-08-16
Execute pg_dump via bash, for content-store postgresql dump CronJob

aldavidson:
Previous attempts at invoking `pg_dump` with the correct parameters have all failed with pg_dump errors on the arguments supplied (#1180 , #1249, #1250).

This attempt implements [@richardTowers' suggestion](https://github.com/alphagov/govuk-helm-charts/pull/1250#discussion_r1296126215) to invoke it with a `bash -c` 

[Trello card](https://trello.com/c/lRyopges/801-fix-invalid-output-format-error-in-the-overnight-dump-of-content-store-postgresql-to-s3)

### DIFF
--- a/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
+++ b/charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
@@ -174,11 +174,10 @@ spec:
                   mountPath: /pg-dump
             - name: dump-postgresql
               image: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/github-cli:latest
-              command: ["pg_dump"]
+              command: ["bash"]
               args:
-                - --format=custom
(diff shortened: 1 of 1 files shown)


charts/govuk-jobs/templates/content-store-mongo-to-postgres-cronjob.yaml
```diff
@@ -174,11 +174,10 @@ spec:
                   mountPath: /pg-dump
             - name: dump-postgresql
               image: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/github-cli:latest
-              command: ["pg_dump"]
+              command: ["bash"]
```
richardTowers, line 177: Only other thing I wanted to check was that `bash` is actually the shell in the image - `sh` might have been more portable if not. But [it is bash](https://github.com/alphagov/govuk-infrastructure/blob/80331f8455453414c311740376bfb1c330a079dd/images/github-cli/Dockerfile#L1) so all good.

richardTowers (approved):

PR #1258, opened 2023-08-24
Add image-tag file for new content-store-postgresql-branch app in staging

aldavidson:
The newly-deployed `content-store-postgresql-branch` app in staging can't actually start running, due to an `invalid image name` error in Argo.

I think this is because there is no corresponding file in `charts/image-tags/staging/` - so this PR adds one.

[Trello card](https://trello.com/c/vfvX0CLX/813-add-cronjobs-to-overnight-populate-content-store-postgresql-in-staging-production)


sengi (approved):
Ah yeah — feel free to add the file for prod too if you like, no harm in doing that ahead of time.

PR #1257, opened 2023-08-24
Add content-store-postgresql-branch (& draft-) apps in staging

aldavidson:
[Trello card](https://trello.com/c/vfvX0CLX/813-add-cronjobs-to-overnight-populate-content-store-postgresql-in-staging-production) - completes [step 1 of the rollout plan](https://docs.google.com/presentation/d/1BMvv71helDENeaBo23gDHjoN74A9lqtGU5NDgQiUab4/edit#slide=id.g2762d4e8b40_0_36). 
Allows us to start populating the PostgreSQL databases in staging before introducing the proxy app in front of both PostgreSQL & Mongo versions (a subsequent PR will do that separately).

PR #1260, opened 2023-08-24
Add content-store-postgresql-branch apps & cronjobs in production (for live and draft)

aldavidson:
[Trello card](https://trello.com/c/vfvX0CLX/813-add-cronjobs-to-overnight-populate-content-store-postgresql-in-staging-production)

This PR adds the `...-postgresql-branch` applications and mongo-to-postgres cronjobs, but does not introduce the proxy - just like the previous PRs for staging (#1257, #1258 and #1259), but a) for production, and b) adding the image-tag & cronjobs at the same time.

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -690,6 +690,48 @@ govukApplications:
         - name: WEB_CONCURRENCY
           value: '4'
 
+  - name: content-store-postgresql-branch
+    repoName: content-store-postgresql-branch
+    helmValues:
+      <<: *content-store
+      rails:
+        createKeyBaseSecret: false
+        # use the same secret as the mongo content-store, it will make the
+        # eventual switchover easier and reduce toil
+        secretKeyBaseName: content-store-rails-secret-key-base
+      sentry:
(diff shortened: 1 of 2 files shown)


charts/app-config/values-production.yaml
```diff
@@ -726,6 +768,48 @@ govukApplications:
...
+        - name: DEFAULT_TTL
+          value: "1"
+        - name: PLEK_HOSTNAME_PREFIX
+          value: draft-
+        - name: DATABASE_URL
+          valueFrom:
+            secretKeyRef:
+              name: draft-content-store-postgres
+              key: DATABASE_URL
+        - name: DISABLE_ROUTER_API
```
sengi, line 810: Presumably this has to be removed at cutover?

sengi (approved):

charts/app-config/values-production.yaml
aldavidson: Yeah, this setting is to prevent both content stores hitting router-api when they get updates.

At the point we switch primary & secondary on the proxy, we should swap this setting to the Mongo app, and at the point we get rid of the Mongo app entirely, this setting should be removed

PR #1268, opened 2023-09-09
Add Postgres backup/restore jobs for content-store.

sengi:
This is designed to replace [govuk_env_sync] in the very near future.

For now, we just configure the jobs for the new (not yet serving user traffic in production) Postgres-based version of content-store.

[govuk_env_sync]: https://github.com/alphagov/govuk-puppet/tree/main/modules/govuk_env_sync

https://trello.com/c/JjTHY62V

### DIFF
--- a/charts/db-backup/values.yaml
+++ b/charts/db-backup/values.yaml
@@ -22,26 +22,26 @@ govukEnvironment: staging
 cronjobs:
   production:
     content-store-postgres:
-      schedule: "16 1 * * *"
+      schedule: "16 2 * * *"
       db: content_store_production
       operations:
         - op: backup
     draft-content-store-postgres:
-      schedule: "16 1 * * *"
+      schedule: "36 2 * * *"
       db: draft_content_store_production
       operations:
         - op: backup
   staging:
     content-store-postgres:
       db: content_store_production
-      schedule: "16 2 * * *"
+      schedule: "16 3 * * *"
       operations:
         - op: restore
           bucket: s3://govuk-production-database-backups
         - op: backup
     draft-content-store-postgres:
       db: draft_content_store_production
-      schedule: "16 2 * * *"
+      schedule: "36 3 * * *"
       operations:
         - op: restore
           bucket: s3://govuk-production-database-backups
@@ -49,7 +49,7 @@ cronjobs:
   integration:
     content-store-postgres:
       db: content_store_production
-      schedule: "16 3 * * *"
+      schedule: "16 4 * * *"
       operations:
         - op: restore
           bucket: s3://govuk-staging-database-backups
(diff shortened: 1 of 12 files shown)


aldavidson (approved):
Looks good to me. Just one thing springs to mind about the [filtering that we do in the old env_sync job](https://github.com/alphagov/govuk-puppet/blob/3e242fd3476e229d7d3739db424f6aea4d2ed835/modules/govuk_env_sync/files/govuk_env_sync.sh#L389-L395) - I don't see any equivalent here, is that going to be a separate change?

sengi:
Yeah that'll be the next step.

sengi:
I also need to restructure the config a bit so that we're not tied to having at most one job per database host/cluster (so that e.g. we don't have to wait to finish dumping content_store_production before starting to dump draft_content_store_production or vice versa). That's an easy change though.

charts/db-backup/values.yaml
```diff
@@ -0,0 +1,98 @@
...
+        - op: restore
+          bucket: s3://govuk-staging-database-backups
+    # TODO: stop copying draft content into integration once the design quirks
+    # of publishing-api that require it are rectified.
+    draft-content-store-postgres:
+      db: draft_content_store_production
+      schedule: "7 4 * * *"
+      operations:
+        - op: restore
+          bucket: s3://govuk-staging-database-backups
```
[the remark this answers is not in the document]
sengi: It's an expensive way to do it, but we could. I think there might be better ways though. If you can point me at their config I can take a look and come up with a good solution.

sengi:
RFAL. Timings fixed up so data should flow from prod mongo->prod psql->staging->integration without getting held up for a day, and permissions should work now so staging can read from the prod bucket etc.

charts/db-backup/values.yaml
```diff
@@ -0,0 +1,100 @@
...
+#
+# Each environment has a map of job names to job configs. If the job name (i.e.
+# the map key) is of the form <app-name>-<dbms>, for example
+# link-checker-api-postgres, then the DBMS hostname and DBMS type (postgres,
+# mysql etc.) can be inferred. These can be specified explicitly via the `host`
+# and `dbms` keys in the job config.
+cronjobs:
+  production:
+    content-store-postgres:
+      schedule: "16 1 * * *"
```
aldavidson, line 25: I think this timing could be a bit too tight - if I'm reading it correctly, the mongo-to-postgres cron job starts at 00:16, and the backup of content-store-postgres starts an hour later at 01:16 . But the mongo-to-postgres cron job _currently_ takes about 1hr 15mins to run - [example from prod](https://argo.eks.production.govuk.digital/applications/govuk-jobs?orphaned=false&resource=&node=%2FPod%2Fapps%2Fcontent-store-mongo-to-postgres-cron-28274790-gsdfk&tab=summary):  

```
          finishedAt: '2023-10-05T07:56:31Z'
          reason: Completed
          startedAt: '2023-10-05T06:41:22Z'
```

sengi: Ah good catch, thanks!

sengi:
Rebase to avoid merge conflicts in `app-config/values-*.yaml`.

PR #1276, opened 2023-09-12
Don't time out whitehall-admin requests prematurely.

sengi:
Some whitehall-admin requests are slow; let's not make it worse by giving up in the reverse proxy before the application does.


ollietreend (approved):
Looks good to me! Thanks for doing this @sengi.

PR #1281, opened 2023-09-14
Set `workerEnabled` for `search-api-v2`

csutter:
Turns out simply setting `workers` isn't enough as this value is false by default!


theseanything (approved):

mlk-kc (approved):
:shipit:

PR #1290, opened 2023-09-26
Delete existing branch before updating image tag

theseanything:
This prevents failures on update image tag workflow re-runs. This is because subsequent runs attempt to re-use the existing branch which can be out of date, cause subsquent pushed to be rejected.

We should just always start from a clean branch, as we don't care what previous workflow runs have done.

### DIFF
--- a/charts/argo-services/scripts/update-image-tag.sh
+++ b/charts/argo-services/scripts/update-image-tag.sh
@@ -20,6 +20,12 @@ if [[ "${current_image_tag}" = "${IMAGE_TAG}" ]]; then
   exit 0
 fi
 
+# Delete branch at origin if already exists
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Still feels like it could be race condition-y

(e.g. if more than one of these workflows is running simultaneously, one could do `git push -u origin "${BRANCH}" and another could do `git push origin --delete "${BRANCH}"` before the first workflow manages to do the merge.

I think some level of raciness is inevitable though. At least with this approach it shouldn't get stuck in a broken state.

theseanything:
Yeah agree there's still the race condition, but at least you can retry the work and have it succeed. Its way way less likely that the race condition is between workflows deploying the same app/version/environment.

PR #1302, opened 2023-10-04
Update request rates, errors, durations dashboard

richardTowers:
There are a few changes here, which I've tried to describe in individual commits to make it easier to review.

The request rates, errors, durations dashboard has numerous issues. The most immediately serious one is that the metric it's using to work out which apps exist no longer includes most apps (because of the switch to histograms in https://github.com/alphagov/govuk_app_config/pull/318), but there were also a few issues stemming from the confusing nature of prometheus summaries.

I've also taken the opportunity to switch the request durations visualisation to use a heatmap, which I think is a more useful representation of the information. 

| Before | After (note: different time span) |
|-----|-----|
| image | image |

You can play with the new version of the dashboard here: https://grafana.eks.production.govuk.digital/d/cdeb0607-02e2-4c51-8b7a-398a1c1e2936/richard-towersat-app-request-rates-errors-durations

### DIFF
--- a/charts/monitoring-config/dashboards/app-requests.json
+++ b/charts/monitoring-config/dashboards/app-requests.json
@@ -56,6 +56,7 @@
               "tooltip": false,
               "viz": false
             },
+            "insertNulls": false,
             "lineInterpolation": "linear",
             "lineWidth": 1,
             "pointSize": 5,
@@ -186,6 +187,7 @@
               "tooltip": false,
               "viz": false
             },
+            "insertNulls": false,
             "lineInterpolation": "linear",
             "lineWidth": 1,
             "pointSize": 4,
@@ -346,6 +348,7 @@
               "tooltip": false,
               "viz": false
             },
+            "insertNulls": false,
             "lineInterpolation": "linear",
             "lineWidth": 1,
             "pointSize": 5,
@@ -512,7 +515,7 @@
           "reverse": false
         }
       },
-      "pluginVersion": "9.5.5",
+      "pluginVersion": "10.1.5",
       "targets": [
         {
           "datasource": {
--- a/charts/monitoring-config/dashboards/app-requests.json
+++ b/charts/monitoring-config/dashboards/app-requests.json
@@ -603,51 +603,20 @@
           "type": "prometheus",
           "uid": "prometheus"
         },
-        "definition": "label_values(http_request_duration_seconds{namespace=\"${namespace}\"}, job)",
+        "definition": "label_values(http_requests_total{namespace=\"${namespace}\"},job)",
         "hide": 0,
         "includeAll": true,
         "multi": true,
         "name": "app",
         "options": [],
         "query": {
-          "query": "label_values(http_request_duration_seconds{namespace=\"${namespace}\"}, job)",
+          "query": "label_values(http_requests_total{namespace=\"${namespace}\"},job)",
           "refId": "StandardVariableQuery"
         },
         "refresh": 2,
         "regex": "",
         "skipUrlSync": false,
         "sort": 1,
-        "type": "query"
-      },
-      {
-        "current": {
-          "selected": true,
-          "text": [
-            "All"
-          ],
-          "value": [
-            "$__all"
-          ]
-        },
-        "datasource": {
-          "type": "prometheus",
-          "uid": "prometheus"
-        },
-        "definition": "label_values(http_request_duration_seconds{namespace=\"${namespace}\"}, quantile)",
-        "hide": 0,
-        "includeAll": true,
-        "label": "quantile",
-        "multi": true,
-        "name": "quantile",
-        "options": [],
-        "query": {
-          "query": "label_values(http_request_duration_seconds{namespace=\"${namespace}\"}, quantile)",
-          "refId": "StandardVariableQuery"
-        },
-        "refresh": 2,
-        "regex": "",
-        "skipUrlSync": false,
-        "sort": 4,
         "type": "query"
       },
       {
--- a/charts/monitoring-config/dashboards/app-requests.json
+++ b/charts/monitoring-config/dashboards/app-requests.json
@@ -555,7 +555,7 @@
     }
   ],
   "refresh": "1m",
-  "schemaVersion": 37,
+  "schemaVersion": 38,
   "style": "dark",
   "tags": [],
   "templating": {
--- a/charts/monitoring-config/dashboards/app-requests.json
+++ b/charts/monitoring-config/dashboards/app-requests.json
@@ -429,7 +429,7 @@
           },
           "editorMode": "code",
           "exemplar": true,
-          "expr": "sum by (job) (increase(http_request_duration_seconds_sum{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval])) / sum by (job) (increase(http_request_duration_seconds_count{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval]))",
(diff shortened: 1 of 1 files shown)


richardTowers:
I have other thoughts about this dashboard, but I didn't want to go too far down the rabbit hole.

Jotting a few of them down here:
- It's probably confusing bundling up request counts / durations across all the publishing apps, frontends and APIs - is it really sensible to have one panel (or even one dashboard) trying to cover all of these things? It would be cool if we could add a label to the metrics which categorised which _kind_ of thing something was (e.g. whitehall is a publishing app, publishing-api is an internal API, frontend is a frontend) to help group these things onto more focused dashboards
- Full width graphs means we're not using vertical space very effectively. My beautiful heat map is below the fold on my screen 😭. I think it would look better as a 2 x 2 - requests and errors on the top row, average durations and heatmap on the bottom.
- HTTP Errors is not very useful without a filter on, because it's absolutely dominated by 404s (most of which are totally legit responses, not errors at all)
- It would be nice to link out to other dashboards where you could drill down into more detail. E.g. the heatmaps are much more useful if you [drill down to a single controller](https://grafana.eks.production.govuk.digital/d/a4c489b1-1346-469c-8727-4715cbdec833/richard-towersat-heatmap-experiments?orgId=1&var-namespace=apps&var-app=whitehall-admin&var-controller=admin%2Fdetailed_guides&var-action=All)

charts/monitoring-config/dashboards/app-requests.json
```diff
@@ -426,7 +429,7 @@
           },
           "editorMode": "code",
           "exemplar": true,
-          "expr": "avg(http_request_duration_seconds{namespace=\"${namespace}\", job=~\"${app}\",quantile=~\"${quantile}\"}) by (job)",
+          "expr": "sum by (job) (increase(http_request_duration_seconds_sum{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval])) / sum by (job) (increase(http_request_duration_seconds_count{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval]))",
```
sengi, line 432: `increase` is unnecessary here and obscures the intent. (`increase` is just `rate(...) * $__rate_interval` so it's a redundant factor hiding inside some unusual syntactic sugar.)

```suggestion
          "expr": "sum by (job) (rate(http_request_duration_seconds_sum{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval])) / sum by (job) (rate(http_request_duration_seconds_count{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval]))",
```

charts/monitoring-config/dashboards/app-requests.json
```diff
@@ -438,7 +441,7 @@
             "uid": "prometheus"
           },
           "editorMode": "code",
-          "expr": "avg(http_request_duration_seconds{namespace=\"${namespace}\", quantile=~\"${quantile}\"})",
+          "expr": "sum (increase(http_request_duration_seconds_sum{namespace=\"${namespace}\"}[$__rate_interval])) / sum (increase(http_request_duration_seconds_count{namespace=\"${namespace}\"}[$__rate_interval]))",
```
sengi, line 444: as above (only duplicating for convenience in case you wanna use the suggested changes feature)
```suggestion
          "expr": "sum (rate(http_request_duration_seconds_sum{namespace=\"${namespace}\"}[$__rate_interval])) / sum (rate(http_request_duration_seconds_count{namespace=\"${namespace}\"}[$__rate_interval]))",
```

charts/monitoring-config/dashboards/app-requests.json
```diff
@@ -453,109 +456,82 @@
...
           "datasource": {
             "type": "prometheus",
             "uid": "prometheus"
           },
           "editorMode": "code",
-          "exemplar": true,
-          "expr": "label_replace(avg without (pod, instance) (http_request_duration_seconds{namespace=\"${namespace}\", job=~\"${app}\", quantile=~\"${quantile}\"}), \"a\", \"$1\", \"job\", \"(.*)\")",
-          "interval": "",
-          "legendFormat": "{{job}} {{controller}} {{action}} {{quantile}}",
+          "expr": "sum by (le) (increase(http_request_duration_seconds_bucket{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval]))",
```
sengi, line 522: In this one, `increase` actually gives us something other than what we want. This is because `increase` gives the difference in value between the first and last non-null samples of its argument (range vector) whereas we need the per-second rate. In other words, we expect our output to be independent of the chosen sliding window (rate interval) but we were erroneously multiplying each output point by the number of seconds in the rate interval.

You can see this e.g. if you plot just one bucket (say `le="+Inf"` i.e. cumulative total request count) for some particular job as a timeseries and compare the values to the request rate graph.
```suggestion
          "expr": "sum by (le) (rate(http_request_duration_seconds_bucket{namespace=\"${namespace}\", job=~\"${app}\"}[$__rate_interval]))",
```

sengi (approved):
Cool, thanks for fixing this broken dashboard! Been on my to-do list for a while!

charts/monitoring-config/dashboards/app-requests.json
sengi: Oh wait, I think I've got my wires crossed here — our output points should be durations not counts of requests.

sengi: Yeah no I think I was thinking ok in the first post. Do check my working though! 😅

charts/monitoring-config/dashboards/app-requests.json
richardTowers: Haha, I went through this thought process - in my mind `rate` is just `increase(...) / $__rate_interval`, so by pretty much the exact same logic I made the opposite decision.

Happy to switch to rate though, since they're equivalent and rate is probably more idiomatic (and is fewer characters to type).

charts/monitoring-config/dashboards/app-requests.json
richardTowers: I think I could argue it both ways... It's these numbers we're talking about (sloppily underlined in red) I think:

image

- `increase` gives "the absolute number of requests in this bit of the heatmap"
- `rate` gives "the number of requests per second in this bit of the heatmap"

Because it's a heatmap, rather than a line graph, the size of the bucket is observable in the visualisation, so `increase` still makes sense in context, I think. That said, I think it's highly debateable which is more intuitive.... 🤔

I think the argument for using `increase` is strongest for the buckets with a small number of samples:

image

i.e. (with increase) 

> there were 4 requests between 03:02 and 03:03 which had a duration of between 5 and 10 seconds.

vs. (with rate)

> there were 0.067 requests per second between 03:02 and 03:03 which had a duration of between 5 and 10 seconds.

I think the argument for using `rate` is that it's consistent with other visualisations on the dashboard.

I'm honestly not sure which is better. Interested in your thoughts?

richardTowers: One other thing to note: even with `increase`, the values are not always integers, because `$__rate_interval` is not always an even multiplier of the sample interval. So you do get confusing datapoints like "there were 2.67 requests between 03:02 and 03:03".

richardTowers: Screenshot of it with rate with units added (which do help clear it up a bit):

image

richardTowers: Or with increase, but with the value in the tooltip labeled so it's more clear what it means:

image

sengi: Ah yeah that's a fair point; I think if we're labelling it clearly like in your last screenshot then it's fine.

PR #1310, opened 2023-10-11
Stop using transition-config-cron in integration

mike29736:
https://trello.com/c/8lQCsx3y/862-identify-all-things-related-to-transition-config-elsewhere-in-alphagov-and-update-or-remove-the-references

Stop using transition-config-cron in integration and instead, replace
it with a standard cron task for the Transition app.

If I've understood transition-config-cron correctly, it makes data from
the transition-config repository available to the Transition app and
then invokes one of the Transition app's Rake tasks.

The Rake task that it invokes:

1. already lives in the Transition app, and
2. no longer requires data from the transition-config repository (since
   https://github.com/alphagov/govuk-helm-charts/pull/1296)

I'm having trouble fully understanding the transition-config-cron file,
so I'm testing the waters first by limiting this change to integration.
If it works as expected, I'll then make this change for production and
staging and remove transition-config-cron altogether.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2816,8 +2816,10 @@ govukApplications:
           - name: transition.{{ .Values.k8sExternalDomainSuffix }}
       dbMigrationEnabled: true
       workerEnabled: true
-      transitionConfigEnabled: true
       cronTasks:
+        - name: import-all-organisations
+          task: "import:all:organisations"
+          schedule: "00 09-19 * * 1-5"
         - name: import-dns
           task: "import:dns_details"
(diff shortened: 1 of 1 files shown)


sengi (approved):
How have you tested this?

If you're not confident that you understand the problem space then I'd recommend not rolling it out to all three environments at the same time.

mike29736:
> How have you tested this?
> 
> If you're not confident that you understand the problem space then I'd recommend not rolling it out to all three environments at the same time.

Thanks @sengi,

I haven't tested this in any way so far.

I'll edit this pull request so that it only makes the change for one environment, if that's a reasonable way to do it

sengi (approved):

PR #1311, opened 2023-10-12
Move transition-config cron job to Transition app

mike29736:
https://trello.com/c/8lQCsx3y/862-identify-all-things-related-to-transition-config-elsewhere-in-alphagov-and-update-or-remove-the-references

* Remove custom transition-config-cron and
* replace it with a standard cron task in the Transition app's config.

transition-config-cron made data from the transition-config repository available to the Transition app and then invoked one of the Transition app's Rake tasks.

The Rake task that it was invoking:

1. already lives in the Transition app, and
2. no longer requires data from the transition-config repository (since https://github.com/alphagov/govuk-helm-charts/pull/1296).

This change is already live in integration
(https://github.com/alphagov/govuk-helm-charts/pull/1310), now this makes the same change for staging and production, too.


sengi (approved):
Awesome, thanks for keeping the config tidy!

PR #1314, opened 2023-10-12
Add runbook to monitoring alert annotations

KludgeKML:
Adds runbook annotation for signon tokens alert pointing to the relevant developer docs. Actually consumed by https://github.com/alphagov/govuk-infrastructure/pull/961

### DIFF
--- a/charts/monitoring-config/rules/signon.yaml
+++ b/charts/monitoring-config/rules/signon.yaml
@@ -12,3 +12,4 @@
         summary: Signon API User token is due to expire soon
         description: >-
           A token is about to expire within the next two weeks and needs rotating.
+        runbook_url: https://docs.publishing.service.gov.uk/manual/alerts/signon-api-user-token-expires-soon.html


charts/monitoring-config/rules/signon.yaml
```diff
@@ -12,3 +12,4 @@
         summary: Signon API User token is due to expire soon
         description: >-
           A token is about to expire within the next two weeks and needs rotating.
+        runbook: https://docs.publishing.service.gov.uk/manual/alerts/signon-api-user-token-expires-soon.html
```
theseanything, line 15: Nit: runbook_url is more explicit as what the field is

theseanything (approved):

charts/monitoring-config/rules/signon.yaml
KludgeKML: Oh yeah, good idea.

PR #1323, opened 2023-10-17
Increase the memory and cpu limits of govuk-mirror

theseanything:
This is to prevent the scraping container from being OOMKilled and throttled.

### DIFF
--- a/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
+++ b/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
@@ -45,11 +45,11 @@ spec:
               workingDir: /data
               resources:
                 limits:
-                  cpu: 2
-                  memory: 8000Mi
+                  cpu: 4
+                  memory: 10000Mi
                 requests:
-                  cpu: 1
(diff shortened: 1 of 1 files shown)


charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
```diff
@@ -45,11 +45,11 @@ spec:
...
                 limits:
-                  cpu: 2
-                  memory: 8000Mi
+                  cpu: 4
+                  memory: 10000Mi
                 requests:
-                  cpu: 1
-                  memory: 2000Mi
+                  cpu: 2
+                  memory: 5000Mi
```
sengi, line 52: Wow. Room for optimisation at some point I guess! (I agree with just chucking resources at it for now.)

sengi (approved):

charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
theseanything: Yeah 😢  - need to have another look at why it's gobbling so much up.

PR #1327, opened 2023-10-17
Avoid creating broken cronjobs in govuk-jobs chart.

sengi:
Don't create the (draft)contentStoreMongoPostgresCron cronjobs unless they have a schedule configured. Fixes a minor bug introduced in #1268.

Also an inconsequential fix for the schedule on contentStoreMongoPostgresCron (it was using the schedule for the draft one, but they're set the same anyway at the moment).

https://trello.com/c/JjTHY62V


aldavidson (approved):
:+1:

PR #1329, opened 2023-10-18
Enable mirroring assets domain

theseanything:
This enables the crawler to also download assets on the asset domain and upload them to S3.

Tested in staging.

### DIFF
--- a/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
+++ b/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
@@ -62,7 +62,7 @@ spec:
                 - name: SITE
                   value: https://www.{{ .Values.externalDomainSuffix }}/sitemap.xml
                 - name: ALLOWED_DOMAINS
-                  value: www.{{ .Values.externalDomainSuffix }}
+                  value: www.{{ .Values.externalDomainSuffix }},{{ .Values.assetsDomain }}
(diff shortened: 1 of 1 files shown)


charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
```diff
@@ -106,5 +106,6 @@ spec:
                 - /bin/sh
                 - -c
                 - >-
-                  /s5cmd sync --size-only "/data/${WWW_DOMAIN}" "s3://${BUCKET_NAME}/"
+                  /s5cmd sync --size-only "/data/${WWW_DOMAIN}" "s3://${BUCKET_NAME}/";
```
sengi, line 109: optional: could make this `&&` instead of `;` so that the job fails conspicuously if the first command fails but the second succeeds. (Not saying that's necessarily the right tradeoff — up to you — just mentioning in case you hadn't already considered it.)

sengi (approved):
Ooh s5cmd, shiny!

charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
theseanything: Ah good catch - I was thinking to this allows it to attempt to sync as much stuff as possible. But I suppose if the first command fails and second succeeds then the job won't bubble up the failure.

PR #1331, opened 2023-10-18
Fix upload to S3 command in govuk-mirror-sync

theseanything:
The new line gets stripped causing the \ to escape the " " (space), which is then intepreted as part of the command string.


sengi (approved):
evil YAML!

PR #1332, opened 2023-10-18
Add missing IAM Role value for govuk-mirror-sync in prod

theseanything:
🤦🏻

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1188,6 +1188,8 @@ govukApplications:
       - "search-api"
       - "search-api-learn-to-rank"
     helmValues:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -1,4 +1,4 @@
-# Defaults for all apps in this environment. These can be overridden for
+#831600 Defaults for all apps in this environment. These can be overridden for
```
sengi, line 1: accidental paste? :)

sengi (approved):

charts/app-config/values-production.yaml
theseanything: Yuppp :/

PR #1333, opened 2023-10-18
Ensure `search-api-v2` Google secrets get decoded

csutter:
These are base64 encoded at the source, and need to be decoded through the operator.


csutter:
This fails validation because the schema is outdated. To unblock ourselves, closing this in favour of base64 decoding at the source instead: https://github.com/alphagov/search-v2-infrastructure/pull/28

PR #1350, opened 2023-10-27
Update Google Cloud configuration for `search-api-v2`

csutter:
- Remove read/write credentials separation (`search-api-v2` will stay a monolithic app for the time being)
- Add new ExternalSecret for Discovery Engine configuration (`DISCOVERY_ENGINE_DATASTORE` and `DISCOVERY_ENGINE_ENGINE`)
- Add new credentials and configuration to integration app environment

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2263,17 +2263,12 @@ govukApplications:
         - command: ['bin/rake', 'document_sync_worker:run']
           name: document-sync-worker
       extraVolumes:
-        - name: search-api-v2-google-key-read
+        - name: search-api-v2-google-cloud-credentials
           secret:
-            secretName: search-api-v2-google-key-read
-        - name: search-api-v2-google-key-write
-          secret:
-            secretName: search-api-v2-google-key-write
+            secretName: search-api-v2-google-cloud-credentials
       appExtraVolumeMounts:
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2282,10 +2277,18 @@ govukApplications:
...
-        - name: GOOGLE_KEYFILE_READ
-          value: /etc/keys/google-key-read.json
-        - name: GOOGLE_KEYFILE_WRITE
-          value: /etc/keys/google-key-write.json
+        - name: GOOGLE_CLOUD_CREDENTIALS
+          value: /etc/credentials/google-cloud/credentials.json
+        - name: DISCOVERY_ENGINE_DATASTORE
+          valueFrom:
+            secretKeyRef:
+              name: search-api-v2-google-cloud-discovery-engine-configuration
```
sengi, line 2285: info, no action needed: this name is close to the max length (63) but that's ok for now cos we're not prefixing it with the chart name etc.

sengi (approved):
Awesome! Thanks so much for doing this! Helps keep us secure and sets a good example for others ❤️✨

charts/app-config/values-integration.yaml
csutter: It _is_ a bit of a mouthful 😬

PR #1359, opened 2023-10-27
Add prometheus alerts for mirror freshness

theseanything:
This alerts fires when the mirrors have exceed two days since they've been updated.

### DIFF
--- a/charts/monitoring-config/rules/mirrors.yaml
+++ b/charts/monitoring-config/rules/mirrors.yaml
@@ -1,0 +1,13 @@
+- name: MirrorAlerts
+  rules:
+    - record: global:govuk_mirror_freshness_seconds
+      expr: |
+        time() - min(govuk_mirror_last_updated_time) by (backend)
+
+    - alert: MirrorFreshnessAlert
+      expr: |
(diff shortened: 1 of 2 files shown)


charts/monitoring-config/rules/mirrors.yaml
```diff
@@ -0,0 +1,14 @@
+- name: MirrorAlerts
+  rules:
+    - record: global:govuk_mirror_freshness_seconds
+      expr: |
+        time() - min(govuk_mirror_last_updated_time) by (backend)
+
+    - alert: MirrorFreshnessAlert
+      expr: |
+        govuk_mirror_freshness_seconds < 172800
+      for: 1h
```
sengi, line 10: optional: not sure the `for` is doing much here (and it's an optional field)

sengi (approved):
If you have time, could you add some tests for these please?

charts/monitoring-config/rules/mirrors.yaml
sengi: (slight doubt in the back of my mind that maybe this is working around counter resets or missing data points or something, in which case ignore me!)

PR #1360, opened 2023-10-30
Install Reloader in integration and add annotation to any app using generic-govuk-app chart

samsimpson1:
https://trello.com/c/52c99wvl/1235-solve-the-problem-of-configmap-secrets-changes-requiring-user-initiated-rollouts-which-are-inevitably-forgotten-some-of-the-time


sengi (approved):
Neat! Thanks for using Argo CD so it stays up to date!

PR #1364, opened 2023-10-30
Record 100% trace data for Whitehall integration

mike29736:
https://trello.com/c/jDSDGNyn/814-measure-and-record-our-publishing-latency-sli

This overrides[^1] the sample rate that's set for all apps in `charts/app-config/templates/env-configmap.yaml`. It controls how much data OpenTelemetry sends to Grafana Tempo.

We're building a Service Level Indicator for publishing (Whitehall->Content Store) and experimenting with traces in Tempo to figure out whether that's the right tool for the job. Since our OpenTelemetry sample rate is currently set to 0.1%, there isn't actually much data in Tempo to experiment with.

[^1]: Does it? I haven't got to the bottom of the `env`/`envFrom` precedence or spotted any existing examples, but surely these variables override the configmap's?

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -3069,6 +3069,10 @@ govukApplications:
             secretKeyRef:
               name: whitehall-admin-mysql
               key: DATABASE_URL
+        - name: OTEL_TRACES_SAMPLER
+          value: traceidratio
+        - name: OTEL_TRACES_SAMPLER_ARG
+          value: "1"
       extraVolumes:
         - name: asset-uploads-efs
           nfs:


sengi (approved):
Distributed tracing isn't really appropriate for measuring an SLI. It's ok to set this to 100% temporarily if you're debugging some tricky problem, but it's not feasible to leave it that way. The performance and storage costs will be extreme, plus it'll affect reliability and security.

sengi:
Approving so you can go ahead and play with otel, but please bear in mind it's not gonna be feasible to do this in production.

sengi:
For your SLI, consider adding a Prometheus metric to measure what you're interested in. That's by far the most common and straightforward way. Less common but sometimes necessary alternatives (depending on what you're trying to measure) include: external ("black-box") probe (synthetic but consistent, good for complex workflows), post-processing logs (good for extremely high accuracy stuff or where direct instrumentation isn't feasible).

mike29736:
Thanks, @sengi.

This isn't worth me exploring at all if it isn't going to work long-term and in production, so I appreciate you making that clear.

I'm going to leave this pull request open while I go back to the person who suggested Tempo to me to see if they have anything to add.

sengi:
Thanks, that's helpful in understanding where you're coming from. I don't think this is a good approach for an SLI though.

Thinking in terms of logs vs metrics vs traces "observables" is not the most helpful mental model here, because only two of those are relevant for designing SLIs (logs and metrics) and you might need to use either or both of those depending on what you're aiming to measure.

The scenario you've described there sounds like a good fit for a prober (either running a synthetic scenario or checking how long real changes take to appear on the site).

Totally fine if we want to enable 100% tracing on publish actions (including in prod) if there are only hundreds per day. (You'd almost certainly need to do that from application code rather than just setting the sampling to 100%, so that you're only tracing those publishing requests and not everything else that the app does.)

Traces really aren't the right tool for an SLI, especially not for a system's first SLI.

I've already approved the PR btw, so that you can try out whatever you want with OpenTelemetry traces on the app in integration.

richardTowers:
Discussed this a bit offline with Xander (PM on the team), and we're going to pursue an approach to measuring this using information from the logs instead.

I think until we're more confident with traces and their long term future on GOV.UK, it's wise to avoid building important functionality (like SLIs) on top of them.

I'm definitely up for thoughts on better ways to measure things like this though.

PR #1382, opened 2023-11-03
Add AlertManager configuration

samsimpson1:


sengi:
Approach lgtm! I just skimmed the template stuff (since a fair bit of it is preexisting cruft anyway right, like those duplicated receiver configs).

PR #1384, opened 2023-11-03
Fix location field not being supported by AlertmanagerConfig CRD

samsimpson1:


sengi (approved):
Oh no! Hopefully we can fix this upstream before spring so we don't have to faff with DST 😂

PR #1387, opened 2023-11-04
Add MySQL support to db-backup chart. 

sengi:
Support `mysqldump` backup and restore for MySQL databases. Enable the cronjobs for the MySQL databases except whitehall-mysql, which needs a (yet to be implemented) data transformation between staging and integration.

Also fix the buffering on the progress log output.

https://trello.com/c/JjTHY62V

Tested: successfully ran the jobs for `signon-mysql` end-to-end from production through to integration (and `service-manual-publisher-postgres` backup/restore in staging as a regression test).

### DIFF
--- a/charts/db-backup/backup-mysql.sh
+++ b/charts/db-backup/backup-mysql.sh
@@ -1,34 +1,14 @@
 #!/bin/bash
 set -euo pipefail
 
-export LC_ALL=C.UTF-8  # Prevent i18n from affecting command outputs (e.g. `type`).
-export HOME=${TMPDIR:-/tmp}  # For aws-cli and the mysql* tools.
-
-usage () {
-  self=$(basename "$0")
-  cat >&2 <<EOF
-$self is a tool for streaming mysqldump output to/from Amazon S3 (or
-compatible), for backup and to copy data between environments.
-EOF
-  exit 64
-}
-
-progress () {
-  stdbuf -eL -- pv -fi 10 -F '%t %r %b'
-}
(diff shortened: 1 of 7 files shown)


charts/db-backup/backup-mysql.sh
```diff
@@ -0,0 +1,52 @@
...
+}
+
+write_config () {
+  cat > "$HOME/.my.cnf" <<EOF
+[client]
+user=$DB_USER
+password=$DB_PASSWORD
+host=$DB_HOST
+[mysqldump]
+set-gtid-purged=OFF  # https://bugs.mysql.com/bug.php?id=109685#c530123
```
sengi, line 42: This was a fun one to track down 🙃

charts/db-backup/backup-mysql.sh
```diff
@@ -0,0 +1,52 @@
...
+  fi
+  local s3_url="$BUCKET/$DB_HOST/$FILENAME"
+  s3_url="$s3_url" dump_is_readable
+  mysqladmin drop -f "$DB_DATABASE" || true
+  mysqladmin create "$DB_DATABASE"
+  aws s3 cp "$s3_url" - | gzip -cd | progress | mysql "$DB_DATABASE"
+}
+
+write_config () {
+  cat > "$HOME/.my.cnf" <<EOF
```
sengi, line 36: This appears to be the least sh�t way to get creds into the `mysql*` tools that still works. Spent ages trying to get it working using env vars but `MYSQL_PWD` doesn't work consistently across all the tools.

samsimpson1 (approved):

PR #1391, opened 2023-11-06
Install kube-prometheus-stack via Argo in production

samsimpson1:


sengi (approved):
Not sure what's up with the kubeconform schema but in case it's related, I think we're running a slightly old version of external-secrets-operator so might be some differences in its CRDs vs the current version. (hopefully the kubeconform schemas aren't actually out of sync with what we're running, but that's also possible!)

sengi:
Aha cool I see you're already on it!

samsimpson1:
Not sure whats up with the argo application one, it's complaining that an object is set to null, but it needs to be null to disable automatic sync. I'm going to merge it anyway, since I'll be turning on auto-sync after I make sure prod isn't going to break

PR #1398, opened 2023-11-06
Add Google Tag Manager to apps owned by Publishing Experience team

ollietreend:
This pull request adds Google Tag Manager to publishing apps owned by the Publishing Experience team.

- Contacts Admin
- Content Publisher
- Manuals Publisher
- Maslow
- Service Manual Publisher
- Short URL Manager
- Specialist Publisher
- Travel Advice Publisher
- Whitehall

Production environments only need the `GOOGLE_TAG_MANAGER_ID` variable, which means GTM will default to the 'live' configuration.

For integration environments, we also set the `GOOGLE_TAG_MANAGER_AUTH` and `GOOGLE_TAG_MANAGER_PREVIEW` variables so that GTM uses the 'integration' configuration. This allows us to test GTM configs before publishing them to the 'live' production environment.

Note: GTM was already configured on Content Publisher. However, the performance analyst on our team has asked that we re-configure it to point to the same GTM container as our other applications. They no longer need the old GTM container.

Also: GTM was already configured on Whitehall in production, but I've also added it to integration.

Trello: https://trello.com/c/Wcuq3R1S/2167-add-ga4-page-view-tracking-to-apps-that-we-own


charts/app-config/values-integration.yaml
```diff
@@ -395,6 +395,12 @@ govukApplications:
               key: bearer_token
         - name: REDIS_URL
           value: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
+        - name: GOOGLE_TAG_MANAGER_ID
+          value: GTM-W573BJPG
+        - name: GOOGLE_TAG_MANAGER_AUTH # Non-prod only
+          value: 2jaYNm4QqCxqsb10EcYo5w
```
sengi, line 401: Recommend using a YAML anchor for (each of) these values to save repeating them over and over. (Anchors are easy to overuse, but this is probably a good example of where they help.) E.g.
```suggestion
          value: &gtm-id GTM-W573BJPG
        - name: GOOGLE_TAG_MANAGER_AUTH  # Non-prod only.
          value: &gtm-auth 2jaYNm4QqCxqsb10EcYo5w
```

then the rest become

```
        - name: GOOGLE_TAG_MANAGER_ID
          value: *gtm-id
        - name: GOOGLE_TAG_MANAGER_AUTH  # Non-prod only.
          value: *gtm-auth
```
etc.

sengi (approved):

sengi:
If it saves some time with the linter:

```sh
find . -type f -name \*.yaml | xargs gsed -Ei -e 's/(\w) #/\1  #/' -e 's/\w  # .*[^.?!()]$/\0./'
```

charts/app-config/values-integration.yaml
ollietreend: Thanks @sengi 👍🏻 you definitely caught me being verbose there. I've just added YAML aliases and have fixed up the lint errors.

PR #1399, opened 2023-11-06
Pass `suspend` field from cronTasks to CronJobs.

sengi:
This lets us use cronjobs as templates for manually-triggered / one-shot jobs by specifying `suspend: true` in the job's entry in the `cronTasks` Helm value.

Also document the `cronTasks` value.


csutter (approved):
🌺

PR #1401, opened 2023-11-06
Improve env config in `cron-task` template

csutter:
These currently work differently from deployments and workers, when they should be consistent and have the same environment configuration available to them.

- Add `SENTRY_DSN` if Sentry is enabled for the project
- Don't add `SENTRY_RELASE` if Sentry is not enabled (for consistency)
- Add `SECRET_KEY_BASE` is Rails is enabled for the project


sengi (approved):
Thanks!

PR #1416, opened 2023-11-07
Fix incorrectly formatted ALB listen ports for OAuth2 proxies

samsimpson1:


sengi (approved):
Oh yeah those annotations are nasty 😄

PR #1427, opened 2023-11-08
Dump/restore content-store before publishing-api.

sengi:
There's a data consistency issue in the design of Publishing API and Content Store; see https://trello.com/c/HRwJTf6S for details (internal only, sorry). The upshot is that some kinds of publishing operations don't work in the non-production environments after the nightly restore.

Work around this for now by running the content-store backup jobs before the publishing-api backup jobs in the their respective environments.

Also enable the restore/transform/backup job for content-publisher in staging (missed earlier).

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1367,7 +1367,7 @@ govukApplications:
         sagemakerIamRole: arn:aws:iam::172025368201:role/learn-to-rank-sagemaker
         serviceAccountIamRole: arn:aws:iam::172025368201:role/search-api-learn-to-rank-govuk
       draftContentStoreMongoPostgresCron:
-        schedule: "16 0 * * *"
+        schedule: "26 22 * * *"
         mongoExport:
           mongoDbUri: "mongodb://\
             mongo-1.production.govuk-internal.digital,\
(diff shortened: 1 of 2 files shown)


aldavidson:
I agree this should fix the problem, with one caveat - one of the main reasons for choosing 02:16 as the start time was that the (temporary) contentStoreMongoPostgresCron job starts at 00:16, and currently finishes around 01:40, and it's desirable to get the latest updates in from Mongo before copying that DB to staging & integration.

I guess there's no technical reason why that job needs to start after midnight, apart from it just being conceptually easier to think about if it's close to the day boundary..... so maybe we could also shift that cronjob to start at .... 22:30? That would give it time to complete before the backups run.

aldavidson:
I'll approve this anyway, as shifting the mongo -> postgres cron job could totally be a separate PR

aldavidson (approved):

sengi:
Ah thanks, that's a good point. Shall I shift the Mongo->Postgres job earlier as you suggest then?

aldavidson:
> Ah thanks, that's a good point. Shall I shift the Mongo->Postgres job earlier as you suggest then?

Yes please, if you would

PR #1448, opened 2023-11-14
Increase limits for Asset Manager virus scanning

brucebolt:
We are seeing issues where users uploading some files have those marked as "infected" when they are not.

In the two cases that we have investigated, this is caused by ClamAV raising either `Heuristics.Limits.Exceeded.MaxFiles` or `Heuristics.Limits.Exceeded.MaxScanSize`.  Full details of the investigation are in a comment on the linked Trello card.

Therefore increasing the size limit for MaxFiles from the default of 10,000 to 20,000 and MaxScanSize from 1200M to 2000M.

These new limits have been tested on a local version of ClamAV and have confirmed the affected files are now properly scanned.

[Trello card](https://trello.com/c/aL2EUaIv)

### DIFF
--- a/charts/asset-manager/templates/clamav-configmap.yaml
+++ b/charts/asset-manager/templates/clamav-configmap.yaml
@@ -20,11 +20,12 @@ data:
     Foreground yes
     LogTime yes
     LogVerbose yes
+    MaxFiles 20000
     # Whitehall and Specialist Publisher allow up to 500 MB uploads. Keep in
     # sync with StreamMaxLen, because clamdscan streams when connecting via TCP
     # (as opposed to UNIX socket).
     MaxFileSize 500M
-    MaxScanSize 1200M
+    MaxScanSize 2000M
     MaxScanTime 0
     MaxThreads 20
     SelfCheck 900


sengi (approved):
Those limits are getting quite large, to the point where users probably shouldn't be uploading a file that needs limits that high (because it'll likely be a shoddy experience for end users who download it).

How sure are you that the limit change is the thing that fixed the problem? For example, last time I investigated one of these, it was the act of retrying that led to a successful scan rather than the limit increase.

That said, might as well put this in since the downside is relatively small.

sengi:
Btw we're about to update the rather old version of `clamd` that's in there, so it'll be interesting to see if that improves things.

cc @nsabri1

brucebolt:
> Those limits are getting quite large, to the point where users probably shouldn't be uploading a file that needs limits that high.

The issue seems to arise from users uploading XLSX files (which are really just ZIP files) around 50MB, but are actually 650MB+ once unzipped.  ClamAV doesn't appear to deal with these very efficiently and reports a large scan size on such archives.

> How sure are you that the limit change is the thing that fixed the problem? For example, last time I investigated one of these, it was the act of retrying that led to a successful scan rather than the limit increase.

I could repeatedly reproduce the problem on my local machine using the same settings as previously when scanning the files sent to us by these publishers.  The files get scanned correctly with these new settings.

The best solution would be for us to move away from ClamAV to something else.  That's something we'll need to look at within Publishing Platform.

sengi:
Oh and it's worth checking that the files involved are well-formed too — we've seen malformed PDFs run into scanning limits for example. (I'm hoping that's the sort of thing that might improve with a more up-to-date clamd, but it's worth looking out for anyway.)

sengi:
Totally agree — running ClamAV ourselves was always madness.

sengi:
Kills me that we're publishing stuff in XLSX but heigh ho. Thanks for the explanation! All sounds good.

PR #1451, opened 2023-11-14
Maximise availability during Postgres restores.

sengi:
- Restore into a new database, then switch names with the old one.
- Create the target database if it doesn't already exist.
- Try to set the database owner (defaulting to the GOV.UK convention of the client app's username being the database name without the `_production` suffix), so that Rails database "migrations" (i.e. automated schema changes) should continue to work.
- Another stab in the dark at trying to get the progress logs to work in real time. (Not expecting any change; stderr is always supposed to be unbuffered anyway, but setting it to line-buffered was silly in any case.)

Tested: ran it on a copy of content_data_admin in staging; restore-then-dump job succeeded with and without a pre-existing target database.

### DIFF
--- a/charts/db-backup/scripts/backup-postgres
+++ b/charts/db-backup/scripts/backup-postgres
@@ -7,13 +7,13 @@ source_dir=$(dirname "${BASH_SOURCE[0]}")
 export PGUSER=$DB_USER
 export PGPASSWORD=$DB_PASSWORD
 export PGHOST=$DB_HOST
-export PGDATABASE=$DB_DATABASE
-readonly PGUSER PGPASSWORD PGHOST PGDATABASE
+export PGDATABASE=postgres
 
 backup () {
   local s3_path
-  s3_path="$PGHOST/$(date -u +%Y-%m-%dT%H%M%SZ)-$PGDATABASE.gz"
-  pg_dump -Fc -Z1 | progress | aws s3 cp - "$BUCKET/$s3_path"
+  s3_path="$DB_HOST/$(date -u +%Y-%m-%dT%H%M%SZ)-$DB_DATABASE.gz"
+  PGDATABASE=$DB_DATABASE \
+    pg_dump -Fc -Z1 | progress | aws s3 cp - "$BUCKET/$s3_path"
 }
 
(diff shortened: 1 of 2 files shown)


charts/db-backup/scripts/backup-postgres
```diff
@@ -31,10 +31,17 @@ restore () {
...
-  dropdb -ef --if-exists "$PGDATABASE"
-  aws s3 cp "$s3_url" - | progress | pg_restore -Cd postgres --no-comments
+
+  dropdb -ef --if-exists "$DB_DATABASE-restore"
+  createdb -eT template0 "$DB_DATABASE-restore"
+  aws s3 cp "$s3_url" - | progress | pg_restore -d "$DB_DATABASE-restore" --no-comments
+  psql -c "ALTER DATABASE \"$DB_DATABASE-restore\" OWNER TO \"$DB_OWNER\";" || true
+
+  psql -c "ALTER DATABASE \"$DB_DATABASE\" RENAME TO \"$DB_DATABASE-old\";" || true
+  psql -c "ALTER DATABASE \"$DB_DATABASE-restore\" RENAME TO \"$DB_DATABASE\";"
```
aldavidson, line 43: Would it be safer to combine these two into one `-c "..."` block, and therefore implicitly into one transaction? That way even if something goes wrong, you're still going to be left with a usable original database. The risk is small, admittedly, but maybe worth pre-emptively handling. 

```suggestion
  psql -c "ALTER DATABASE \"$DB_DATABASE\" RENAME TO \"$DB_DATABASE-old\"; ALTER DATABASE \"$DB_DATABASE-restore\" RENAME TO \"$DB_DATABASE\";"
```

sengi: I may eat my words if one of the really slow ones gets nobbled by some client query right at the end 😂

sengi:
Let's see how it goes 😅

charts/db-backup/scripts/backup-postgres
aldavidson: Totally makes sense, happy to go with this as-is - thanks! :rocket:

PR #1456, opened 2023-11-17
Try to avoid racing drop/create database.

sengi:


sengi:
I'll test this before merging; raising now cos running out of time before tonight's cronjobs 🙈

samsimpson1 (approved):

sengi:
Thanks @samsimpson1! 🙇

aldavidson:
I _think_ the explicit `BEGIN`/`COMMIT` statements should be redundant, but there's no harm in keeping them even if so.

aldavidson (approved):

sengi:
I removed the `-1` so the initial `DROP` should be in its own transaction (because autocommit). Dunno if I've fully correctly understood the situation though!

I think I've just figured out why the mysterious races happened on last night's run though. Pretty sure I left my test chart installed and hadn't bothered patching the values on that occasion, so there were two separate jobs racing for each restore/backup. Also explains the mysterious failures on the MySQL jobs that we'd never seen before. D'oh.

So having just realised that, I'd better see what happens tonight before merging this and then go from there.

Thanks for the quick reviews anyway though — might still come in handy!

sengi:
Yup, all works just fine when it's not trying to run two of each job simultaneously 🤦

Thanks for bearing with me.

PR #1465, opened 2023-11-20
Enable `uploadAssets` for `search-v2-evaluator`

csutter:


sengi (approved):
s/Enable/Disable/ in the description / commit message?

csutter:
> s/Enable/Disable/ in the description / commit message?
I'm enabling it by disabling the disabling... if that makes sense 😂

PR #1474, opened 2023-11-23
Install Dex via argocd in integration

samsimpson1:
This will cause the chart testing to break temporarily, since the schema for argo applications doesn't match actual functionality

### DIFF
--- a/charts/app-config/templates/dex.yaml
+++ b/charts/app-config/templates/dex.yaml
@@ -1,0 +1,23 @@
+{{ if has .Values.govukEnvironment (list "integration") }}
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: dex
+  namespace: {{ .Values.argoNamespace | default .Release.Namespace }}
+spec:
+  project: default
+  source:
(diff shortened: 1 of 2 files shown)


charts/app-config/templates/dex.yaml
```diff
@@ -0,0 +1,23 @@
...
+    chart: dex
+    targetRevision: "0.9.0"
+    helm:
+      values: |
+        {{- include "app-config.dex-values" . | nindent 8 }}
+  destination:
+    server: https://kubernetes.default.svc
+    namespace: {{ .Values.argoNamespace }}
+  syncPolicy:
+    automated: null
```
sengi, line 20: Just checking this is intentional? (no probs if so, just mentioning in case you'd left it in unintentionally)

sengi (approved):
Yay 🚀

charts/app-config/templates/dex.yaml
samsimpson1: Yep, setting this to null disables auto-sync so it won't make any changes if I've made a mistake in copying the config over

PR #1476, opened 2023-11-23
Use correct domain suffix for dex

samsimpson1:


sengi (approved):
Ah sorry didn't spot that either!

PR #1480, opened 2023-11-23
Log dep-resolution src content_id only where used.

sengi:
Only content-store actually logs the
[`govuk_dependency_resolution_source_content_id` request header](https://sourcegraph.com/search?q=repo%3Aalphagov+govuk_dependency_resolution_source_content_id), but we were including it on every nginx request log from every app even though it's always empty for all other apps.

There's also no reason for us to include it in nginx logs even for content-store, because it's already included in the app's own request logs and nginx doesn't do anything differently based on the presence or value of that header.

Conservatively: 50 M records/day * 50 bytes = 2.5 GB/day. That's 1.7% of our current 150 GB/day log quota in production.


aldavidson (approved):
LGTM :+1:

PR #1494, opened 2023-11-27
Add static mirror request monitoring from Grafana

MahmudH:
The static mirror dashboard was created manually using the Grafana UI. This PR replicates the dashboard to terraform.

Trello: https://trello.com/c/xrq1YM3B/3359-add-grafana-dashboard-for-monitoring-requests-to-the-mirror-to-govuk-helm-charts-2

### DIFF
--- a/charts/monitoring-config/dashboards/static-mirror-requests.json
+++ b/charts/monitoring-config/dashboards/static-mirror-requests.json
@@ -35,7 +35,7 @@
             "custom": {
               "axisCenteredZero": false,
               "axisColorMode": "text",
-              "axisLabel": "",
+              "axisLabel": "Total requests",
               "axisPlacement": "auto",
               "barAlignment": 0,
               "drawStyle": "line",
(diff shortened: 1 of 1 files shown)


sengi:
Thanks for adding this.

A few little things to improve:

- Don't forget the units! I can't be sure what the numbers mean on these graphs. I'm guessing requests per second?
- The dashboard doesn't fit on either of my screens (14in hidpi laptop or 1920x1200 16:10) without vertical scrolling, even though there are only three graphs. I'd suggest making the graphs less tall (like on most of our other dashboards) so that they can be seen at a glance.
- It would be helpful to call the second graph "requests by HTTP method" rather than just "requests" (since all the graphs show requests).
- You can just say "errors" rather than "error requests" (it's also clearer that way).

MahmudH:
Screenshot 2023-11-29 at 12 21 44

sengi (approved):
Thanks. I still don't see any units on the axes, but I see you've put them in the title on 2 out of 3 graphs so at least it's guessable now.

I'll leave it up to you if you wanna merge as is or polish this up a bit more.

PR #1509, opened 2023-11-30
Bump replica count for Content Store - 2ndline drill

davidgisbey:
## Description

We're scaling vertically as part of the 2ndline drill.


JamesCGDS (approved):

sengi:
FYI that's horizontally, but thanks for trying things out!

horizontal = more boxes
vertical = bigger/taller boxes

PR #1524, opened 2023-12-04
Temporarily re-add draftContentSoreMongoPostgresCron in staging

aldavidson:
...so that we can trial manually triggering this job, and make sure it completes while the DB is under daytime load


sengi (approved):
Can set `suspended: true` if you like, but no probs either way.

PR #1526, opened 2023-12-04
Fix bad regex in db_exists check.

sengi:
Wasn't allowing for database names containing dashes.

Tested: command now returns as expected for names like `email-alert-api_production`.


sengi:
`|` is a [special character in ERE, but only outside a bracket expression](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_04_03).

PR #1534, opened 2023-12-05
Tidy up email-alert-api pseudonymisation SQL script.

sengi:
- Use a temporary table so it'll be cleaned up automatically at the end of the session, even if something goes wrong.
- Cheaper to copy the relatively few rows we want to keep into a new table and drop the old one than to delete millions of rows.
- Cheaper to upsert instead of UNION DISTINCT.

It'd be better still to replace the whole thing with a decent set of test data that isn't based on information about real subscribers at all, but I'll leave that for another time.

Tested: ran successfully in staging; finishes in less than a minute now rather than several.


KludgeKML (approved):
Looks okay to me! (Although I'm unfamiliar with `-- noqa: AM04`, I assume that comment is disabling some linting?)

sengi:
Yup that's so sqlfluff doesn't complain about the `SELECT *` (which in most circumstances would be bad, but in this case we really do mean "whatever columns exist and we don't care what they are").

PR #1540, opened 2023-12-06
Run govuk-sli-collector jobs less often

mike29736:
I've broken something and these jobs are running for ages without succeeding.

I don't have time right this minute to figure out what's wrong, so it could be better not to have so many jobs spinning up ':|


sengi (approved):

sengi:
5 mins is right on the edge of what's suitable for a cronjob anyway tbh 😅

sengi:
10 mins sgtm 👍

PR #1545, opened 2023-12-06
Give whitehall and mongodb restores more RAM again.

sengi:
Tempted to set up VPA just for these :/

It shouldn't be using this much memory, but this time I couldn't find the corresponding oom-killer logs and the Prometheus metrics are way too coarse for this sort of thing, so kinda just guessing here.

Tested: ran the whitehall-mysql one successfully in staging.


sengi:
Thanks @MuriloDalRi 🙇

PR #1552, opened 2023-12-07
Re-enable staging publishing-api workers

aldavidson:
After switching the draft-content-store proxy over to Postgres the other day, looks like we forgot to re-enable the publishing-api workers on staging.


csutter (approved):
:shipit:

PR #1554, opened 2023-12-08
Make publish-intents asynchronous in integration

aldavidson:
This temporary change will allow us to test the following 2 PRs as part of preparing to switchover the live content-store in production on Monday:

https://github.com/alphagov/publishing-api/pull/2566

Once this is deployed, I'll manually test the change has the desired effect, then revert this PR

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1826,7 +1826,7 @@ govukApplications:
               name: publishing-api-postgres
               key: DATABASE_URL
         - name: ENQUEUE_PUBLISH_INTENTS
-          value: true
+          value: "true"
 
   - name: release
     helmValues:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1825,6 +1825,8 @@ govukApplications:
             secretKeyRef:
               name: publishing-api-postgres
               key: DATABASE_URL
+        - name: ENQUEUE_PUBLISH_INTENTS
+          value: true
```
sengi, line 1829: Can't remember off the top of my head whether bool true gets coerced to string somewhere along the way or not, but clearer to be explicit anyway.
```suggestion
          value: "true"
```

sengi (approved):

charts/app-config/values-integration.yaml
aldavidson: everything comes through to the apps as string values anyway, and the app is [checking for a string](https://github.com/alphagov/publishing-api/pull/2566/files#diff-b1570bf3b35a592d5767f288657003c340052a04b4480a34528327fee4f24a3fR9)

but I'm happy to be explicit :+1:

PR #1585, opened 2023-12-11
Have yamllint catch duplicate merge keys.

sengi:
Duplicate merge keys (`<<: foo`) are brittle because they're not valid in any YAML standard so far and they rely on implementation-dependent behaviour. There's also no advantage to them over specifying a single merge key with a list as its value.

Use the `yamllint` bundled with GitHub's runner image instead of a third-party action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,9 +62,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           show-progress: false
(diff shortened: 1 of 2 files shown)


samsimpson1 (approved):

sengi:
Ah I guess we need to wait for [yamllint 1.34](https://pypi.org/project/yamllint/) and for the [runner image](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) to catch up.

PR #1587, opened 2023-12-11
Add email alert api metric alerts and tests

KludgeKML:
Add alerts if `_current_` and `_delivered_` metrics for travel advice and medical safety alerts coming out of email-alert-api do not match.

https://trello.com/c/2R3zfYoj/2291-add-pagerduty-alert-from-new-email-alert-api-cronjob-output


charts/monitoring-config/rules/email_alert_api_alerts_tests.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+    alert_rule_test:
+      - alertname: TravelAdviceAlert
+        exp_alerts:
+          - exp_labels:
+              severity: page
+            exp_annotations:
+              summary: No delivered emails for Travel Advice update created >1 hour ago
+              # yamllint disable-line rule:line-length
+              runbook_url: 'https://docs.publishing.service.gov.uk/manual/alerts/email-alerts-travel-medical.html'
```
sengi, line 50: The linter wants you to do this:
```suggestion
              runbook_url:
                https://docs.publishing.service.gov.uk/manual/alerts/email-alerts-travel-medical.html
```
(it should just fix stuff like this automatically rather than complaining ofc, but GitHub makes that awkward to implement)

charts/monitoring-config/rules/email_alert_api_alerts_tests.yaml
```diff
@@ -0,0 +1,96 @@
+rule_files:
+  - email_alert_api_alerts.yaml
+
+evaluation_interval: 1m
+
+tests:
+  # Alert does not fire if current and delivered are identical
+  - name: travel_advice_alert_test_identical
+    interval: 1m
```
sengi, line 9: These [default to evaluation_interval](https://prometheus.io/docs/prometheus/latest/configuration/unit_testing_rules/#test_group).
```suggestion
```

sengi (approved):
Thanks!

charts/monitoring-config/rules/email_alert_api_alerts.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - name: EmailAlertAPIAlerts
+    rules:
+      - alert: TravelAdviceAlert
+        expr: >
+          email_alert_api_current_travel_advice_alerts -
+            email_alert_api_delivered_travel_advice_alerts > 0
+        labels:
+          severity: page
+        annotations:
+          summary: No delivered emails for Travel Advice update created >1 hour ago
```
sengi, line 11: maybe slightly clearer like this? what you reckon?
```suggestion
          summary: No delivery reports for Travel Advice update created >1 hour ago
```

KludgeKML:
Uh, what? I rebased this, and it removed the commits.

sengi:
Weird — were those commits already merged via another PR or something? 😕

KludgeKML:
I don't... think so? Let me look at the current state of things.

KludgeKML:
I now have a reopen option at least...

charts/monitoring-config/rules/email_alert_api_alerts_tests.yaml
KludgeKML: Done

charts/monitoring-config/rules/email_alert_api_alerts.yaml
KludgeKML: Yeah, that makes sense (it's also more truthful, since even with our fallback it doesn't necessarily mean no emails were delivered).

sengi (approved):

charts/monitoring-config/rules/email_alert_api_alerts.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - name: EmailAlertAPIAlerts
+    rules:
+      - alert: TravelAdviceAlert
+        expr: >
+          email_alert_api_current_travel_advice_alerts -
+            email_alert_api_delivered_travel_advice_alerts > 0
+        labels:
+          severity: page
+        annotations:
+          summary: No delivery reports for a Travel Advice update created >1 hour ago
```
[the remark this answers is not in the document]
KludgeKML: The metric only includes alerts it's valid to alarm on, which are ones older than an hour and younger than 2 days, so I _think_ we don't need to control for it here.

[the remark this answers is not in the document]
sengi: Ah ok fair enough 👍

PR #1590, opened 2023-12-12
Remove live content-store proxy & mongo apps in integration only

aldavidson:
As per #1548, but a) for the live content-store, and b) also removing the flag that disables router updates.

After switching the live content-store proxy to PostgreSQL yesterday, we can now remove the proxy in all environments, starting with integration (see steps 6 & 7 of the [plan](https://docs.google.com/document/d/1JFmnDoZCJBFF2J6-fbaFOzbeImGbNbE0D1nViIzS6aE/edit#heading=h.2g572px7zpvz), and this [Trello card](https://trello.com/c/6MxUbaHm/960-remove-production-live-content-store-proxy) ) . 

This will leave us with a single `content-store` application for each of live & draft, and no more proxy in front. 

This single app will still be using the `content-store-postgresql-branch` ECR repo for now - we have a separate Step 8 of the plan for how to return it to just `content-store`.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -678,8 +678,8 @@ govukApplications:
               name: signon-token-content-publisher-whitehall
               key: bearer_token
 
-  - name: content-store-mongo-main
-    repoName: content-store
+  - name: content-store
+    repoName: content-store-postgresql-branch
     helmValues: &content-store
       nginxClientMaxBodySize: 20M
       cronTasks:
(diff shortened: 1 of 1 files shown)


sengi (approved):

mike29736:
I'm struggling to fully understand the implications, but wondering if we need to make changes to Content Store deployments (at least switch off deployment promotion) if "content-store" refers to entirely different things in different environments?

mike29736:
Thanks @aldavidson. I am still having trouble with this, but I'll try thinking it through out loud for a minute and hope I'm not entirely wasting your time :thinking:

Since this change, there's no longer an `app-config` application in the integration environment that's associated with the `content-store` image. Which I think based on
> any applications in that cluster using that image get updated
means that we can still build and push the `content-store` image based on Content Store's `main` branch, but we can't spin it up in integration Kubernetes?

aldavidson:
Yes, that's right. But that should not be a problem -  since switching the production proxy last Monday to use the postgres content-store as its primary  (#1543), the `content-store` image is no longer used for anything on the critical path. It's only still there as an absolute worst-case failover until we remove it.

Issue #1588, opened 2023-12-12
Run Prometheus ruletests automatically


sengi:
We have [Prometheus ruletests and alert tests](https://github.com/alphagov/govuk-helm-charts/tree/main/charts/monitoring-config/rules) in but we haven't gotten around to setting them up to run automatically (and of course they therefore tend to rot and this leads to unfair extra hassle for whoever touches them).

Naturally, we want these to run automatically as pre-merge checks (for convenience and to stave off the bitrot).

The [GitHub Actions runner](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) doesn't include promtool yet (2023-12-12), but it might by the time you implement this.

PR #1618, opened 2023-12-13
Alert checks in CI

KludgeKML:
Adds promtool checking of alerting rules based on KludgeKML/promtool-action@v1 (forked from MatsuriJapon/promtool-action@v1 because that was pinned to an older version of prometheus that rejected some newer durations used in one test)

Updates rules files to work with linter, helm, and promtool.


KludgeKML:
Should resolve: https://github.com/alphagov/govuk-helm-charts/issues/1588

KludgeKML:
Made obsolete by https://github.com/alphagov/govuk-helm-charts/pull/1633

PR #1631, opened 2023-12-18
Fix smokey config dir for new headless chrome

theseanything:
The new headless chrome uses more of the same source code as the Chrome browser. This had led it trying to read and write user config. Here we're setting the user config folder in the /tmp folder (the only writable directory).


robinjam (approved):
🌟

PR #1633, opened 2023-12-19
Add Promtool checking into CI

KludgeKML:
Adds promtool checking of alerting rules with a local action.

Updates rules files to work with linter, helm, and promtool.

Should resolve https://github.com/alphagov/govuk-helm-charts/issues/1588

### DIFF
--- a/charts/monitoring-config/templates/prometheus-rules.yaml
+++ b/charts/monitoring-config/templates/prometheus-rules.yaml
@@ -10,6 +10,6 @@ spec:
 {{- range $path, $_ := .Files.Glob "rules/**.yaml" }}
 {{- if not (hasSuffix "_tests.yaml" $path) }}
   # {{ $path }}
-  {{- $.Files.Get $path | regexReplaceAll "^Agroups:" "" | nindent 2  }}
+  {{- $.Files.Get $path | replace "groups:\n" "" | nindent 2  }}
 {{ end }}
 {{ end }}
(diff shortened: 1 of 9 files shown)


.github/actions/promtool/Dockerfile
sengi, line 1: You can just run `prom/prometheus` directly — no need to wrap it. I tried it out: https://github.com/alphagov/govuk-helm-charts/commit/692b56a (feel free to cherry-pick + squash if you like)

sengi (approved):
Thanks for fixing all this! 🙇

charts/monitoring-config/templates/prometheus-rules.yaml
```diff
@@ -10,6 +10,7 @@ spec:
 {{- range $path, $_ := .Files.Glob "rules/**.yaml" }}
 {{- if not (hasSuffix "_tests.yaml" $path) }}
   # {{ $path }}
-  {{- $.Files.Get $path | nindent 4 }}
+  # {{- $.Files.Get $path | replace "groups:" "" | nindent 2  }}
```
sengi, line 13: Oh, just twigged these are YAML comments so we're still evaluating + outputting stuff from the Go/Helm template but just commenting out the first line of the template output 😅

sengi: The syntax highlighting doesn't help here!

PR #1636, opened 2023-12-19
Add env var to Content Data Admin

MuriloDalRi:
Add an env var and reduce the page size to 1000 to reduce memory usage. I have no particular reason to pick this number.

https://trello.com/c/MU9KwkRP/3368-improve-performance-of-content-data-csv-exports-3


sengi (approved):
Could maybe use a little more context — `FIND_CONTENT_QUERY_PAGE_SIZE` doesn't appear in github.com/alphagov/content-data-admin yet.

"Set Content Data Admin query page size" would be a better description.

sengi:
Ah thanks for linking the other PR 👍

PR #1640, opened 2023-12-19
Reinstate image-tags for govuk-sli-collector job

mike29736:
I've mostly copied this from the `govuk-application` template.

We recently extracted this standalone chart out of the deprecated `govuk-jobs` chart. Since that change meant no longer using the `govuk-application` template, we lost the ability for the cronjob to pull arbitrary images from ECR using `image-tags`. It can now only pull the `latest`-tagged (i.e. `main` branch) image.

After some deliberation, I've decided to add the use of `image-tags` to this standalone chart, allowing us to deploy arbitrary images again. The main motivation for this is for ease/flexibility of future development. It's hard to ignore that significant changes could benefit from the ability to deploy feature branches to integration. And while temporarily enabling the cronjob in integration will require only a small change, re-enabling the use of arbitrary images will be harder work that isn't guaranteed to be approachable for whoever is tasked with making the application code changes.

`automatic_deploys_enabled` and `promote_deployment` are set to false because (as I understand it) they don't mean anything for this non-`app-config` chart.

(https://trello.com/c/jDSDGNyn/814-measure-and-record-our-publishing-latency-sli)

### DIFF
--- a/charts/app-config/image-tags/production/govuk-sli-collector
+++ b/charts/app-config/image-tags/production/govuk-sli-collector
@@ -1,0 +1,3 @@
+image_tag: latest
+automatic_deploys_enabled: false
+promote_deployment: false
--- a/charts/govuk-sli-collector/templates/cronjob.yaml
+++ b/charts/govuk-sli-collector/templates/cronjob.yaml
@@ -1,4 +1,6 @@
 {{- $_ := .Values.govukEnvironment | required "govukEnvironment is required" }}
+{{- $imageTagConfig := $.Files.Get (printf "charts/app-config/image-tags/%s/govuk-sli-collector" .Values.govukEnvironment) | fromYaml }}
+{{- $imageTag := $imageTagConfig.image_tag }}
 ---
 apiVersion: batch/v1
 kind: CronJob
@@ -30,7 +32,7 @@ spec:
               emptyDir: {}
           containers:
             - name: main
-              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              image: "{{ .Values.imageRepository }}:{{ $imageTag }}"
               imagePullPolicy: "Always"
               env:
                 - name: INTERVAL_MINUTES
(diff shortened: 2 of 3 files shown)


charts/govuk-sli-collector/templates/cronjob.yaml
```diff
@@ -1,4 +1,6 @@
 {{- $_ := .Values.govukEnvironment | required "govukEnvironment is required" }}
+{{- $imageTagConfig := $.Files.Get (printf "charts/app-config/image-tags/%s/govuk-sli-collector" .Values.govukEnvironment) | fromYaml }}
```
mike29736, line 2: I'm not sure if this path will work or how to test it. Any ideas?

sengi: `helm template`? (see https://github.com/alphagov/govuk-helm-charts/#installing-an-application-chart-without-argo-cd)

sengi (commented):
I'm not getting what this buys you — it seems like it's just adding complexity for no gain, but maybe I'm missing something about your intended use case.

You can already deploy arbitrary images for testing just by running the build/push action in GitHub, no?

sengi:
`latest` != `main` btw

charts/app-config/image-tags/production/govuk-sli-collector
```diff
@@ -0,0 +1,3 @@
+image_tag: latest
+automatic_deploys_enabled: false
+promote_deployment: false
```
sengi, line 3: Better omit these two since they won't work (as you correctly identified already).
```suggestion
```

sengi (approved):
I'll not hold you up if decide you still want this. I think having the image tag file is fine, just as long as it doesn't have those other dead fields left in it to cause confusion.

mike29736:
> `latest` != `main` btw

Ah OK. I wasn't sure if the `latest` tag gets applied more than one way but I've only seen this one place that does it explicitly.

And that script is beyond my git comprehension, though I have only ever seen the `HEAD` of the `main` branch come back from that `REMOTE_HEAD_SHA=$(git ls-remote origin HEAD | cut -f 1)` command.

theseanything:
Closed in favour of #1654

PR #1643, opened 2023-12-19
Switch production draft content-store back to the content-store ECR repo

aldavidson:
We removed the proxy and the mongo content-store applications yesterday (#1598), we've also removed the legacy EC2 mongo instances, so we are ready to switch the content-store applications back to the `content-store` ECR repo in production.

Similar PRs in integration (#1635) and staging (#1637) went through with zero disruption so we're expecting the same for prod.


sengi (approved):
Woohoo!

aldavidson:
Amended the commit message, after realising I'd mistakenly said this was for the live content-store, when it should have said 'draft'

PR #1654, opened 2023-12-21
Enable deployment via image tag files for govuk-sli-collector

theseanything:
This adds the image tag files for govuk-sli-collector and references them from the argo-app. This allows us to use the existing deployment process to deploy the govuk-sli-collector.

The govuk-sli-collector is only deployed into the production environment. However, the image tags files are available incase it's needed to be deployed into either integration or staging for testing.

### DIFF
--- a/charts/govuk-sli-collector/templates/cronjob.yaml
+++ b/charts/govuk-sli-collector/templates/cronjob.yaml
@@ -30,7 +30,7 @@ spec:
               emptyDir: {}
           containers:
             - name: main
-              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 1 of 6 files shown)


charts/govuk-sli-collector/templates/cronjob.yaml
```diff
@@ -55,7 +55,7 @@ spec:
                       name: govuk-sli-collector-sentry
                       key: dsn
                 - name: SENTRY_RELEASE
-                  value: "{{ .Values.image.tag }}"
+                  value: "{{ .Values.images.GovukSliCollector.tag }}"
```
mike29736, line 58: :+1:

mike29736 (approved):
Thanks for this :bow:

PR #1660, opened 2024-01-02
Updating config

nsabri1:
Missed include `app-config.helm-release`

### DIFF
--- a/charts/app-config/templates/kubescape.yaml
+++ b/charts/app-config/templates/kubescape.yaml
@@ -7,22 +7,21 @@ spec:
   destination:
     name: kubescape
     namespace: kubescape
-    server: 'https://kubernetes.default.svc' # change to your server
+    server: https://kubernetes.default.svc
   source:
-    path: charts/kubescape-operator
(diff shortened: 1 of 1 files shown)


charts/app-config/templates/kubescape.yaml
```diff
@@ -7,22 +7,21 @@ spec:
...
     name: kubescape
     namespace: kubescape
-    server: 'https://kubernetes.default.svc' # change to your server
+    server: https://kubernetes.default.svc
   source:
-    path: charts/kubescape-operator
-    repoURL: 'https://github.com/kubescape/helm-charts'
-    targetRevision: HEAD
-    helm:
+    {{- include "app-config.helm-release"
```
samsimpson1, line 12: This only works if there's an entry in these files that you want renovate to keep up to date: https://github.com/alphagov/govuk-helm-charts/tree/main/charts/app-config/helm-versions

nsabri1: ahh ok then no need for this - ok thank you

PR #1664, opened 2024-01-03
Chart for migrating Router's legacy MongoDB to k8s.

sengi:
- Service/LoadBalancer creates an NLB (L4 LB) via aws-load-balancer-controller so replicaset members outside the cluster can connect to the one inside.
- New node bootstraps itself from the existing replicaset when added via `rs.add('<new-hostname>')`.
- Just one in-cluster replica for now, to keep the networking straightforward. We can add another in-cluster replica once the out-of-cluster replicas and the Service are gone.

Tested: running in staging.


theseanything (approved):

charts/router-mongo/templates/service.yaml
```diff
@@ -0,0 +1,20 @@
+---
```
samsimpson1, line 1: I wonder if this will cause issues with mongo's replica set logic. Iirc, the hostname that is defined in the replica set config has to match the host set on the mongo host. It should be possible to work around it though

samsimpson1 (approved):

charts/router-mongo/templates/service.yaml
sengi: Yeah it'll only work for one replica, for sure. Here's my [rollout plan](https://docs.google.com/document/d/1_6nAsqAnrYpCOMCVw1cde6xdymgnRPCv6bvbQZ_Yyic/edit#) so far.

sengi: Figured I should stick a description on it — thanks!

PR #1677, opened 2024-01-04
Add whitehall scheduled publications overdue prom rule

samsimpson1:
This is to replace the old Icinga-based check for overdue scheduled editions

See also: https://github.com/alphagov/whitehall/pull/8694


sengi (approved):
Neat!

PR #1687, opened 2024-01-04
Remove unnecessary router-mongo svc template kludge.

sengi:
It's much simpler to have just the final, headless service in the Helm template and create separate services (outside Helm/Argo) for the temporary load balancer stuff that's only needed during the migration. (Especially since we need one service for each new node, because we have to add the new nodes before we remove the old ones.)


theseanything (approved):

sengi:
(fixed merge conflict)

Issue #1661, opened 2024-01-02
Cut down on renovate noise


samsimpson1:
We're getting (at least) 1 renovate update a day at the moment, and it's generating a lot of noise in slack and toil having to merge them in manually. We could try to cut down on noise generated by renovate by auto-merging updates where appropriate

PR #1706, opened 2024-01-10
Auto-merge more patch updates

samsimpson1:
These packages don't seem to include any breaking changes in patch version updates

#1661


sengi (approved):
Good call — for all of those, we're definitely better off being on the latest patch release by default.

PR #1708, opened 2024-01-10
Add email-alert-api tokens for content-data-admin in  staging/production

KludgeKML:
Values added for integration earlier, add them for other environments.

https://trello.com/c/xng5X4uX/30-add-email-statistics-into-content-data-admin

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -455,6 +455,11 @@ govukApplications:
             secretKeyRef:
               name: signon-token-content-data-admin-content-data-api
               key: bearer_token
+        - name: EMAIL_ALERT_API_BEARER_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: signon-token-content-data-admin-email-alert-api
+              key: bearer_token
         - name: SITEIMPROVE_API_CLIENT_USERNAME
           valueFrom:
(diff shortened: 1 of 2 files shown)


peteglondon (approved):

sengi (approved):
fyi
```
k get secret signon-token-content-data-admin-email-alert-api
... "signon-token-content-data-admin-email-alert-api" not found
```

KludgeKML:
@sengi Ah yeah, the tokens have only just been made in production, so I don't think the job has run to copy them. But they won't be needed until tomorrow (at which point it'll also have been copied to Staging I assume?) . I think that's okay, unless I've misunderstood something?

sengi:
It's fine, but your rollout will just be stuck until the cronjob runs and creates the missing secrets.

sengi:
You can run [the cronjob](https://argo.eks.production.govuk.digital/applications/signon?orphaned=false&resource=&node=batch%2FCronJob%2Fapps%2Fsignon-sync-token-secrets-to-k8s%2F0) sooner to make the crashloopy pods go away and let the rollout finish, if you like (three-vertical-dots menu -> create job)

KludgeKML:
Ah, I didn't realise that would block the rollout. Thanks! Done.

PR #1713, opened 2024-01-11
Auto-merge kube-prometheus-stack patches

samsimpson1:
These should be fine to auto-merge, even though they do seem to make small changes to the values it takes (to fix regressions or typos)

#1661


sengi (approved):
🚀

PR #1714, opened 2024-01-11
Reduce log verbosity on locations-api in prod/staging.

sengi:
PostcodesCollectionWorker generates about 30k log entries an hour and they're not useful (mostly just messages like "start" containing no other info).


theseanything (approved):

sengi:
Aha it's `RAILS_LOG_LEVEL` on this one apparently.

PR #1721, opened 2024-01-15
Use range instead of seperate templates for licensify

theseanything:
Each of the licensify application have a common set of templates that don't differ. Switching to using a range loop to create the resources for all apps from a single template, instead of duplicating the templates per app. Makes it much easier to make changes to common configuration.


charts/licensify/values.yaml
```diff
@@ -18,43 +18,49 @@ clamav:
...
+    replicaCount: 1
+    port: 9950
+    healthcheckPath: "/"
+    ingress:
+      enabled: true
+      annotations: {}
+    service:
+      port: 80
+
+  licensifyBackedn:
```
sengi, line 33:
```suggestion
  licensifyBackend:
```

charts/licensify/templates/ingress.yaml
```diff
@@ -1,26 +1,31 @@
-{{ $app := .Values.licensifyFrontend }}
-{{ $_ := set .Values "appName" $app.name }}
+{{- range .Values.apps }}
+{{- if .ingress.enabled }}
+{{ $_ := set $.Values "appName" .name }}
```
sengi, line 3: Do you need to modify `.Values`? Might be cleaner to just define a variable, like `$appName := .name` (same as you're doing for `$svcPort`)?

sengi (approved):

charts/licensify/templates/ingress.yaml
theseanything: Ah so the reason is to make "appName" available for the "licensify.labels", by just adding it to the Values dict we pass it. (There might be a nicer way to do that.)

charts/licensify/values.yaml
theseanything: Ah thank you! I couldn't workout why the diff was so weird

charts/licensify/templates/ingress.yaml
sengi: Oh I see. Yeah, you probably wanna just change that partial template in _helpers.tpl to do what you want (probably just delete the `.Values` and pass it `.`) — but no biggie. As long as the labels are coming out the way you want them!

PR #1722, opened 2024-01-15
Add whitehall unenqueued scheduled publications prom rule

samsimpson1:
Related: https://github.com/alphagov/whitehall/pull/8735


sengi (approved):
Ah ok does it turn out we still need this after all? Haha. Thanks!

PR #1724, opened 2024-01-15
Add nginx to licensify

theseanything:
This adds a nginx server infront of each of the licensify apps.

### DIFF
--- a/charts/licensify/templates/nginx-configmap.yaml
+++ b/charts/licensify/templates/nginx-configmap.yaml
@@ -1,0 +1,125 @@
+{{- range .Values.apps }}
+{{ $_ := set $.Values "appName" .name }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .name }}-nginx-conf
+  labels:
+    {{- include "licensify.labels" $ | nindent 4}}
+    app: {{ .name }}
+    app.kubernetes.io/component: nginx
+data:
+  nginx.conf: |-
+    error_log  /dev/stderr warn;
+    pid        /tmp/nginx.pid;
+
+    events {
+      worker_connections  1024;
+    }
+
+    http {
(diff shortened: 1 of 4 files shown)


charts/licensify/templates/nginx-configmap.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+        # Endpoint that isn't cached, which is used to assert that an external
+        # service can receive a response from GOV.UK origin on www hostname. It
+        # is intended for pingdom monitoring
+        location = /__canary__ {
+          default_type application/json;
+          add_header cache-control "max-age=0,no-store,no-cache";
+          return 200 '{"message": "Tweet tweet"}\n';
+        }
+
```
sengi, line 124: I think we can probably drop this bit — `/__canary__` was a www.gov.uk thing and was never used for probing Licensify.

charts/licensify/values.yaml
```diff
@@ -18,6 +18,22 @@ clamav:
...
     port: 3310
 
+nginx:
+  port: 8080
+  proxyConnectTimeout: 30s
+  proxyReadTimeout: 15s
+  clientMaxBodySize: 200M
+  extraHttpConf: |
+    limit_req_zone $binary_remote_addr zone=licensing:1m rate=1r/s;
+    limit_req_status 429;
```
sengi, line 28: We can drop this too (as long as we stick the AWS WAF config on the load balancer).

charts/licensify/values.yaml
```diff
@@ -111,6 +127,10 @@ config:
   noReplyMailAddress: test@example.com
 
 images:
+  nginx:
+    repository: public.ecr.aws/nginx/nginx
+    tag: stable-perl
```
sengi, line 132: Looks like we don't need perl this time?

sengi (approved):

PR #1725, opened 2024-01-15
Use nginx /readyz for ALB healthcheck

theseanything:
Now that we've got Nginx running in front of the apps we can use the default /readyz healthcheck path for our ALBs.


sengi (approved):
Nice!

PR #1726, opened 2024-01-15
Add Licensify to staging and production

theseanything:
This shouldn't interfere with current version on the old infra as works via the eks.production.govuk.digital domain


charts/app-config/values-production.yaml
```diff
@@ -1367,6 +1367,60 @@ govukApplications:
...
+        frontendBaseUrl: https://licensify.eks.production.govuk.digital
+        signonBaseUrl: https://signon.publishing.service.gov.uk
+        licenceFinderUrl: https://www-origin.publishing.service.gov.uk/licence-finder
+        performancePlatformUrl: >-
+          https://www.performance.service.gov.uk/data/licensing/application
+        googleAnalytics:
+          accountAdmin: UA-34530684-2
+          domainAdmin: publishing.service.gov.uk
+          accountFrontend: UA-26179049-1
+          domainFrontend: .www.gov.uk
```
sengi, line 1420: Is the leading `.` intended?

charts/app-config/values-production.yaml
```diff
@@ -1367,6 +1367,60 @@ govukApplications:
...
+            - licensify-documentdb-0.cafmt8xopqlp.eu-west-1.docdb.amazonaws.com
+            - licensify-documentdb-1.cafmt8xopqlp.eu-west-1.docdb.amazonaws.com
+            - licensify-documentdb-2.cafmt8xopqlp.eu-west-1.docdb.amazonaws.com
+        elmsPdfApiHost: https://gdsprodaws.aptosolutions.co.uk/GDSELMS-5/
+        baseUrl: https://licensify-admin.eks.production.govuk.digital
+        frontendBaseUrl: https://licensify.eks.production.govuk.digital
+        signonBaseUrl: https://signon.publishing.service.gov.uk
+        licenceFinderUrl: https://www-origin.publishing.service.gov.uk/licence-finder
+        performancePlatformUrl: >-
+          https://www.performance.service.gov.uk/data/licensing/application
```
sengi, line 1415: This is long dead but presumably doesn't matter?

dj-maisy (approved):
I'm liking this - especially like that the DB hosts are now a proper list. 👍

charts/app-config/values-staging.yaml
```diff
@@ -1331,6 +1331,58 @@ govukApplications:
...
+        baseUrl: https://licensify-admin.eks.staging.govuk.digital
+        frontendBaseUrl: https://licensify.eks.staging.govuk.digital
+        signonBaseUrl: https://signon.staging.publishing.service.gov.uk
+        licenceFinderUrl: https://www-origin.staging.publishing.service.gov.uk/licence-finder
+        performancePlatformUrl: >-
+          https://www.staging.performance.service.gov.uk/data/licensing/application
+        googleAnalytics:
+          accountAdmin: UA-34530684-2
+          domainAdmin: www-origin.staging.publishing.service.gov.uk
+          accountFrontend: UA-26179049-1
```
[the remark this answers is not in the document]
theseanything: So I've copied these values directly from the licensify boxes on EC2. Looks as tho integration goes to some other GA account (probably one that apto have access to?). They've got staging and production to the same GA app. I'll update staging to go to staging, but leave integration as is.

charts/app-config/values-production.yaml
theseanything: Copied over from what was already in production - will correct!

charts/app-config/values-production.yaml
theseanything: Yeah - shouldn't have an effect. Sadly the app still expects the config to be there.

charts/app-config/values-staging.yaml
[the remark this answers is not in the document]
sengi: Heh, maybe staging got messed up back when it was moved from UKCloud to EC2 😅 Anyway no worries, thanks for fixing!

charts/app-config/values-production.yaml
sengi: I can't actually find any usage of this in the app 😂 never mind haha

theseanything:
> Do you need to disable `feed` so it doesn't pick up tasks in prod?

Ah yes - because I assume it'll pick jobs up from the db.

theseanything:
> Do you need to disable `feed` so it doesn't pick up tasks in prod?

I've switched off generating the feed deployment for production - tested with helm template and the production values.

PR #1735, opened 2024-01-16
Allow `search-api-v2` workers to export metrics

csutter:
This app has a persistent background worker that needs the Prometheus exporter to be running in order to be able to export metrics.


sengi (approved):
Ah sorry about this :(

PR #1742, opened 2024-01-17
Add govuk_content_production back into backup config.

sengi:
It turns out this MongoDB/DocumentDB database was only *mostly* owned by content-store (which was migrated to Postgres). It's still accessed by publisher, specialist-publisher and service-manual-publisher.

Need to tease out what on earth is going on with these three apps and figure out how bad the situation is and how to fix it, but that's a different matter. For now, just keep backing this stuff up so at least we're not SOL if we lose prod.

### DIFF
--- a/charts/db-backup/values.yaml
+++ b/charts/db-backup/values.yaml
@@ -219,6 +219,8 @@ cronjobs:
         - op: backup
           db: travel_advice_publisher_production
         - op: backup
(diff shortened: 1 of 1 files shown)


theseanything (approved):

sengi:
I suppose we still would have had DocDB snapshots so would have still been covered, but hey.

PR #1743, opened 2024-01-17
Use /api/licences as healthcheck for licensify-frontend

theseanything:
The previous path was a external redirect causing probe warnings. This path should return a 200.


theseanything:
> This might be a little on the expensive side for a healthcheck but let's launch and iterate 🚀

For sure! I couldn't see another route that wasn't dependent on values in a database.

PR #1746, opened 2024-01-18
Make Licensify config consistent with EC2

theseanything:
This adds in missing config options that are present in the configure of licensify on EC2 and tidies up the structure.

### DIFF
--- a/charts/licensify/values.yaml
+++ b/charts/licensify/values.yaml
@@ -92,6 +92,7 @@ extraEnv: []
 
 config:
   clamAntivirusHost: clamav
+  scheduledVirusScanCronExpression: 0 0 16 * * ?
 
   mongo:
     dbHosts: []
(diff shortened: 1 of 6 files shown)


dj-maisy (approved):
🚀

charts/licensify/values.yaml
```diff
@@ -104,29 +105,56 @@ config:
...
+    emailPeriodicEnabled: true
+    periodicEmailCron: 0 0 6 3 * ?
+    emailOverrideRecipient: ""
+
+  # Performance platform has been decomission useless config
+  performancePlatformUrl: >-
+    https://performance/data/licensing/application
+  performanceDataSenderCronExpression: 0 0 1 * * ?
+
+  placesApiUrl: http://places-api:9700/places
```
sengi, line 155: I don't think this exists, but presumably that doesn't matter if it was already like that 😄

sengi (approved):

charts/licensify/values.yaml
sengi: Oh totally — def the right call to leave it in for now!

PR #1747, opened 2024-01-18
Add logging config for Licensify

theseanything:
Logging for these applications is configured via XML file 😢. This might be overly verbose, but can turn it down after migration.

### DIFF
--- a/charts/licensify/templates/deployment.yaml
+++ b/charts/licensify/templates/deployment.yaml
@@ -35,6 +35,9 @@ spec:
         - name: licensify-config
           secret:
             secretName: licensify-config
+        - name: logging-config
+          configMap:
+            name: licensify-logging-config
         - name: app-tmp
           emptyDir: {}
         - name: nginx-conf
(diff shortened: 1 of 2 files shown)


charts/licensify/templates/deployment.yaml
```diff
@@ -61,6 +64,8 @@ spec:
               value: do-not-use
             - name: GDS_CONFIG_FILE
               value: /etc/licensing/config.properties
+            - name: LOGGING_CONFIG_FILE
+              value: ../../../app/conf/logging.xml
```
sengi, line 68: optional: is there a benefit to using this odd-looking relative path here? (as opposed to just having it match `/app/conf/logging.xml` below?

sengi (approved):

charts/licensify/templates/deployment.yaml
theseanything: Ah no, it's just that what ever loads the config always assumes its relative even with a starting "/", so setting "/app/conf/logging.xml" ends up being something like "/app/feed/bin/app/conf/logging.xml".

sengi: Haha oh wow. No worries! (thanks for clearing that up!)

PR #1753, opened 2024-01-19
Add JAVA_OPTIONS for licensify frontend

theseanything:
This is extra config specific for licensify frontend. Pulled from puppet here:
https://github.com/alphagov/govuk-puppet/blob/5fe296410ed2a438acdc356b58fa72dd66fe0c21/modules/licensify/templates/gds-licensify-config.conf.erb#L23

This also fixes the file path for the logging config to be absolute, than a weird relative path.

This is possible due to these changes with image entrypoint:
https://github.com/alphagov/licensify/pull/645


sengi (approved):
🚀

PR #1771, opened 2024-01-22
Update Licensify domains to use EKS instance in production

theseanything:
This switches Licensify to be served from EKS in production.


sengi (approved):

sengi (approved):
LGTM

PR #1773, opened 2024-01-22
Fix DNS records for licensify in integration

theseanything:
We want external DNS to create our internal DNS records and we update the public facing records in govuk-dns-tf instead.


sengi (approved):
Sorry I didn't spot this in the earlier PRs!

theseanything:
> Sorry I didn't spot this in the earlier PRs!

No worries! I should have noticed the config on other publishing apps.

PR #1774, opened 2024-01-22
Run Smokey against failover CDN

ChrisBAshton:
In https://github.com/alphagov/smokey/pull/1221, Smokey was
configured to be able to run against the failover CDN (Cloudfront)
or against any of our [three mirrors](https://docs.publishing.service.gov.uk/manual/fall-back-to-mirror.html),
by means of a proxy that can be enabled and configured through
ENV variables.

Scenarios that should pass for the failover CDN are:
- Any scenarios NOT tagged `@notcloudfront`
Scenarios that should pass for the mirrors are:
- Any scenarios tagged `@worksonmirror`

We don't want to run these Smokeys as often as the 'main' Smokey
test suite, but we do want to periodically run them to ensure that
the failover solution and mirrors continue to work as expected.

For now, we'll just set up the failover CDN Smokey. In future,
it should be fairly straightforward to add the mirror tests
into the same cronjob.

Tested by running a modified snippet from the README:

```sh
cd charts/app-config

ENVIRONMENT=production
APP=smokey
helm template $USER-${APP?} ../smokey --values <(
  helm template . --values values-${ENVIRONMENT}.yaml |
  yq e '.|select(.metadata.name=="'$APP'").spec.source.helm.values'
) --set sentry.enabled=false --set rails.createKeyBaseSecret=false
```

...then looking for the `smokey-failover` cronjob in the output.
I also confirmed that no such job exists when swapping `ENVIRONMENT`
to `integration`.

Trello: https://trello.com/c/2XwAz6JD/3365-run-automatic-smokey-tests-against-the-standby-cdn-5

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2561,6 +2561,28 @@ govukApplications:
 
   - name: smokey
     chartPath: charts/smokey
+    helmValues:
+      cronJobs:
+        - name: smokey
+          schedule: "*/10 * * * *"
+          args:
+            - cucumber
+            - --profile={{ .Values.govukEnvironment }}
+            - --strict-undefined
+        - name: smokey-failover
+          schedule: "0 9 * * 1-5"
+          args:
+            - cucumber
+            - --strict-undefined
+            - --tags=not @notcloudfront
+          extraEnv:
+            - name: GOVUK_PROXY_PROFILE
+              value: failoverCDN
+            - name: FAILOVER_CDN_HOST
+              valueFrom:
+                secretKeyRef:
+                  key: FAILOVER_CDN_HOST
+                  name: smokey-failover-cdn-host
 
   - name: static
     helmValues:
--- a/charts/smokey/templates/cronjob.yaml
+++ b/charts/smokey/templates/cronjob.yaml
@@ -1,15 +1,17 @@
+{{- range .Values.cronJobs }}
+---
 apiVersion: batch/v1
 kind: CronJob
 metadata:
-  name: smokey
+  name: {{ .name }}
 spec:
   concurrencyPolicy: Allow
   failedJobsHistoryLimit: 3
   successfulJobsHistoryLimit: 3
-  schedule: {{ .Values.cronSchedule | quote }}
+  schedule: {{ .schedule | quote }}
   jobTemplate:
     metadata:
-      name: smokey
+      name: {{ .name }}
     spec:
       activeDeadlineSeconds: 600
       backoffLimit: 0
@@ -27,12 +29,11 @@ spec:
             name: tmp
           containers:
           - args:
-            - cucumber
-            - --profile={{ .Values.govukEnvironment }}
-            - --strict-undefined
-            - --tags=not @not{{ .Values.govukEnvironment }}
-            image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
-            name: smokey
+            {{- with .args }}
+              {{- tpl (toYaml .) $ | nindent 14 }}
+            {{- end }}
+            image: "{{ $.Values.appImage.repository }}:{{ $.Values.appImage.tag }}"
+            name: {{ .name }}
             resources:
               limits:
                 cpu: "2"
@@ -53,21 +54,21 @@ spec:
                   key: bearer_token
                   name: signon-token-smokey-asset-manager
             - name: ENVIRONMENT
-              value: {{ .Values.govukEnvironment }}
+              value: {{ $.Values.govukEnvironment }}
             - name: GOVUK_APP_DOMAIN
               value: ""
             - name: GOVUK_APP_DOMAIN_EXTERNAL
-              value: {{ .Values.publishingDomainSuffix }}
+              value: {{ $.Values.publishingDomainSuffix }}
             - name: GOVUK_ASSET_ROOT
-              value: https://{{ .Values.assetsDomain }}
+              value: https://{{ $.Values.assetsDomain }}
             - name: GOVUK_WEBSITE_ROOT
-              value: https://www.{{ .Values.externalDomainSuffix }}
+              value: https://www.{{ $.Values.externalDomainSuffix }}
             - name: PLEK_SERVICE_ASSETS_URI
-              value: https://{{ .Values.assetsDomain }}
+              value: https://{{ $.Values.assetsDomain }}
             - name: PLEK_SERVICE_ASSETS_ORIGIN_URI
-              value: https://assets-origin.{{ .Values.publishingDomainSuffix }}
+              value: https://assets-origin.{{ $.Values.publishingDomainSuffix }}
             - name: PLEK_SERVICE_CONTENT_DATA_ADMIN_URI
-              value: https://content-data.{{ .Values.publishingDomainSuffix }}
+              value: https://content-data.{{ $.Values.publishingDomainSuffix }}
             - name: PLEK_USE_HTTP_FOR_SINGLE_LABEL_DOMAINS
               value: "1"
             - name: SIGNON_EMAIL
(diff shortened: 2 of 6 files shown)


charts/smokey/values.yaml
```diff
@@ -5,6 +5,13 @@ publishingDomainSuffix: test.publishing.service.gov.uk
 assetsDomain: assets.test.publishing.service.gov.uk
 
 cronSchedule: "*/10 * * * *"
+smokeyCronTasks:
```
sengi, line 8: The `smokey` prefix is redundant. (Also "jobs" might be more intuitive, since each represents a cronjob?)
```suggestion
cronJobs:
```

charts/smokey/templates/cronjob.yaml
```diff
@@ -54,6 +52,10 @@ spec:
                   name: signon-token-smokey-asset-manager
             - name: ENVIRONMENT
               value: {{ .Values.govukEnvironment }}
+            {{- if .proxyProfile }}
```
sengi, line 55: Could just make this a generic `extraEnv` and pass GOVUK_PROXY_PROFILE in.

charts/smokey/templates/cronjob.yaml
```diff
@@ -1,15 +1,17 @@
...
 kind: CronJob
 metadata:
-  name: smokey
+  name: {{ .name }}
 spec:
   concurrencyPolicy: Allow
   failedJobsHistoryLimit: 3
   successfulJobsHistoryLimit: 3
-  schedule: {{ .Values.cronSchedule | quote }}
+  schedule: {{ .schedule | default .Values.cronSchedule | quote }}
```
sengi, line 11: I'd just drop `.Values.cronSchedule` for the sake of simplicity.

charts/smokey/templates/cronjob.yaml
```diff
@@ -26,13 +28,9 @@ spec:
           - emptyDir: {}
             name: tmp
           containers:
-          - args:
-            - cucumber
-            - --profile={{ .Values.govukEnvironment }}
-            - --strict-undefined
-            - --tags=not @notreplatforming and not @not{{ .Values.govukEnvironment }}
+          - args: {{ .task }}
```
sengi, line 31: I would call this `args` (since it literally is) rather than `task`.

sengi (commented):
For the sake of developer experience / usability, I'd recommend having just one additional cronjob for these new scenarios rather than 4 extra copies.

charts/app-config/values-production.yaml
```diff
@@ -2590,6 +2590,30 @@ govukApplications:
...
+          args:
+            - cucumber
+            - --profile={{ .Values.govukEnvironment }}
+            - --strict-undefined
+            - --tags=not @notreplatforming and not @not{{ .Values.govukEnvironment }}
+        - name: smokey-failover
+          schedule: "0 9 * * 1-5"
+          args:
+            - env GOVUK_PROXY_PROFILE=failoverCDN
+            - ruby govuk_proxy_runner.rb &
```
sengi, line 2606: Does this work? It looks a bit like shell script but isn't gonna be interpreted as such. You can either fork/exec this govuk_proxy_runner from Smokey (based on whether `GOVUK_PROXY_PROFILE` is non-empty`) or define a separate container for the proxy (by changing the template to let you pass it in from here).

charts/app-config/values-production.yaml
```diff
@@ -2590,6 +2590,30 @@ govukApplications:
...
+          schedule: "*/10 * * * *"
+          args:
+            - cucumber
+            - --profile={{ .Values.govukEnvironment }}
+            - --strict-undefined
+            - --tags=not @notreplatforming and not @not{{ .Values.govukEnvironment }}
+        - name: smokey-failover
+          schedule: "0 9 * * 1-5"
+          args:
+            - env GOVUK_PROXY_PROFILE=failoverCDN
```
sengi, line 2605: https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/

```suggestion
          env:
            - name: GOVUK_PROXY_PROFILE
              value: failoverCDN
          args:
```

charts/smokey/templates/cronjob.yaml
```diff
@@ -26,13 +28,9 @@ spec:
           - emptyDir: {}
             name: tmp
           containers:
-          - args:
-            - cucumber
-            - --profile={{ .Values.govukEnvironment }}
-            - --strict-undefined
-            - --tags=not @notreplatforming and not @not{{ .Values.govukEnvironment }}
+          - args: {{ .args }}
```
sengi, line 31: You're defining `.args` as a list in `values.yaml`, but interpreting it as a string here (which makes it harder to see what's going on). (You probably want a `toYaml` here somewhere.)

sengi (commented):
How have you tested this? (You can install the chart temporarily via Helm, if you haven't already — example in the top-level README.)

charts/app-config/values-production.yaml
sengi: Oh wait we're still in values.yaml here sorry — you'll need to pass this into the template, but the idea is still the same.

charts/app-config/values-production.yaml
```diff
@@ -2557,6 +2557,30 @@ govukApplications:
...
     chartPath: charts/smokey
+    helmValues:
+      cronJobs:
+        - name: smokey
+          schedule: "*/10 * * * *"
+          args:
+            - cucumber
+            - --profile={{ .Values.govukEnvironment }}
+            - --strict-undefined
+            - --tags=not @not{{ .Values.govukEnvironment }}
```
sengi, line 2568: The [profiles](https://github.com/alphagov/smokey/blob/aadea2b/config/cucumber.yml) already do this. (Similarly in values.yaml below.)
```suggestion
```

charts/smokey/templates/cronjob.yaml
```diff
@@ -81,17 +82,21 @@ spec:
...
-              value: "{{ .Values.appImage.tag }}"
+              value: "{{ $.Values.appImage.tag }}"
             - name: SE_CACHE_PATH
               value: "/tmp/.cache/selenium"
             - name: XDG_CONFIG_HOME
               value: "/tmp/.chrome"
             - name: XDG_CACHE_HOME
               value: "/tmp/.chrome"
+            {{- with .extraEnv }}
+              {{- (tpl (toYaml .) $) | trim | nindent 12 }}
```
sengi, line 100: Best to avoid the `tpl` hack until you definitely need it; otherwise template parsing tends to get slow (and it encourages overabstraction).
```suggestion
              {{- . | toYaml | nindent 12 }}
```

sengi (approved):

charts/app-config/values-production.yaml
ChrisBAshton: Ah yes, good shout. I've also removed the `profile` from the `smokey-failover` job, where there is currently no corresponding profile (so are setting the tags explicitly).

ChrisBAshton:
Paired with @theseanything to test (thanks, Sean!):

```
# env setup
APP=smokey
ENVIRONMENT=integration
k config use-context integration
eval $(gds aws govuk-integration-poweruser -e --art 8h)

# check output: no-op as expected
helm template $USER-${APP?} ../smokey --values <(\n  helm template . --values values-${ENVIRONMENT}.yaml |\n  yq e '.|select(.metadata.name=="'$APP'").spec.source.helm.values'\n) --set sentry.enabled=false --set rails.createKeyBaseSecret=false
```

Swapped `helm template` for `helm install` and it broke:
> Error: INSTALLATION FAILED: Unable to continue with install: CronJob "smokey" in namespace "apps" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "christopher.ashton-smokey"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "apps"

This is because the smokey cronjob already exists. For other apps, this is handled by dynamically generating the name, but Smokey is just hardcoded.

I had to rename [smokey](https://github.com/alphagov/govuk-helm-charts/blob/0e815867e1aaaac97d3b9e05713383fb8808ee4d/charts/smokey/templates/cronjob.yaml#L4) to `smokey-tmp`,  [smokey-sentry](https://github.com/alphagov/govuk-helm-charts/blob/1ac9bde6de3d6fa79f67456975ee7e884a4ddd75/charts/smokey/templates/sentry-external-secret.yaml#L4) to `smokey-sentry-tmp` and [smokey-test](https://github.com/alphagov/govuk-helm-charts/blob/4f917d620130f3ac0f94fd1f75b8021d34030bba/charts/smokey/templates/workflow-templates/smoke-test.yaml#L4) to `smoke-test-tmp`. Then the install ran fine.

```
k get cronjobs
# saw `smokey-tmp` in output

k create job --from=cronjob/smokey-tmp smokey-chris-tmp
# to create a job (rather than wait until 9am!)

k get jobs
# saw job in output

k logs jobs/smokey-chris-tmp
# fetch the logs
# bonus: follow the logs with `k logs -f jobs/smokey-chris-tmp`, or get a summary of job pass/failure with `k describe jobs/smokey-chris-tmp`
```

Undid everything with `helm uninstall $USER-${APP?}`.

Repeated the above for Production, but this time checking the `smokey-failover` job, as that's what's being added. Hardcoded the `smokey-failover-cdn-host` name so that the secret could be found. All worked fine, so merging.

PR #1778, opened 2024-01-23
Wait a little bit longer for healthcheck on licenisfy

theseanything:
Still seeing some occasions where the healthchecks happen before the app is ready and are coming back unhealthy.


sengi (approved):

sengi:
No need to worry about this just yet but you can almost certainly drop initialDelaySeconds and just tune startupProbe.

PR #1781, opened 2024-01-23
Move Licensify into separate namespace

theseanything:
This allows us to restrict access for third parties to just the Licensify resources. This should also keep the Argo Application resource in the apps namespace.

As per [the docs](https://argocd-operator.readthedocs.io/en/latest/usage/deploy-to-different-namespaces/), they suggest that namespaces created by Argo require the "managed-by' annotation. Also useful anyway to keep track of what created the ns.


sengi (approved):
Thanks!

PR #1787, opened 2024-01-24
Add replicas for Licensify frontend

theseanything:
Noticed that we had 2 instances of licenisfy-frontend running in Production. So should be able to run replicas for it here too.


sengi (approved):

dj-maisy (approved):
Ah yes, the power of Kubernetes! Scale up simply by changing a number! 🚀

PR #1808, opened 2024-01-28
Fix the params as app is failing due to missing account

nsabri1:


charts/app-config/templates/kubescape.yaml
```diff
@@ -13,13 +13,18 @@ spec:
     repoURL: 'https://github.com/kubescape/helm-charts'
     targetRevision: HEAD
     helm:
-      values: |
+      valueFiles:
+        - values.yaml
+      parameters:
```
sengi, line 18: You can keep the original syntax if you use [valuesObject](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/) instead of `parameters`.

charts/app-config/templates/kubescape.yaml
```diff
@@ -13,13 +13,18 @@ spec:
     repoURL: 'https://github.com/kubescape/helm-charts'
     targetRevision: HEAD
     helm:
-      values: |
+      valueFiles:
+        - values.yaml
```
sengi, line 17: Is values.yaml not included by default?

sengi (approved):

charts/app-config/templates/kubescape.yaml
nsabri1: I will have a look at that if this fails again

charts/app-config/templates/kubescape.yaml
nsabri1: following this example https://github.com/kubescape/helm-charts/blob/b54d494a7631214f4355636fb36b551e908c611f/ArgoCDApplication.yaml

nsabri1: Saying that the values.yaml should be included by default

PR #1812, opened 2024-01-29
Fix values on Kubescape 

nsabri1:
Error:

`one or more objects failed to apply, reason: Application.argoproj.io "kubescape" is invalid: spec.source.helm.parameters[3].value: Invalid value: "boolean":`


samsimpson1 (approved):

sengi (approved):
Ah, YAML

PR #1814, opened 2024-01-29
Use HTTP 1.1 for proxing requests

theseanything:
This is because licensify need to support chunked responses. https://medium.com/@yatskevich/chunked-responses-with-nginx-and-play-framework-3ffada39ea0c


dj-maisy (approved):
🚀

PR #1818, opened 2024-01-31
Increase router nginx's client_max_body_size for licensify requests

theseanything:
We allow files upto 10MB to be uploaded to licensify, however since moving to EKS  nginx at router has served 413 responses to users. The reason we're only seeing it since the migration, is that previously file upload request bypassed router completely. They were uploaded on a seperate `uploadlicence` domain that went directly to licensify frontend.

Tested in integration.


sengi (approved):
Aha good catch.

sengi:
Are all the forms < 10 meg?

sengi:
Sorry just re-read the description — ignore me 😅

PR #1819, opened 2024-01-31
Remove place-api and performance platform config from Licensify

theseanything:
This is no longer used as per:
- https://github.com/alphagov/licensify/pull/652
- https://github.com/alphagov/licensify/pull/651


sengi (approved):
Thanks!

PR #1825, opened 2024-02-06
Add publishing api secret envvar to government-frontend

unoduetre:
## Make sure the secrets with the matching names exist and are configured in SIgnon before this PR is merged.

Use PUBLISHING_API_BEARER_TOKEN envvar to propagate signon-token-government-frontend-publishing-api secret to government-frontend app.

This is required by the `publishing_api:publish_static_pages` rake task in government-frontend.

[Trello ticket](https://trello.com/c/R9OkZLQw/2362-missing-description-for-1-horse-guards-road-history-page-on-search-results-m-l)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1167,6 +1167,11 @@ govukApplications:
       extraEnv:
         - name: MEMCACHE_SERVERS
           value: frontend-memcached-govuk.eks.integration.govuk-internal.digital
+        - name: PUBLISHING_API_BEARER_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: signon-token-government-frontend-publishing-api
+              key: bearer_token
 
   - name: draft-government-frontend
     repoName: government-frontend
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1180,6 +1185,11 @@ govukApplications:
       extraEnv:
         - name: PLEK_HOSTNAME_PREFIX
           value: draft-
+        - name: PUBLISHING_API_BEARER_TOKEN
```
sengi, line 1188: Are you sure the draft deployment needs this? Is there a use case where you need to run the rake task from the draft version? No problem if so, but if not, please remove the unneeded secretKeyRefs from the draft deployments.

sengi (approved):
You shouldn't need to create any secrets for this btw; they should already exist. If they don't, then something isn't configured right in Signon.

charts/app-config/values-integration.yaml
unoduetre: Thanks, I've updated the comment to instead make sure they exist. The comment is exactly about what you said - to check if the secrets have been configured in SIgnon.

I am not sure about the second comment, could you explain a bit about how draft works (I joined reasonably recently)? If it's not enabled in draft, how would updates propagate there? Is there another mechanism than running a rake task there?

sengi: Do you need to run the rake task on the -draft version? Or just the regular version?

unoduetre: I am not sure how draft works. If publishing api is shared between draft and non-draft, I don't think that'd be necessary, as the requests would be exactly the same.

sengi: It literally just comes down to whether you would run the Rake task on the draft version or just the regular version.

PR #1826, opened 2024-02-07
Add alerting for `search-api-v2` invariant tests

csutter:
These tests run against a number of datasets in `search-api-v2` to ensure that when searching for certain important queries, a set of expected results are returned, and publish a Prometheus gauge labelled by dataset name and type.

- Alert whenever the reported value of a dataset of type `invariant` drops below `1` (100%)
- Add routing destination for search team to general Alertmanager configuration


sengi (approved):
LGTM!

Issue #1597, opened 2023-12-12
Kubeconform schemas need updating?


sengi:
We have [kubeconform schemas](https://github.com/alphagov/govuk-helm-charts/tree/main/schemas) in-tree for the CRDs that we use (for those that kubeconform doesn't already know about).

We haven't updated them in a while so we might start tripping over validation errors where schemas lack fields etc. from newer versions of CRDs.

samsimpson1:
We could point kubeconform to this repo: https://github.com/datreeio/CRDs-catalog

The CRDs there are updated pretty often, and they seem to have all of the CRDs we're using

PR #1836, opened 2024-02-12
Use CRD library instead of schemas

nimalank7:
Description:
- Currently we use CRD schemas which are located in the /schema directory. However updating these is a chore which we'd rather automate
- Using [datreeio](https://github.com/datreeio/CRDs-catalog)'s CRD catalogue will have `kubeconform` download the schemas instead
- Closes [issue 1597](https://github.com/alphagov/govuk-helm-charts/issues/1597)


.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           args: >
             -kubernetes-version ${{ env.KUBERNETES_VERSION }}
             -schema-location default
-            -schema-location "./schemas/{{ .ResourceKind }}_{{ .ResourceAPIVersion }}.json"
+            -schema-location "https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json"
```
sengi, line 45: I think you can fix the line length lint like this:
```suggestion
            -schema-location
            "https://raw.githubusercontent.com/datreeio/CRDs-catalog/main/{{.Group}}/{{.ResourceKind}}_{{.ResourceAPIVersion}}.json"
```

README.md
```diff
@@ -76,8 +76,29 @@ You can then run `helm search repo <alias>` to see the charts.
...
+locally.
+
+You can run the validation tests locally by installing `kubeconform` and running
+
+```shell
+mkdir helm-dist
+for c in $(ls charts)
+do
+  helm template "${c}" "charts/${c}" --output-dir helm-dist
+done
```
sengi, line 95: I know it was already like this, but it didn't work as it was 😄 
```suggestion
for c in charts/*; do
  helm template "$(basename "$c")" "$c" --output-dir helm-dist
done
```

sengi (approved):
Yay, thanks!

You can drop the issue number from the PR title btw, GitHub doesn't link them and it just makes it harder to read the titles.

README.md
nimalank7: Yep fixed!

.github/workflows/ci.yml
nimalank7: Fixed by adding a `\` to break the string over into multiple lines.

PR #1849, opened 2024-02-19
Stop copying nonprod databases around on weekends.

sengi:
Little if anything changes over the weekend, so there's really no reason to keep running these restore jobs on Saturday and Sunday mornings. Some of them are nontrivial in terms of resource usage.

~Leaving the small ones for now.~ Actually let's just do all of 'em, arguably slightly less surprising that way.

### DIFF
--- a/charts/db-backup/values.yaml
+++ b/charts/db-backup/values.yaml
@@ -29,7 +29,7 @@ resources:
     cpu: 400m
     memory: 384Mi
 
-# TODO: figure out why memory usage is so spiky in these cases.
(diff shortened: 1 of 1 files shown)


samsimpson1 (approved):

aldavidson (approved):
Seems sensible to me. I wonder if it might be easier to conceptualise / remember if it was all of them (i.e. everything gets copied during the week, nothing at the weekend)  but no biggie,  I'm happy either way

sengi:
Thanks, yeah that's a fair point — certainly easier to describe. Let's go with that 👍

PR #1850, opened 2024-02-19
Combine the two promtool check jobs.

sengi:
These are both < 1s so running two jobs doesn't buy us much. There's maybe a tiny bit of benefit to having them both run independently, but it doesn't seem worth the duplication in this case IMHO.


theseanything (approved):

theseanything:
Yup, see the pros for separately or combines jobs, but either is fine!

PR #1854, opened 2024-02-20
Use kubedns, not CloudFlare, for nginx's async resolver.

sengi:
This should help to reduce the annoying GuardDuty alerts that Cybersecurity team spend time manually forwarding to us.

#### Alternatives considered

We could configure a VPC endpoint for S3 with a fixed IP address and point nginx at that, avoiding the need for nginx to use its weird resolver implementation, but that just seems complicated.

### DIFF
--- a/charts/asset-manager/templates/nginx-configmap.yaml
+++ b/charts/asset-manager/templates/nginx-configmap.yaml
@@ -210,9 +210,7 @@ data:
           # want to unset a request header sent TO S3 by nginx.
           proxy_set_header Authorization "";
 
-          # Add CloudFlare and Google DNS server to avoid "no resolver defined to resolve"
-          # errors when trying to connect to S3.
-          resolver 1.1.1.1 8.8.8.8 8.8.4.4;
+          resolver kube-dns.kube-system.svc.cluster.local.;
 
           # Download the file and send it to client
           proxy_pass $download_url;


charts/asset-manager/templates/nginx-configmap.yaml
```diff
@@ -210,9 +210,7 @@ data:
           # want to unset a request header sent TO S3 by nginx.
           proxy_set_header Authorization "";
 
-          # Add CloudFlare and Google DNS server to avoid "no resolver defined to resolve"
-          # errors when trying to connect to S3.
-          resolver 1.1.1.1 8.8.8.8 8.8.4.4;
+          resolver 172.20.0.10;  # kube-dns.kube-system.svc.cluster.local.
```
samsimpson1, line 213: according to the docs, this can be a hostname (not sure how it resolves it), would it make sense to set this to the hostname in case kube-dns isn't on that IP?

samsimpson1: Also, we aren't setting this in the router nginx config, and that instance also needs to resolve hostnames

sengi: Oh cool, thanks for spotting that! Done.

samsimpson1 (approved):

charts/asset-manager/templates/nginx-configmap.yaml
sengi: Sorry only just saw your second comment there! I'll check what's going on with the Router one. Good catch!

PR #1868, opened 2024-02-29
Create default RBAC resources for Argo Workflows

theseanything:
Currently Argo Workflows use the default service account and the executor-role is bound to it. This makes it difficult to understand which service account and permissions are being used by workflows. These RBAC resources will replace the "executor" role and role bindings, which will be explicity defined as the workflow default service account.


sengi (approved):
Awesome, this is gonna save a whole bunch of faff on our next security audit! ❤️

PR #1873, opened 2024-03-01
Add event binding and rbac resources for deploy image

theseanything:
This enables the deploy image workflow to be triggered using the Argo Workflows Event API, instead of via the Argo Events. Replacing usage of Argo Events will allow us to remove it from our cluster as a dependency.


sengi (approved):
Yay!

PR #1877, opened 2024-03-01
Fix argo workflow job names

theseanything:
This restores the workflow job naming to when we used Argo Events to trigger them.


sengi (approved):
Cool, thanks! Nice bit of UX detail there! 🙌

PR #1878, opened 2024-03-01
Fix argo workflow job names

theseanything:
Jobs must be unique (so adding unix timestamp) and fix referencing the payload.


sengi (approved):
LGTM (could potentially use `generateName` here but I think the timestamp thing you've done here is actually a bit nicer 👍)

PR #1887, opened 2024-03-05
Remove info-frontend

1pretz1:
Info pages have been unpublished and thus info-frontend isn't serving any traffic regarding info pages.

We are seeing the odd request for legacy /needs pages [1] (~1 a day) which are being received by info-frontend. info-frontend doesn't know about these pages which result in 404 being returned. needs do not exist on the public facing website and only exist in the publishing stack (published by Maslow). After we remove info-frontend, these requests will return a 503, as the backend app serving the needs pages will no longer be available.

Returning 503s isn't ideal but we agreed [2] to is the best approach given:
- the traffic is tiny
- the traffic mainly comes from crawlers
- we will only return 503s up until we retire Maslow and then we'll return a normal 404

The alternative would be to unpublish the 679 editions / routes now in our publishing stack but we run the risk of strange behaviour / errors as we have inconsistent states between our apps.

Trello:
https://trello.com/c/NcSjgmmq/2385-remove-from-govuk-helm-charts

[1]: https://kibana.logit.io/s/13d1a0b1-f54f-407b-a4e5-f53ba653fac3/goto/9f68346dc881bd4913aaeaf185522608?security_tenant=global
[2]: https://gds.slack.com/archives/C03D4A72HGE/p1709568749883129?thread_ts=1709556305.963639&cid=C03D4A72HGE

### DIFF
--- a/charts/app-config/templates/env-configmap.yaml
+++ b/charts/app-config/templates/env-configmap.yaml
@@ -29,7 +29,7 @@ data:
   {{- end }}
   GOVUK_SIDEKIQ_JSON_LOGGING: "1"
   GOVUK_WEBSITE_ROOT: https://www.{{ .Values.externalDomainSuffix }}
-  PLEK_UNPREFIXABLE_HOSTS: account-api,asset-manager,feedback,imminence,info-frontend,licensify,local-links-manager,locations-api,search-api,signon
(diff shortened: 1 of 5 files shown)


KludgeKML (approved):
LGTM!

PR #1894, opened 2024-03-11
Update Kubeconform to use Kubernetes 1.29 schemas

nimalank7:
Description:
- K8s has been updated to 1.29.0 in EKS so this change configures Kubeconform to use the 1.29.0 schemas


sengi (approved):
Thanks!

PR #1897, opened 2024-03-12
Add `search-api-v2` query failure AlertManager alert

csutter:
This adds a new alert for when the rate of successful `search-api-v2` queries drops below a certain threshold.

### DIFF
--- a/charts/monitoring-config/rules/search_api_v2.yaml
+++ b/charts/monitoring-config/rules/search_api_v2.yaml
@@ -11,3 +11,30 @@ groups:
           description: >-
             Invariant dataset '{{ $labels.dataset_name }}' has dropped below 100% recall (i.e. the
             expected results are not being returned for one or more queries in this dataset)
+
+      - alert: HighQueryFailureRate
+        expr: >-
+          sum(increase(http_requests_total{
+            namespace="apps",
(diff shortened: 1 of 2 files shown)


charts/monitoring-config/rules/search_api_v2.yaml
```diff
@@ -11,3 +11,30 @@ groups:
...
+            action="show",
+            status="200"
+          }[1m]))
+          /
+          sum(increase(http_requests_total{
+            namespace="apps",
+            job="search-api-v2",
+            controller="searches",
+            action="show"
+          }[1m]))
```
sengi, line 30: You need a window of 2m or more for this to work, otherwise you're always getting an empty result.

Also while `increase` works here, `rate` communicates the intent more clearly and is the more usual way to write a ratio query such as this. (`increase` is literally just `rate` \* the sampling interval.)

```suggestion
          sum(rate(http_requests_total{
            namespace="apps",
            job="search-api-v2",
            controller="searches",
            action="show",
            status="200"
          }[2m]))
          /
          sum(rate(http_requests_total{
            namespace="apps",
            job="search-api-v2",
            controller="searches",
            action="show"
          }[2m]))
```

sengi (approved):

PR #1912, opened 2024-03-21
Make Elasticsearch index restore cron more robust.

sengi:
Lately this cronjob has failed a couple of times with what looks like a race where we issue the restore request before the index deletion has fully completed.

I couldn't find a way to make the deletion request synchronous/blocking, so for now just add a hacky fixed delay.

Also:

- Exclude system indices (the ones with names starting with a dot). This doesn't matter right now for ES6, but makes it less likely to break if we upgrade.
- Log JSON error messages (response bodies) in the trace for easier debugging.

Tested: ran successfully in the staging cluster.

### DIFF
--- a/charts/search-index-env-sync/snapshot.sh
+++ b/charts/search-index-env-sync/snapshot.sh
@@ -52,9 +52,14 @@ snapshot_valid () {
 
 restore () {
   snapshot_valid
-  $curl -XDELETE "$ES_URL/_all"
+  $curl -XDELETE "$ES_URL/*,-.*" >&2  # Delete all except system indices.
+  sleep 2  # TODO: use ?wait_for_shards=all once our ES version supports it.
   local result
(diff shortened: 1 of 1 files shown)


charts/search-index-env-sync/snapshot.sh
```diff
@@ -52,9 +52,14 @@ snapshot_valid () {
 
 restore () {
   snapshot_valid
-  $curl -XDELETE "$ES_URL/_all"
   local result
-  result=$($curl -XPOST "$ES_URL/_snapshot/$SNAPSHOT_REPO/$SNAPSHOT_NAME/_restore")
+  result=$($curl -XDELETE "$ES_URL/*,-.*")  # Delete all except system indices.
```
samsimpson1, line 56: What does setting this output to results do if it's getting overwritten by the restore curl?

samsimpson1 (approved):

PR #1917, opened 2024-03-22
Be clear the Smart Answer Zendesk credentials are same as Support

ChrisBAshton:
As per https://docs.publishing.service.gov.uk/kubernetes/manage-app/manage-secrets/#add-secrets-to-your-app, we should avoid duplicating secrets:

> If you decide that you need to use a secret, check if that secret
> already exists in the AWS Secrets Manager. If the secret does
> exist, use this secret instead of duplicating it.

See the production secrets - the username and password are the same:
- https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=govuk%2Fsupport%2Fzendesk&region=eu-west-1
- https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=govuk%2Fsmartanswers%2Fzendesk-client&region=eu-west-1

In Integration and Staging, the password is blank, but given this secret is only used by a [smart answers rake task](https://github.com/alphagov/smart-answers/commit/dfe2f8af089b28f36b9922d53a0e08deedfe6ab1), it is only intended to be run in Production anyway.


sengi (approved):
Oh nice, good find! Thanks!

sengi:
I think you might be able to get rid of the corresponding external-secrets/templates/smartanswers/zendesk-client.yaml too?

PR #1924, opened 2024-03-26
Add cronjob for publishing finders on Specialist Publisher

GDSNewt:
Currently we have to manually publish pre-production flagged finders in integration. This cronTask will automate the process in integration.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2656,6 +2656,10 @@ govukApplications:
             ]}}]
         hosts:
           - name: specialist-publisher.{{ .Values.k8sExternalDomainSuffix }}
+      cronTasks:
+        - name: publish-finders
+          task: "publishing_api:publish_finders"
+          schedule: "49 7 * * 1-5"
       extraEnv:
         - name: ASSET_MANAGER_BEARER_TOKEN
           valueFrom:


charts/app-config/values-integration.yaml
```diff
@@ -2656,6 +2656,10 @@ govukApplications:
             ]}}]
         hosts:
           - name: specialist-publisher.{{ .Values.k8sExternalDomainSuffix }}
+      cronTasks:
+        - name: publish-finders
+          task: "publishing_api:publish_finders"
+          schedule: "0 0 * * *"
```
sengi, line 2662: I recommend randomising the minute (so your job is less likely to cause a cluster scale-out and have to wait for a new node) and picking an hour that's after the nightly environment sync jobs will have finished (so the changes in publishing-api aren't overwritten before the morning).

Also might as well avoid running on weekends.

```suggestion
          schedule: "49 7 * * 1-5"
```

sengi (approved):

charts/app-config/values-integration.yaml
sengi: (sadly Kubernetes' cronjob still doesn't have a proper random-minute feature like some other cron implementation do)

GDSNewt: Thanks Chris, all sensible suggestions. I'll change it now 😃

PR #1925, opened 2024-03-27
Suspend search-index-env-sync cron in integration.

sengi:
Now that site search no longer uses Elasticsearch, there's little reason to copy the Elasticsearch indices from staging to production every day.

The job has lately started flaking because (for as yet unknown reasons) the search-api-worker deployment seems to be doing a lot of inserts which re-create the index in between the `DELETE` and the `POST _restore` (even if we issue them in immediate succession). This leaves the integration environment in a non-working state every few days, which is way worse than some data being a bit stale (that probably nobody even needs anyway).

There are a few features that still use Elasticsearch, such as [finders], but at this point it's likely that the disadvantages of clobbering the integration indices every morning outweigh the advantages.

It's easy enough to run the job on-demand (from the ⋮ menu -> Create job in the [Argo CD web UI](https://argo.eks.integration.govuk.digital/applications/cluster-services/search-index-env-sync?view=tree&node=batch%2FCronJob%2Fapps%2Fsearch-index-env-sync-copy-from-stag%2F0), or from the [command line](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_job/)) if it's needed.

Also remove the delay between the delete and the restore, since we now know this exacerbates the race condition rather than mitigating it.

[finders]: https://github.com/alphagov/finder-frontend#live-examples

### DIFF
--- a/charts/search-index-env-sync/snapshot.sh
+++ b/charts/search-index-env-sync/snapshot.sh
@@ -53,7 +53,6 @@ snapshot_valid () {
 restore () {
   snapshot_valid
   $curl -XDELETE "$ES_URL/*,-.*" >&2  # Delete all except system indices.
-  sleep 20  # TODO: use ?wait_for_active_shards=all once our ES version supports it.
   local result
   result=$(
     # TODO: use curl --json once available.
--- a/charts/search-index-env-sync/values.yaml
+++ b/charts/search-index-env-sync/values.yaml
@@ -26,6 +26,7 @@ cronjobs:
       args: [create_and_clean_up]
   integration:
     - name: copy-from-stag
+      suspend: true
       schedule: "17 6 * * *"
       args: [restore_latest]
       extraEnv:


theseanything (approved):

ryanb-gds:
Hi @sengi, I think this is potentially going to cause issues for departments testing specialist finders in integration. Most likely they're going to be seeing a mismatch between what Specialist Publisher shows them and what appears in the Specialist Finder search results. Obviously we could run the sync job ourselves on demand on a regular basis, and there's a strong argument that departments shouldn't be testing things in integration, but we don't want to be doing even more manual toil to support Specialist Finders. Would you consider reverting this?

Alternatively could departments test using staging?

sengi:
Hi, @ryanb-gds. We could revert this, but I'm pretty sure that would only worsen your situation in the short term. It'd also break other teams who rely on the integration environment for rollout automation, for example.

The staging environment (unlike integration) isn't available outside GDS, so it probably doesn't help us here unfortunately. If you don't need users outside GDS to access your demos then for sure you could use staging.

I still think this change (i.e. making this cronjob on-demand-only) is the best we can do right now to unblock your team's day-to-day workflow (and avoid breaking other teams' ability to roll out releases) while giving us some time to figure out a more sustainable solution.

I've already spent a good couple of days on this issue (which so far appears to be largely outside my team's sphere of control). I've done this because I want to do the right thing and help your team to be productive. Unfortunately there are limits to this; I'm accountable to other stakeholders so I can't drop too many of my team's priorities in order to work on what is essentially a 3rd-party tech-debt issue (neither my team's nor yours, but there's nobody else to fix it so we're all doing the best we can with what we've got).

There is an aspect of the problem which may be under your team's control, though. It seems that a part of the problem is that `search-api-worker` is making very frequent requests that create the `govuk` index pretty continuously and this appears to be what leads to the race condition that's causing the cronjob to fail (and leaving integration in an non-working state).

When @lauraghiorghisor-tw and I chatted yesterday we agreed to:

- suspend the cronjob (so that at least integration doesn't keep breaking altogether)
- find out whether this actually does break your workflow for testing finders or not, and if so, exactly what data is missing
- if it does, you can run the job (it only takes two clicks, though you may need to retry it if it flakes) as and when you need it
- I (or my team) can then use what we learn from that to figure out a sensible next step towards a long-term, maintainable solution

I don't see how automatically running a job every morning that has a >50% chance of breaking the whole integration environment for multiple teams would be an improvement over running the job on-demand where you can see if it fails right away and do something about it (even if it's just retrying it a few times and asking for help if that doesn't work).

sengi:
Or are you saying that other departments are continually end-to-end testing their own content changes in integration using the finders feature? i.e. are we dealing with another major gap in preview functionality here?

(If that's the case then I'd really appreciate if you could bang the drum for prioritising fixing these preview feature gaps as they severely hurt our ability as a programme to respond to user needs because they prevent us from moving towards good testing practices, which severely slows down development.)

ryanb-gds:
Hi @sengi, firstly, thanks very much for your time looking into this. I appreciate you're doing it entirely out of goodwill, apologies if I'm seeming presumptuous. I'm operating from a place of not understanding enough about how Search API works.

That's very interesting about the search-api-worker causing issues, I wasn't aware of that and we will look into it. Unfortunately we've set an expectation with some of our users that they will have a fresh copy of data from production available on integration each day to test with members of the public, so we're in a slightly difficult position. However, that's our problem, not yours and it's up to us to solve it. We're not talking about previewing in this case, but I do wonder if that might offer a solution to allow this sort of testing to take place in production. Thanks again for your help with this.

sengi:
No worries! Thanks for the extra context, that's useful to help me prioritise potential fixes/workarounds.

I wasn't previously aware that your users' current expectations would result in you needing to run the job every day — I agree that's a pain.

So what I'm hearing is there's value in us finding a less toily workaround in at least the short term. I have some ideas that *might* fix the cronjob but unfortunately none of them are particularly straightforward to implement nor guaranteed to work, so the tradeoffs are a bit tricky.

I'll keep looking into it as and when I can. Let's keep sharing any findings.

PR #1926, opened 2024-03-27
Add Cron Task for Travel Advice Publisher

baisa:
This is to publish any potentially unpublished editions that were scheduled for the past.

[Trello card](https://trello.com/c/UX0jNZC5/2419-cronjob-for-scheduling)

Rake task was added to the publisher here https://github.com/alphagov/travel-advice-publisher/pull/1859

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2975,6 +2975,10 @@ govukApplications:
 
   - name: travel-advice-publisher
     helmValues:
+      cronTasks:
+        - name: publish-scheduled-editions
+          task: "publish_scheduled_editions"
+          schedule: "29 5 * * *"
       nginxClientMaxBodySize: *max-upload-size
       workerEnabled: true
       ingress:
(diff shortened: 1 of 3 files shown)


charts/app-config/values-production.yaml
```diff
@@ -2975,6 +2975,10 @@ govukApplications:
 
   - name: travel-advice-publisher
     helmValues:
+      cronTasks:
+        - name: publish-scheduled-editions
+          task: "publish_scheduled_editions"
+          schedule: "0 0 * * *"
```
sengi, line 2981: Please randomise the minute (so that you're not stacking up load at precisely on the hour) and consider picking an hour that's not midnight.

e.g.
```suggestion
          schedule: "39 2 * * *"
```

(unfortunately there isn't yet a way to express "any time plus/minus x minutes" like there is in some other cron implementations)

sengi (approved):

charts/app-config/values-production.yaml
baisa: Thank you @sengi , I have pushed the change and merged 👍

PR #1929, opened 2024-03-28
Disable auto index creation during ES snapshot restore.

sengi:
This prevents the search-api-worker incremental indexer from implicitly re-creating the `govuk` index when it does an insert into Elastisearch in between the search-index-env-sync issuing the `DELETE *` and `POST .../_restore` requests. This has been happening almost daily recently and it messes up the integration environment when it happens.

Tested: run a few restores on the integration cluster in parallel with search-api-worker doing its thing and all attempts succeeded.

Kudos to @lauraghiorghisor-tw @ryanb-gds for analysis and insight that was crucial to identifying this solution 🙇

### DIFF
--- a/charts/search-index-env-sync/snapshot.sh
+++ b/charts/search-index-env-sync/snapshot.sh
@@ -44,6 +44,14 @@ list () {
   $curl "$ES_URL/_snapshot/$SNAPSHOT_REPO/_all" | jq -e '.snapshots'
 }
 
+# Enable or disable automatic index creation on insert.
+# usage: set_auto_create_index true|false
+set_auto_create_index () {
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
I'm not much of a bash scripter, but looks about right to me. Thanks @sengi 🙌

sengi:
Thanks! I'll go ahead and merge so we get this in before the long weekend. Post-merge comments still welcome ofc.

I'll try to keep an eye on the cronjob for the next few days in case there are any lingering issues from, say, the schedule change we made in #1916 (or if there are other issues still to fix).

Hopefully this will be a big improvement though.

Issue #1895, opened 2024-03-11
Eliminate long-lived IAM access keys.


sengi:
A few GOV.UK applications are still using long-lived IAM creds (`AKIA...` access keys) to authenticate to AWS services such as S3. This dates all the back to when GOV.UK was hosted outside AWS in colo facilities. We shouldn't be using these any more.

We've already dealt with most of the easy cases. The remaining ones are probably just the few apps that use [Fog](https://github.com/fog/fog). Fog rolls its own when it comes to authenticating to AWS and doesn't work well with the more modern authentication methods like instance profile and IRSA.

https://github.com/alphagov/content-data-admin/pull/1377 and https://github.com/alphagov/support-api/pull/911 are examples of replacing Fog with the standard AWS client library.

We can then use instance profile creds or IRSA for these apps and get rid of the long-lived keys and the secrets that store them.

theseanything:
Created a PR to remove Fog from specialist-publisher: https://github.com/alphagov/specialist-publisher/pull/2575

theseanything:
Done.

PR #1930, opened 2024-04-04
Remove long-lived AWS creds for specialist publisher

theseanything:
The relevant permissions are now available from the node instance profile. These creds had expired anyway, so the export CSV feature hasn't been working. Rolling this out directly to production as a fix forward.

Part of work to remove long-lived cred: https://github.com/alphagov/govuk-helm-charts/issues/1895


sengi (approved):
🚀

PR #1936, opened 2024-04-08
Add Google Tag Manager to Support

deborahchua:
This allows us to enable basic GA4 tracking for the Support app.

In Integration we also set the `GOOGLE_TAG_MANAGER_AUTH` and `GOOGLE_TAG_MANAGER_PREVIEW` variables so that we can test the configuration before proceeding to Production.

Trello card: https://trello.com/c/owLrfsOP/3473-implement-tracking-in-ga4-for-support-3

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2846,6 +2846,8 @@ govukApplications:
             secretKeyRef:
               name: support-zendesk
               key: ticket_email
+        - name: GOOGLE_TAG_MANAGER_ID
+          value: *publishing-bootstrap-gtm-id
 
   - name: support-api
     helmValues:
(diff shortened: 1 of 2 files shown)


MuriloDalRi (approved):

charts/app-config/values-production.yaml
```diff
@@ -2846,6 +2846,8 @@ govukApplications:
             secretKeyRef:
               name: support-zendesk
               key: ticket_email
+        - name: GOOGLE_TAG_MANAGER_ID
+          value: *publishing-design-system-gtm-id
```
AgaDufrat, line 2850: Support app is not fully using Design System yet. Only the first page was migrated.
```suggestion
          value: **publishing-bootstrap-gtm-id
```

deborahchua: Ah okay this did confused me - thanks @AgaDufrat !

AgaDufrat (approved):

PR #1941, opened 2024-04-08
Remove redundant ingress-class chart.

sengi:
https://github.com/alphagov/govuk-infrastructure/issues/1196


theseanything (approved):

sengi:
Might clean up the places where we're specifying the default class in Ingress resources later.

PR #1946, opened 2024-04-10
Add external secrets for govuk-chat

AlexAvlonitis:
Added seperate files/sercrets for basic-auth, chat-api, postgres, open-ai and rabbitMQ as an initial setup.

RabbitMQ has been added as a placeholder for now in the secrets-manager, will update it once we add the service

Trello card: https://trello.com/c/ewV5ilJ6/1229-initial-terraform-for-integration-environment

### DIFF
--- a/charts/external-secrets/templates/govuk-chat/postgres.yaml
+++ b/charts/external-secrets/templates/govuk-chat/postgres.yaml
@@ -1,0 +1,23 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: govuk-chat-postgres
+  labels:
+    {{- include "external-secrets.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      Credentials for govuk-chat Postgres DB hosted in RDS.
+spec:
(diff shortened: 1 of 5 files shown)


charts/external-secrets/templates/govuk-chat/postgres.yaml
```diff
@@ -0,0 +1,23 @@
...
+  refreshInterval: {{ .Values.externalSecrets.refreshInterval }}
+  secretStoreRef:
+    name: aws-secretsmanager
+    kind: ClusterSecretStore
+  target:
+    deletionPolicy: {{ .Values.externalSecrets.deletionPolicy }}
+    name: govuk-chat-postgres
+    template:
+      data:
+        DATABASE_URL: '{{ $.Files.Get "externalsecrets-templates/postgis-conn-string.tpl" | trim }}/chat'
```
kevindew, line 20: This looks to have a reference to Postgis whereas we'd not be using a Postgis Postgres

AlexAvlonitis: good point, anyway this is a temp placeholder that is going to change once we have actual postgres in place, but i can change this detail now just in case.

rtrinque (approved):
Looks good

PR #1947, opened 2024-04-10
Add Google Tag Manager to Content Data

deborahchua:
This allows us to enable basic GA4 tracking for the Support app.

In Integration we also set the `GOOGLE_TAG_MANAGER_GA4_AUTH` and `GOOGLE_TAG_MANAGER_GA4_PREVIEW` variables so that we can test the configuration before proceeding to Production.

Trello card: https://trello.com/c/NxwdsJb2/3475-implement-tracking-in-ga4-for-content-data-admin-3

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -478,6 +478,8 @@ govukApplications:
         # https://www.github.com/alphagov/govuk-puppet/pull/8041
         - name: GOOGLE_TAG_MANAGER_ID
           value: GTM-NZG8SF2
+        - name: GOOGLE_TAG_MANAGER_GA4_ID
+          value: &publishing-design-system-gtm-id GTM-P93SHJ4Z
         - name: GDS_SSO_OAUTH_ID
           valueFrom:
             secretKeyRef:
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -660,7 +660,7 @@ govukApplications:
         # These GTM values are not secrets (not even the the gtm_auth one).
         # https://github.com/alphagov/govuk-puppet/pull/8041
         - name: GOOGLE_TAG_MANAGER_ID
-          value: &publishing-design-system-gtm-id GTM-P93SHJ4Z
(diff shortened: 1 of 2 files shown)


charts/app-config/values-production.yaml
```diff
@@ -478,6 +478,8 @@ govukApplications:
         # https://www.github.com/alphagov/govuk-puppet/pull/8041
         - name: GOOGLE_TAG_MANAGER_ID
           value: GTM-NZG8SF2
+        - name: GOOGLE_TAG_MANAGER_GA4_ID
+          value: GTM-P93SHJ4Z
```
AgaDufrat, line 482: Non-blocking but could we use the reference to the anchor for consistency with other apps. It also makes it more obvious that they share the ID.
```suggestion
          value: *publishing-design-system-gtm-id
```
Similar for the integration ones.

AgaDufrat (approved):

charts/app-config/values-production.yaml
AgaDufrat: Sorry if my comment is confusing. The anchors are already defined:
- for integration: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-integration.yaml#L658-L665
- for production: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-production.yaml#L658-L661

So we should be able to reference them by using `*`. 

I found this comment useful to get my head around this YAML anchor shenanigans 😃 : https://stackoverflow.com/a/46644785

deborahchua: No worries! This confused me earlier too - those anchors have been defined further down the file and so can't be used if you try and use the alias first. There's some discussion in [this Slack thread](https://gds.slack.com/archives/C013F737737/p1712742344443979)

deborahchua: I've added an additional commit to remove the duplicate anchors - https://github.com/alphagov/govuk-helm-charts/pull/1947/commits/45c55bbcafdd60ea975098ef12f8f8eb520bb51f

Issue #1744, opened 2024-01-18
Migrate away from Bitnami external-dns to official chart


samsimpson1:
We're using the Bitnami external-dns chart at the moment, but their charts repo is huge and contains hundreds of charts, which makes it difficult to see what has changed when updates are released. There is an official external-dns chart, which we could use instead.

https://github.com/kubernetes-sigs/external-dns/tree/master/charts/external-dns

PR #1951, opened 2024-04-11
Use official Kubernetes external-dns chart instead of Bitnami

nimalank7:
Description:
- Official Kubernetes external-dns chart is located [here](https://github.com/kubernetes-sigs/external-dns/tree/master/charts/external-dns)
- Previously in production we were running 3 but the bitnami chart doesn’t have any leader selection so all external-dnspods were running thinking they were the only one. Kubernetes external-dns doesn’t support multiple replicas so removing this entirely. See the discussion [here](https://github.com/kubernetes-sigs/external-dns/issues/2430)
- PR closes https://github.com/alphagov/govuk-helm-charts/issues/1744

### DIFF
--- a/charts/app-config/helm-versions/integration
+++ b/charts/app-config/helm-versions/integration
@@ -2,6 +2,6 @@
 https://kubernetes-sigs.github.io/aws-ebs-csi-driver aws-ebs-csi-driver: "2.29.1"
 https://kubernetes.github.io/autoscaler cluster-autoscaler: "9.36.0"
 https://charts.dexidp.io dex: "0.17.0"
-https://charts.bitnami.com/bitnami external-dns: "7.1.2"
+https://kubernetes-sigs.github.io/external-dns external-dns: "1.14.4"
 https://charts.external-secrets.io external-secrets: "0.9.14"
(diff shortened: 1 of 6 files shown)


charts/app-config/helm-versions/integration
```diff
@@ -2,6 +2,6 @@
 https://kubernetes-sigs.github.io/aws-ebs-csi-driver aws-ebs-csi-driver: "2.29.1"
 https://kubernetes.github.io/autoscaler cluster-autoscaler: "9.36.0"
 https://charts.dexidp.io dex: "0.17.0"
-https://charts.bitnami.com/bitnami external-dns: "7.1.2"
+https://kubernetes-sigs.github.io/external-dns: external-dns "1.14.4"
```
samsimpson1, line 5:
```suggestion
https://kubernetes-sigs.github.io/external-dns external-dns: "1.14.4"
```

samsimpson1 (approved):
Thanks for sorting this, it'll make managing it a lot easier in the future

charts/app-config/helm-versions/integration
nimalank7: Fixed.

PR #1952, opened 2024-04-11
Update deployment strategy for external-dns to rolling update

nimalank7:
Description:
- Use the correct syntax for external-dns deployment strategy


sengi (approved):
How odd that the chart's default overrides the k8s API default for this. Good find!

sengi:
Aha https://github.com/kubernetes-sigs/external-dns/commit/3d343ae

So it looks like it was intentionally set that way to avoid the old and new pods racing each other if there just so happens to be an update for them to perform during the rollout.

We've clearly been getting away with it (and worse!) for ages but I see how `Recreate` being the default makes sense here. What do you reckon?

nimalank7:
Yeah you're right `Recreate` being the default makes more sense here.

PR #1954, opened 2024-04-11
Add Toleration and Node Selector for ARM64 apps

dj-maisy:
## What?
This adds a `appImage.arch` value to our Helm Charts and a condition in the Chart template to add the necessary Pod Node Toleration and Node Selector for scheduling known supported ARM64 images onto Graviton instances. This should also prevent AMD64/x86 images from being scheduled onto ARM nodes and vice-versa.

We then apply this value to the `release` and `transition` apps for testing.

If we wanted to change this in the future so we can schedule/select certain apps to run on specialised GPU/ML hardware, that's something we can add down the line without too much difficulty - but for now, this should keep our app configuration a bit simpler.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1810,7 +1810,8 @@ govukApplications:
 
   - name: release
     helmValues:
-      appImage.arch: arm64
+      appImage:
+        arch: arm64
       dbMigrationEnabled: true
       ingress:
         enabled: true
@@ -2845,7 +2846,8 @@ govukApplications:
 
   - name: transition
     helmValues:
-      appImage.arch: arm64
+      appImage:
+        arch: arm64
       ingress:
         enabled: true
         annotations:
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1810,6 +1810,7 @@ govukApplications:
 
   - name: release
     helmValues:
+      appImage.arch: arm64
```
sengi, line 1813: Does this definitely get transformed into an object `appImage` containing a key `arch`, rather than being passed through as just a single key named `appImage.arch` (which wouldn't come out as `.Values.appImage.arch` in the chart template)?

I dunno if Argo CD does something special here but with regular Helm values I'm pretty sure you'd have to write it as:
```suggestion
      appImage:
        arch: arm64
```

sengi (approved):

charts/app-config/values-integration.yaml
dj-maisy: I did think about this just after I pushed it and thought "that might not work properly".

PR #1961, opened 2024-04-18
Remove kubernetes.io from the taint key.

dj-maisy:
## What?
This removes `kubernetes.io` from the taint key as per our changes to the node taints.

## Related

* https://github.com/alphagov/govuk-infrastructure/pull/1264


sengi (approved):

sengi (commented):
I think you'll also need to change the nodeSelector below

dj-maisy:
> I think you'll also need to change the nodeSelector below

I think the nodeSelector is targeting the built-in label rather than the taint.

sengi:
Oh yeah good point, sorry ignore me 🙃

PR #1964, opened 2024-04-19
Update signon-secrets-sync for Imminence rename.

sengi:
Imminence is being renamed to Places Manager. The job fails if it doesn't find one of the apps named in the config.

Probably makes sense to merge this once the rename is done in prod, rather than try to split the config (though we could if we really have to).

https://trello.com/c/HAEH2bVF/2338-rename-imminence


KludgeKML:
Just to double-check - this is looking for the User-facing name in signon, is it? (in general we're not changing things yet that are in the backend, but we do have to change the name in Signon so that people see the correct name on their dashboard, and the URL so that signon's callback works )

sengi:
Yeah unfortunately the name in Signon is the least worst key we have here :(

KludgeKML:
Hahaha, fair enough. I'll approve. I'm hoping to do the public name change today (or at least in the next few days), I'll add this to the checklist and merge when it's gone to production.

KludgeKML (approved):

PR #1967, opened 2024-04-19
Scale search-api-v2-worker back in.

sengi:
[Max instantaneous CPU utilisation](https://grafana.eks.production.govuk.digital/goto/criZ6j-IR) has been < 8% for the last 30 days.

Scale them back to ~12% of their current replica count, which should still be plenty.


csutter (approved):
We've left this quite high in case we needed to do one off bulk reimports but I think we won't be needing that again for a while, and even so it might be interesting to see how that performs with fewer workers 👍

PR #1975, opened 2024-04-24
Run 30 threads per search-api-v2-worker.

sengi:
Tested: https://github.com/alphagov/search-api-v2/pull/259#issuecomment-2075287311


csutter (approved):
😱

sengi:
Having experimented with this a fair bit over the course of the afternoon, I'm confident about rolling this straight out — especially given we're still running quite a backlog in production at the moment 😄

It's easy enough to roll back too, in case we notice something wonky later on.

PR #1977, opened 2024-04-24
Switch `search-admin` to ARM64 in all envs

csutter:
This has worked a treat in integration 🎉

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2128,6 +2128,7 @@ govukApplications:
 
   - name: search-admin
     helmValues:
+      arch: arm64
       dbMigrationEnabled: true
       workerEnabled: true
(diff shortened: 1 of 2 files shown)


sengi:
I don't think we have the ARM node pool in staging/prod yet but I think there's no reason we can't turn it up at this point? cc @dj-maisy

dj-maisy:
> I don't think we have the ARM node pool in staging/prod yet but I think there's no reason we can't turn it up at this point? cc @dj-maisy

Indeed - so far, I don't think we've encountered _any_ issues in Integration so I'm happy to start adding ARM capability to our other environments.

...but that's something we're going to have to do before we approve/merge this!

theseanything:
I'm going to close this for now until we have ARM node's available in other environments.

PR #1979, opened 2024-04-25
Enable ARM64 for Whitehall in Integration

dj-maisy:
## What?
What it says on the tin. Run the ARM64 flavour of Whitehall in Integration.


sengi (approved):
\its_happening \o/

Issue #1883, opened 2024-03-04
[Epic] Roll out PSS Restricted profile.


sengi:
We [currently enforce](https://github.com/alphagov/govuk-infrastructure/blob/99ef378/terraform/deployments/cluster-services/argo.tf#L35) the [PSS baseline profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/).

We want tighten that up to Restricted where possible (e.g. `apps` namespace), so that we don't have to worry about regressions in container permissions — i.e. application containers unintentionally/unnecessarily being granted system privileges in future.

In other words, this:

- prevents a set of potential misconfigurations in future
- helps to "lock in" the benefits of the least-privilege configuration that we already have

We still have a couple of NFS clients (e.g. asset-manager), so we _might_ need to work around that temporarily and/or pay down that tech debt and switch them to S3.

Helm charts to modify:

- [x] Licensify
- [x] apps
- [x] ckan
- [x] datagovuk
- [x] dgu-shared

sengi:
Ah nuts, auto-closed this by mistake somehow or other.

sengi:
It's not rolled out until we're in enforcement mode.

Should be mostly just trivial template fixes, plus the not-so-trivial NFS question.

nicholsj:
@nimalank7 will look at documenting this, so people understand what they need to do if their work isn't compliant.

jasonBirchall:
@nimalank7, are you happy this is done?

PR #1995, opened 2024-05-01
Add securityContext from cron-jobs containers

theseanything:
These containers currently violate the PodSecurity.

As part of tightening up restriction: #1883


sengi (approved):
LGTM! thanks for fixing!

PR #1998, opened 2024-05-01
Revert "Disable assets upload for govuk-chat"

AlexAvlonitis:
Reverts alphagov/govuk-helm-charts#1997

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1217,7 +1217,7 @@ govukApplications:
         hosts:
           - name: chat.{{ .Values.k8sExternalDomainSuffix }}
       uploadAssets:
-        enabled: false
+        path: /app/public/assets/chat
       extraEnv:
         - name: BASIC_AUTH_USERNAME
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1216,8 +1216,6 @@ govukApplications:
             ]}}]
         hosts:
           - name: chat.{{ .Values.k8sExternalDomainSuffix }}
-      uploadAssets:
-        enabled: false
```
kevindew, line 1220: I think we'll need this to be:

```
uploadAssets:
  path: /app/public/assets/chat
```

so it matches: https://github.com/alphagov/govuk-chat/pull/169/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9R44

AlexAvlonitis: Done 👍

kevindew (approved):
Change looks good, I think you should just edit the commit message to reflect what was changed before merging.

PR #1999, opened 2024-05-01
Changed the s3Directory name for govuk-chat

AlexAvlonitis:
Since we have specified an asset path of '/chat' in govuk-chat, we have to manually specify the s3Directory to be `chat`, otherwise it defaults to the repo name: `govuk-chat`


kevindew (approved):
looks good to me, it doesn't look like it needs any prefix from what I can see.

PR #2004, opened 2024-05-02
Add GDS-SSO OAuth credentials for govuk-chat

AlexAvlonitis:
trello card: https://trello.com/c/k8KTyMCx/1413-add-gds-sso-to-integration-environment

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1264,6 +1264,16 @@ govukApplications:
             secretKeyRef:
               name: govuk-chat-rabbitmq
               key: RABBITMQ_URL
+        - name: GDS_SSO_OAUTH_ID
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-chat
+              key: oauth_id
+        - name: GDS_SSO_OAUTH_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-chat
+              key: oauth_secret
         - name: REDIS_URL
           value: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
 


charts/app-config/values-integration.yaml
```diff
@@ -1260,6 +1260,16 @@ govukApplications:
             secretKeyRef:
               name: govuk-chat-rabbitmq
               key: RABBITMQ_URL
+        - name: GDS_SSO_OAUTH_ID
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-oauth
+              key: GDS_SSO_OAUTH_ID
```
kevindew, line 1267:
```suggestion
              name: signon-app-govuk-chat
              key: oauth_id
```

charts/app-config/values-integration.yaml
```diff
@@ -1260,6 +1260,16 @@ govukApplications:
...
+        - name: GDS_SSO_OAUTH_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-oauth
+              key: GDS_SSO_OAUTH_SECRET
```
kevindew, line 1272:
```suggestion
              name: signon-app-govuk-chat
              key: oauth_secret
```

charts/external-secrets/templates/govuk-chat/oauth.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: external-secrets.io/v1beta1
```
kevindew, line 1: It doesn't look like other apps need this for signon credentials. I'm not sure how they get them but I have vague memories of there being a system that auto rotates them.

kevindew (commented):
It looks like there's a bit of a pre-existing convention about signon credentials are managed. I've made some suggestions but I don't quite understand how it works so I'm not sure I've given you the right suggestions.

kevindew:
Asked in: https://gds.slack.com/archives/C013F737737/p1714645006120649

charts/app-config/values-integration.yaml
AlexAvlonitis: The name shouldn't change as i think it's a convention to use `app-name + secret` https://docs.publishing.service.gov.uk/kubernetes/manage-app/manage-secrets/#defining-an-externalsecret-in-kubernetes 

but I can change the key name as you suggested 👍

kevindew: It's just that all the other apps have ones prefixed with signon-app as per: https://github.com/alphagov/govuk-helm-charts/blob/2faad1e20782fa4a567a395dbd3e0ad898b85375/charts/app-config/values-integration.yaml#L122-L131

charts/external-secrets/templates/govuk-chat/oauth.yaml
AlexAvlonitis: that's what i noticed too, I only noticed account api using it. Nice thanks for notifying the team

charts/app-config/values-integration.yaml
AlexAvlonitis: hmm good point, they are getting them from somewhere else

sengi:
Yeah just follow the naming of the existing ones and you should be fine.

sengi:
You'll need to add the app to the list here: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/templates/signon-secrets-sync-configmap.yaml

There's probably nothing stopping us dropping that file now, though. I think we only introduced it because of the number of defunct applications in Signon that hadn't been cleaned up at the time. It's much tidier now so just importing everything is probably a good option.

kevindew:
As per the Slack chat I'm going to go and create the application in Signon in each environment before we proceed with merging

kevindew:
> > As per the Slack chat I'm going to go and create the application in Signon in each environment before we proceed with merging
> 
> Added one in integration already, just double check about the naming conventions etc.. so we could use same names in all the ENVs

Ok cool - I'll create the other ones and adjust integration if I have to for any consistency reasons

charts/app-config/templates/signon-secrets-sync-configmap.yaml
```diff
@@ -7,6 +7,7 @@ data:
     [
       "Account API",
       "Asset Manager",
+      "Chat",
```
AlexAvlonitis, line 10: Added this one for now, if we choose a different name I will change it again

kevindew: We won't need this when https://github.com/alphagov/govuk-helm-charts/pull/2006 is merged (which I imagine will be in the next hour).

Thanks for adding it though.

kevindew:
Looks like it's merged now 🎉 

I've created the signon entry in prod, it'll sync to staging tonight. We won't wait for that before merging though (as far as I can tell, from Chris' changes signon won't error without Chat).

I imagine we will need to run a rake task in signon to get these secrets populated. I'm not sure on the automation schedule @sengi do you know?

kevindew (approved):
Great, looks good to me 🤞 that it all works.

AlexAvlonitis:
> These values won't have `govuk-` as a prefix because they're based on the signon name, which is just chat.
> 
> I think with that changed this will be good to be merged.

Done, thanks for the reviews 👍

PR #2006, opened 2024-05-02
Remove disused config for Signon secrets sync job.

sengi:
As of https://github.com/alphagov/signon/pull/2855 we just import all non-retired Doorkeeper application secrets and all non-suspended API user keys, so that we don't have to maintain a separate config file.


kevindew:
Great thanks for this. Looks good to me but I'll leave approving to someone with better codebase knowledge than myself

PR #2008, opened 2024-05-02
Switch off govuk-mirror-job in staging temporarily

theseanything:
During the same time period in which the govuk-mirror job runs, the post-sync workflow is triggered. This is to test if this might be due to the load from the govuk-mirror job. There's potential that extra load is causing pods to be terminated, which causes reconcilation activity in Argo CD, which then causes the post-sync workflow to be triggered.


dj-maisy (approved):
🛌

PR #2011, opened 2024-05-03
Adds published-documents-consumer worker to govuk-chat

AlexAvlonitis:
trello card: https://trello.com/c/VBIfPA9y/1416-hook-up-rabbitmq-in-aws


kevindew (commented):
Looks good, but do we also need Rabbit MQ env vars?

kevindew (approved):

PR #2013, opened 2024-05-03
Temporarily rolling govuk-chat backto AMD/x86

dj-maisy:
## What?
This rolls the GOV.UK Chat app back to AMD64 (x86-64) in Integration due to a hash mismatch when precompiling assets on the two different machine architectures.

FYI @kevindew.


kevindew (approved):
Thanks for this

PR #2016, opened 2024-05-07
Prevent post-sync deployments when nothing changes

theseanything:
This sets a condition to prevent triggering the post-sync workflow when there is no configuration changes. The post-sync notification was being sent when an Application was being sync, even if there wasn't a configuration change (due to a change in the cluster).


sengi (approved):
Good catch!

PR #2020, opened 2024-05-07
Make assets-upload job ARM supported...

dj-maisy:
## What?
This adds the necessary bits to get the `-upload-assets` jobs running on ARM. Hopefully alongside their deployments.

This also moves the "arch" label to _helpers so we don't have to specify it for every "generic GOV.UK app".


sengi (approved):
Looks like we're also missing this from templates/static-error-page-upload-job.yaml?

samsimpson1 (approved):

PR #2030, opened 2024-05-08
Remove basic-auth for govuk-chat

davidgisbey:
## Description

Since we're now using SignOn for auth on our integration environments we no longer need the basic-auth secret so this can be removed.

## Trello card

https://trello.com/c/sInvpIk7/1420-require-signon-access-for-the-chat-in-govuk-environments


AlexAvlonitis (approved):
looks good, if you don't mind removing them from the aws secrets manager as well (integration)

PR #2032, opened 2024-05-09
Update image name for static error job

dj-maisy:
## What?
This is a small one - also updating the Static Error Job page to use the new/correct toolbox image name.


sengi (approved):
Thanks!

PR #2036, opened 2024-05-10
Enable govukchat-worker

AlexAvlonitis:
- Opensearch is now available/reachable in integration, so this worker can be enabled again


rtrinque (approved):
looks good

PR #2037, opened 2024-05-14
Start removing Learning to Rank bits

sihugh:
We've already [disabled Learning to Rank from serving any queries](https://github.com/alphagov/search-api/pull/2905), so we can start removing the things that affect or invoke it.

It's safe to remove the `ENABLE_LTR` environment variable, as [Search-API checks for presence](https://github.com/alphagov/search-api/blob/f2950323f8276b4d6cc4d02fad44521ba12bd20e/lib/search/relevance_helpers.rb#L3).

We no longer need to [generate new LTR models](https://github.com/alphagov/search-api/blob/f2950323f8276b4d6cc4d02fad44521ba12bd20e/lib/tasks/learn_to_rank.rake#L183-L199).

We'll be back to remove the Tensorflow endpoint and the service account soon.


sengi (approved):
Thanks!

PR #2055, opened 2024-05-17
Comment out queue worker for Chat

kevindew:
This is to stop this process as it seems to be caught in an infinite error loop and I'm not sure of a better way to stop the process.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1219,8 +1219,8 @@ govukApplications:
       workers:
         - command: ["sidekiq", "-C", "config/sidekiq.yml"]
           name: worker
-        - command: ['rake', 'message_queue:published_documents_consumer']
-          name: published-documents-consumer
+        # - command: ['rake', 'message_queue:published_documents_consumer']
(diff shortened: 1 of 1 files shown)


peteglondon (approved):

sengi (approved):

sengi (commented):
You could alternatively set workerReplicas: 0 but either is fine.

kevindew:
> You could alternatively set workerReplicas: 0 but either is fine.

Thanks for the tip, is this something that is always done through helm charts? I was trying to work out if I could do it via the Argo UI.

sengi:
Yes — Git should be authoritative for pretty much all GOV.UK's configuration these days, at least that's [the theory](https://www.redhat.com/en/topics/devops/what-is-gitops#what-is-gitops)!

kevindew:
Super, makes a lot of sense.

Final question, as we wanted to just kill one of the worker processes and not both. I imagine that can't be done with `workerReplicas: 0` as that'd stop both right?

sengi:
Oh, in that case you could simply delete the pod.

PR #2062, opened 2024-05-20
Re-enable Chat MQ consumer

kevindew:
Now that we have extra RAM on the OpenSearch instance we believe we can now resume running this worker and not experience errors.


rtrinque (approved):
Looks good

PR #2073, opened 2024-05-23
Increase ClamAV maximum number of files allowed

brucebolt:
In https://github.com/alphagov/govuk-helm-charts/pull/1448/files, we increased the limits for files scanned by ClamAV in Asset Manager.

We have since had a user upload a PDF which (once expanded) has more than 20,000 files.

Therefore increasing the maximum limit to a size that accepts this document.  I've tested this manually on a Asset Manage worker to ensure this limit is sufficient for the file.

[Trello card](https://trello.com/c/ZSZlKf4x)


JonathanHallam (approved):
thanks :)

PR #2075, opened 2024-05-23
Change redis URL for the chat app to the dedicated redis cluster

rtrinque:
In order to reduce the blast radius of the GOV.UK chat app we now use a dedicated redis cluster for it.


ianhowell-gds (approved):
LGTM

PR #2088, opened 2024-06-05
Manage Tempo with Argo

theseanything:
This shifts the installing the Tempo chart from Terraform to ArgoCD.

### DIFF
--- a/charts/monitoring-config/tempo-values.yaml
+++ b/charts/monitoring-config/tempo-values.yaml
@@ -1,0 +1,22 @@
+reportingEnabled: false
+ingester:
+  - persistence:
+      - enabled: true
+        size: 30Gi
+        storageClass: ebs-gp3
+metaMonitoring:
+  - serviceMonitor:
+      - enabled: true
+        namespace: monitoring
+traces:
+  - otlp:
+      - grpc:
+          - enabled: true
+        http:
+          - enabled: true
+overrides:
+  - '*':
+      - ingestion_burst_size_bytes: 20000000
+        metrics_generator_processors:
+          - service-graphs
+          - span-metrics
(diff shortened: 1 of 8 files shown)


charts/monitoring-config/tempo-values.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - persistence:
+      - enabled: true
+        size: 30Gi
+        storageClass: ebs-gp3
+metricsGenerator:
+  - enabled: true
+    config:
+      - storage:
+          - remote_write:
+              - url: "http://kube-prometheus-stack-prometheus:9090/api/v1/write"
```
sengi, line 12: (no action needed on this for this PR)

(You probably already got this, but) re what you mentioned in standup about writing to all pods in the statefulset, I think this might be the place! I'd guess you want something like
```suggestion
              - url: "http://prometheus-kube-prometheus-stack-prometheus-0.kube-prometheus-stack-prometheus:9090/api/v1/write"
              - url: "http://prometheus-kube-prometheus-stack-prometheus-1.kube-prometheus-stack-prometheus:9090/api/v1/write"
```

(though might need some annoying Helm templatery in order to deal with the number of replicas varying by environment)

sengi (approved):

charts/monitoring-config/tempo-values.yaml
sengi: (I dunno if those are the exact names btw but should be something along those lines)

sengi: ah no need to template-fu, I see there's already a `tempo-values-{integration,staging,production}.yaml` that should do.

PR #2089, opened 2024-06-05
Re-enable tracing in production

theseanything:
This was previously disable due to a bug causing the OpenTel SDK to produce significant numbers of error logs. This bug has been fixed in https://github.com/alphagov/govuk_app_config/pull/372.


richardTowers (approved):
Do we need to wait until your new govuk_app_config has been deployed to all the apps? Or is it okay to remove this ahead of time?

theseanything:
> Do we need to wait until your new govuk_app_config has been deployed to all the apps? Or is it okay to remove this ahead of time?

Yeah - not going flip this switch until dependabot PRs are pushed out tomorrow.

PR #2091, opened 2024-06-06
(WIP) Start on a ClamAV Sidecar

dj-maisy:
## What?
This is a WIP for a ClamAV Sidecar so we can de-couple ClamAV from the Asset Manager image. Happy for feedback/constructive criticism/etc.


dj-maisy:
We're not going to worry about this for now - we're going to build a native version of ClamAV into the container instead.

See: https://github.com/alphagov/asset-manager/pull/1400

Issue #2098, opened 2024-06-11
Redis instances in the k8s cluster


samsimpson1:
We are transitioning away from the shared Redis cluster for the Sidekiq queues that a lot of GOV.UK apps use, and we have decided to run these separate redis servers in the k8s cluster. We need a way to create redis servers for all the apps that need them.

samsimpson1:
* Chart changes have been made to allow for creating redis instances in k8s
* Basic monitoring is in place which will ensure these instances don't run out of memory or disk space
* The publishing team has started to deploy redis instances in all environments

PR #2099, opened 2024-06-11
Add ability to create redis instances in generic-govuk-app chart

samsimpson1:
Adds the ability to create redis servers in the generic-govuk-app chart, and creates a test Redis server for link-checker-api in integration

#2098

### DIFF
--- a/charts/generic-govuk-app/values.yaml
+++ b/charts/generic-govuk-app/values.yaml
@@ -20,6 +20,27 @@ workers:
   - command: ["sidekiq", "-C", "config/sidekiq.yml"]
     name: worker
 
+# redis.enabled controls whether a Redis database is set up for the app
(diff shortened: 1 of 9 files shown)


charts/generic-govuk-app/values.yaml
```diff
@@ -14,12 +14,33 @@ replicaCount: 2
 podDisruptionBudget: {}
 
 # workerEnabled should be true if the app uses Sidekiq, false otherwise.
-workerEnabled: false
+workerEnabled: true
```
theseanything, line 17: How come we're changing this default? Do we need to update the app values where using the default?

samsimpson1: Oh my bad, I changed it during testing and forgot to change it back

theseanything (approved):

PR #2105, opened 2024-06-18
Add the Arch selection stuff to Asset Manager

dj-maisy:
## What?
We're working on getting Asset Manager working with ARM. So let's get the templates updated to support this.

### DIFF
--- a/charts/asset-manager/templates/deployment.yaml
+++ b/charts/asset-manager/templates/deployment.yaml
@@ -164,3 +164,13 @@ spec:
             preStop:
               exec:
                 command: ["sleep", "15"] # To allow time for ALB to deregister pod before termination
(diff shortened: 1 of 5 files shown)


charts/asset-manager/templates/deployment.yaml
```diff
@@ -164,3 +164,13 @@ spec:
...
+      {{- if eq "arm64" .Values.arch }}
+      tolerations:
+        - key: arch
+          operator: Equal
+          value: {{ .Values.arch }}
+          effect: NoSchedule
+      nodeSelector:
+        kubernetes.io/arch: {{ .Values.arch }}
+      {{- end }}
+
```
sengi, line 176: (no action needed) something weird going on with the line endings here 🤔😅

sengi (approved):
🚀

PR #2107, opened 2024-06-20
Perform environment sync operations at the same time

brucebolt:
For Publishing API and Content Store, we need their databases to be in sync with each other when the backup is taken and restored during the environment sync.

If we don't, we can get data conflicts that stop content being published to Content Store.

Therefore bringing backups for Publishing API and Content Store to be at the same time, and the same for the restore jobs.


sengi:
Surely running them at the same time doesn't actually solve the race condition? Can you describe the problem more specifically (perhaps with an example) so that we can understand the rationale for the change and figure out whether it's correct?

PR #2110, opened 2024-06-20
Give Freshclam more memory - it got reaped

dj-maisy:
## What?
With the tight limit of 200Mi on the Freshclam spec, it was getting reaped by the OOM Killer. Let's double the request and give it a limit of 1Gi to see if this is enough breathing room.

### DIFF
--- a/charts/asset-manager/values.yaml
+++ b/charts/asset-manager/values.yaml
@@ -68,10 +68,10 @@ workerResources:
 freshclamResources:
   limits:
     cpu: 1000m
-    memory: 200Mi
+    memory: 1Gi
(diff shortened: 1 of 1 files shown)


theseanything (approved):

sengi (commented):
Oh there's a setting to make it not try to load the entire database into memory to "test" it. Just disable that and it should use hardly any RAM. (The test it does is useless in our use case because clamd will complain if there's something wrong with the database anyway.)

sengi:
`TestDatabases disabled` in `freshclam.conf` is what you want.

PR #2123, opened 2024-06-25
Write timestamp to last-updated.txt for GCS mirror

theseanything:
This fixes the govuk_mirror_last_updated_time metric for GCS mirrors. Google Storage Transfer copies files from our S3 Mirror bucket, however ignores files where content hasn't changed. This updates the last-updated.txt file to have changing content.

### DIFF
--- a/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
+++ b/charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
@@ -114,5 +114,5 @@ spec:
                   /s5cmd sync --size-only "/data/${ASSETS_DOMAIN}" "s3://${BUCKET_NAME}/"
 
                   # Upload file for checking mirror freshness
-                  touch "/data/${WWW_DOMAIN}/last-updated.txt"
+                  date "+%Y-%m-%d %H:%M:%S" > "/data/${WWW_DOMAIN}/last-updated.txt"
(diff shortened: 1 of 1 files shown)


samsimpson1 (approved):

charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
```diff
@@ -114,5 +114,5 @@ spec:
                   /s5cmd sync --size-only "/data/${ASSETS_DOMAIN}" "s3://${BUCKET_NAME}/"
 
                   # Upload file for checking mirror freshness
-                  touch "/data/${WWW_DOMAIN}/last-updated.txt"
+                  date "+%Y-%m-%d %H:%M:%S" > "/data/${WWW_DOMAIN}/last-updated.txt"
```
sengi, line 117: optional: might as well use a standard format and include the tz
```suggestion
                  date --iso-8601=seconds > "/data/${WWW_DOMAIN}/last-updated.txt"
```

sengi (approved):

charts/govuk-jobs/templates/govuk-mirror-sync-cronjob.yaml
sengi: ah I was too slow haha

theseanything: https://github.com/alphagov/govuk-helm-charts/pull/2124

PR #2130, opened 2024-06-27
Update Smokey templates for ARM64, add labels

dj-maisy:
## What?
This aims to add the necessary ARM64 stuff to Smokey, including the arch labels to make ARM-powered pods/jobs easier to identify.

I have also tried to add this to the Argo `WorkflowTemplate` but haven't needed to touch these up to now, so would appreciate an extra thorough review of this part!

### DIFF
--- a/charts/smokey/templates/_helpers.tpl
+++ b/charts/smokey/templates/_helpers.tpl
@@ -22,7 +22,7 @@ helm.sh/chart: {{ include "smokey.chart" . }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
-app.kubernetes.io/arch: {{ default "amd64" .Values.arch }}
+app.kubernetes.io/arch: {{ .Values.arch }}
 {{- end }}
 
 {{/*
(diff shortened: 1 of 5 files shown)


charts/smokey/templates/cronjob.yaml
```diff
@@ -3,7 +3,12 @@
 apiVersion: batch/v1
 kind: CronJob
 metadata:
-  name: {{ .name }}
+  name: {{- include "smokey.name" . -}}
```
sengi, line 6: Does this produce valid output with the `-}}` (trim whitespace from right-hand side) there? Haven't run it myself but I suspect it may elide the newline and produce something like `name: smokeylabels:` (not sure tho, just caught my eye as a bit sus)

charts/smokey/templates/_helpers.tpl
```diff
@@ -0,0 +1,33 @@
...
+Common labels
+*/}}
+{{- define "smokey.labels" -}}
+helm.sh/chart: {{ include "smokey.chart" . }}
+{{ include "smokey.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/arch: {{ default "amd64" .Values.arch }}
```
sengi, line 25: Better without the `default` here because you already have a default in the chart's values.yaml. Redundant defaults can lead to painful debugging 😅

sengi (approved):

PR #2142, opened 2024-07-02
Add MIRROR_AVAILABILITY_URL for govuk-exporter

theseanything:
This specifies which URL the exporter should use to monitor the availability of the mirrors from.


dj-maisy (approved):
🚀

PR #2147, opened 2024-07-03
Add a healthcheck file for static assets

nimalank7:
Description:
- Future PR will configure Fastly's backend for static S3 assets (those served by [www.gov.uk/assets/*](http://www.gov.uk/assets/*)) to point directly to the S3 bucket holding the static assets
- In order for this it requires a probe to poll an endpoint. Currently there is no suitable file on S3 so instead add a blank healthcheck.html file into the S3
govuk-app-assets-`${GOVUK_ENVIRONMENT}` bucket so that if that page can't be loaded Fastly knows to switch to the mirror
- See https://github.com/alphagov/govuk-fastly/issues/73

### DIFF
--- a/charts/generic-govuk-app/upload-static-error-pages.sh
+++ b/charts/generic-govuk-app/upload-static-error-pages.sh
@@ -8,5 +8,6 @@ OUTPUT_PATH=/tmp/output
 mkdir -p "${OUTPUT_PATH}"
 cd "${OUTPUT_PATH}"
 curl --fail-early -fo '#1.html' "${SERVICE}/templates/{${ERROR_PAGES_COMMA_SEPARATED}}.html.erb"
+touch healthcheck.html
 eval ls "{$ERROR_PAGES_COMMA_SEPARATED}.html" || (echo Failed to download one or more files.; exit 1)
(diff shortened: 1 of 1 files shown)


sengi:
I'm unclear on what problem this is trying to solve — surely we don't gain anything from having Fastly probe S3? (given the probabilities involved)

sengi:
If your aim is to have Fastly try to fetch from S3 and fall back to something else if it fails, you can do that much more effectively without a healthcheck probe (see our existing config for a working example).

nimalank7:
I'm following the configuration we have for our S3 mirrors buckets in `www` as they all have probes attached. (Similarly with `F_origin`). I think the same reasoning for them applies to here as well no?

sengi:
The other Fastly services definitely don't probe HTTP. They just check DNS. That's important and intentional; otherwise we hammer the backend with (legacy) probe traffic. The failover is per request rather than based on healthchecks. See `.dummy` here: https://www.fastly.com/documentation/reference/vcl/declarations/backend/#syntax

nimalank7:
I tested this out without the healthcheck so I'm closing down this PR.

PR #2148, opened 2024-07-03
Give Smokey WF same as Jobs to prevent OOM Reaping

dj-maisy:
## What?
This gives the Smokey job triggered by Argo Workflows the same resource limits as the other CronJob template in the hopes of resolving the "OOMKilled" errors.


sengi (approved):
Can we get `resources` from a Helm value or something so they can't be out of sync in future?

PR #2151, opened 2024-07-04
Try JSON PodSpecPatch as per Argo Docs

dj-maisy:
## What?
After struggling to fix the OOMKilled errors, it looks like some examples point towards using JSON-formatted PodSpecPatches in the Argo Docs and community. Also try adding one of these patches to the `smoke-test` DAG template.


sengi (approved):
https://giphy.com/gifs/halloween-season-4-parks-and-recreation-3o6ZthnDqxfAOFIhdC

PR #2158, opened 2024-07-07
Reset Content Store `payload_version` to 0 in non-prod envs

ChrisBAshton:
Occasionally, Content Store and Publishing API can fall out of sync in non-production environments after the environment sync process has completed. Content Store can be a day 'ahead' of Publishing API.

Consequently, any publish events from Publishing API in the affected environment will be rejected by Content Store, if the `Event.maximum_id` in Publishing API is behind the `payload_version` of the `ContentItem` being updated in Content Store.

The pragmatic fix is to reset the `payload_version` in Content Store so that it can never be ahead of Publishing API.

### DIFF
--- a/charts/db-backup/scripts/content-store.sql
+++ b/charts/db-backup/scripts/content-store.sql
@@ -1,0 +1,17 @@
+-- sqlfluff:dialect:mysql
+
+-- Occasionally, Content Store and Publishing API can fall out of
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Looks good to me 👍 (with a caveat that I've not reviewed any SQL in Helm Charts PRs before)

brucebolt (approved):

PR #2159, opened 2024-07-08
Add new Helm chart to create index snapshots of the Chat Opensearch clusters

ianhowell-gds:
This is to configure daily snapshots of the Chat Opensearch clusters and to sync them from Production to Staging to Integration.


sengi:
This is a lot of duplicated code. Is it not feasible to reuse the existing chart and its scripts?

rtrinque (approved):
Looks good

PR #2164, opened 2024-07-09
Revert PR https://github.com/alphagov/govuk-helm-charts/pull/2159:

ianhowell-gds:
This PR reverts the previous PR2159 which was to add new Helm chart to create index snapshots of the Chat Opensearch clusters and sync them from Production to Staging to Integration.

This reverts commit 13c143572443895abf29a921b494385b45319669.

The reason for this is that it has been suggested a better way for this would be to refactor the existing Helm chart to be able to work with multiple clusters of differing types (ElasticSearch & OpenSearch).


rtrinque (approved):
Looks good

PR #2170, opened 2024-07-09
Fix wonky crontab spec.

sengi:
Oops, sorry about the extra noise.


theseanything (approved):

sengi:
Bit disappointing that the kubeconform schemas still don't catch this. Oh well.

PR #2171, opened 2024-07-09
Actually scrape pushgateway.

sengi:
Seems we weren't scraping it before. Didn't show up under `/targets`.

Bit odd though. Suspect I'm still missing something cos seems unlikely this would have been set up without attempting to use it. Let's see if this helps.


sengi:
Aha we just missed that serviceMonitor value when porting it over from TF. Mystery solved. (I guess it wasn't being used much back then if nobody noticed!)

samsimpson1 (approved):

PR #2176, opened 2024-07-10
Remove workflow step for "deployed-to-" tags on images

theseanything:
This fails for initial deployments as the image will not have been pulled through from ghcr.io to ECR before tagging it. Since moving to multi-arch images and manifests, the "deployed-to-" doesn't protect the referenced images from being deleted by the ECR lifecycle rules.

Removing for now as will just show up as errors in Argo Workflows. Probably should replace with a cronjob that checks what images are in use by the cluster, and apply tags appropriately.


sengi (approved):
Yeah, I guess the deploy-to- tags are becoming less useful anyway. Maybe we don't need them?

theseanything:
> Yeah, I guess the deploy-to- tags are becoming less useful anyway. Maybe we don't need them?

Yeah, I guess at them moment we'll always have a copy in ghcr.io. I need to take a look at how we clean up images there too.

PR #2177, opened 2024-07-11
Refactor `search-index-env-sync` to work with multiple clusters

ianhowell-gds:
Refactor `search-index-env-sync` Helm chart to work with multiple clusters of differing types (ElasticSearch and OpenSearch). This is required as there are now 2 clusters (`blue-elasticsearch6-domain` and `chat-engine`) and both of them need to have their indices replicated from Production to Staging and Integration each night.


rtrinque (approved):
Looks good

sengi (approved):
Nice! Thanks for tidying up the config to work the same as that similar db-backup chart ❤️

PR #2183, opened 2024-07-17
Amend `snapshot.sh` to fix unbound variable error

ianhowell-gds:
Amend `snapshot.sh` to fix unbound variable error when `SEARCH_USERNAME` and `SEARCH_PASSWORD` are not assigned and modify the double quotes around the variables to allow a colon to be in the password.


rtrinque (approved):
Looks good to me

PR #2186, opened 2024-07-17
Switch integration link-checker-api to use new Redis instance

brucebolt:
This switches the web application in integration away from the shared Redis instance to a new dedicated instance for this app.

A later PR will switch the workers, once all the existing jobs have been processed.

[Trello card](https://trello.com/c/GAj8vwOU)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1415,8 +1415,7 @@ govukApplications:
       redis:
         enabled: true
         redisUrlOverride:
-          app: &link-checker-redis redis://shared-redis-govuk.eks.integration.govuk-internal.digital
-          workers: *link-checker-redis
+          workers: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
       extraEnv:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1415,8 +1415,7 @@ govukApplications:
       redis:
         enabled: true
         redisUrlOverride:
-          app: &link-checker-redis redis://shared-redis-govuk.eks.integration.govuk-internal.digital
-          workers: *link-checker-redis
+          app: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
```
richardTowers, line 1418: Your comment says "switches the web application in integration away from the shared Redis", but this looks like the `app` still points at shared redis, and it's the `workers` that are being switched over.

Is it the comment or the code that's wrong? Or am I misunderstanding?

brucebolt: I pushed the wrong branch 🤦‍♂️.  Have corrected that now.

theseanything (approved):

PR #2188, opened 2024-07-17
Set up Discovery Engine secret for `search-admin`

csutter:
We will soon want to wire up `search-admin` to GCP Discovery Engine ("Vertex AI Search") in the same way we have done for `search-api-v2`. This integration will allow Search Admin to modify some resources on Discovery Engine to allow administrative users from the wider GOV.UK search team to make some modifications to configuration through the Search Admin UI.

- Add external secrets template for `search-admin` Discovery Engine configuration
- Update app config values to retrieve secrets as environment variables (in integration only for now to test it out)


csutter:
The secrets have already been created in `search-v2-infrastructure`: https://github.com/alphagov/search-v2-infrastructure/blob/main/terraform/environment/discovery_engine.tf#L77

samsimpson1 (approved):

PR #2201, opened 2024-07-24
Modify `curl` command in `snapshot.sh` for `search-index-env-sync`

ianhowell-gds:
The way the `curl` command currently works in the script, both username and password for the basic authentication go into the cronjob log. This change aims to fix that by using a config file rather than having the authentication directly in the command.


rtrinque (approved):
Looks good

PR #2215, opened 2024-07-25
Set curl command with alias in `snapshot.sh`

ianhowell-gds:
Using environment variables in a config file for the curl command did not work as the env vars are not expanded when the command is run. Switching to using a function for curl instead to prevent authentication credentials being logged by the cronjob.


rtrinque (approved):
Looks good

PR #2217, opened 2024-07-26
Turn off debugging for authentication in `snapshot.sh`

ianhowell-gds:
With debugging turned on, the username and password authentication for the Opensearch API are logged in clear text. This change aims to prevent this happening when the `curl` command is run.


rtrinque (approved):
Looks good

PR #2219, opened 2024-07-26
Add secret for Notify API key in integration

jackbot:
Stores the API key to allow GOV.UK Chat to send emails via Notify.

Just configured in integration now while we verify it works; will be
rolled out to staging/production at a later stage.

The secret has already been added in Secrets Manager.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1260,6 +1260,11 @@ govukApplications:
             secretKeyRef:
               name: govuk-chat-postgres
               key: DATABASE_URL
+        - name: GOVUK_NOTIFY_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-notify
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1260,6 +1260,11 @@ govukApplications:
             secretKeyRef:
               name: govuk-chat-postgres
               key: DATABASE_URL
+        - name: NOTIFY_API_KEY
```
kevindew, line 1263:
```suggestion
        - name: GOVUK_NOTIFY_API_KEY
```

Sorry nit pick comment: looking at the rest of this file there seems to be a convention to prefix with GOVUK_ before NOTIFY env vars.

jackbot: Cheers @kevindew, have updated it to be consistent with the others.

kevindew (approved):
Thank you ⭐

PR #2221, opened 2024-07-26
Add template for Notify API key

jackbot:
The env var was added to the integration values file but this template
file is needed to map the value to the AWS secrets manager secret.


kevindew (approved):
Nice one 👍

PR #2228, opened 2024-07-26
Add Notify values for GOV.UK chat staging/production

jackbot:
Configures both environments as integration has been configured now we
know everything works.

Both API keys have been added in Secrets Manager.


davidgisbey (approved):
lgtm

PR #2231, opened 2024-07-31
Remove CHAT_API_* env vars from GOV.UK chat

jackbot:
We no longer communicate with this API so the env vars aren't needed.


kevindew (approved):
Nice one

PR #2246, opened 2024-08-05
Fix display of search quality monitoring alerts

csutter:
Turns out `mul` isn't available here causing an error, and we don't really have much by way of options to turn the proportion into a percentage, so just show it raw instead.


dj-maisy (approved):
🚀

PR #2248, opened 2024-08-05
Provision standalone Redis instance in staging for link-checker-api

brucebolt:
This creates the standalone instance of Redis in staging for Link Checker API.

Later commits will switch the application away from the shared instance.

[Trello card](https://trello.com/c/eqvVgy68)

### DIFF
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -1445,6 +1445,11 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &link-checker-redis redis://shared-redis-govuk.eks.staging.govuk-internal.digital
+          workers: *link-checker-redis
       extraEnv:
         - name: GDS_SSO_OAUTH_ID
           valueFrom:


theseanything (approved):

charts/app-config/values-staging.yaml
```diff
@@ -1445,6 +1445,11 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &link-checker-redis redis://shared-redis-govuk.eks.staging.govuk-internal.digital
```
Gweaton, line 1451: The name `shared-redis...` suggests this isn't the standalone instance but the shared one, am I misunderstanding what this is doing?

brucebolt: This is the URL of the old (shared) Redis instance.  Setting these values allows us to override which instance the app and/or worker is using.  In this case, I'm continuing to use the old shared Redis until a later PR.

Gweaton: Ah, I thought that might be the case but was slightly thrown by the commit message's `This creates the standalone instance of Redis in staging for Link Checker API.`, which I read as it creating the new instance (rather than using the old shared one). Thanks for explaining 👍

brucebolt: Adding the `enabled: true` provisions the standalone instance and sets default URLs to point the app to the new Redis.  So we need to override those at the same time as provisioning the new instance.

Gweaton: Right, that makes sense 🙏

PR #2251, opened 2024-08-05
Provision standalone Redis instance in production for link-checker-api

brucebolt:
This creates the standalone instance of Redis in production for Link Checker API.

Later commits will switch the application away from the shared instance.

[Trello card](https://trello.com/c/eqvVgy68)

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1450,6 +1450,11 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &link-checker-redis redis://shared-redis-govuk.eks.production.govuk-internal.digital
+          workers: *link-checker-redis
       extraEnv:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -1450,6 +1450,11 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
```
Gweaton, line 1455: Re: [my confusion on the other branch](https://github.com/alphagov/govuk-helm-charts/pull/2248#discussion_r1704190581), if you've got the energy to do so, it could be worth specifying in the commit message that although we're provisioning the new instance with `enabled: true`, the addition of the manual URL override is what ensures we're not switching to it yet.

Gweaton (approved):

charts/app-config/values-production.yaml
brucebolt: Good idea, I've updated the commit message here.

PR #2254, opened 2024-08-05
Change Integration cronjob to import snapshot from Prod instead of Staging

ianhowell-gds:
### What
This is to change the snapshot restore cronjob to read from the Production S3 bucket instead of the Staging one

### Why
There have been a number of times when the snapshot imported from Staging into Integration has been corrupted somehow, resulting in the Integration Chat Opensearch Cluster Health going Red and the Integration Chat Service experiencing an outage. When the Test and Staging environments import snapshots from Production, this issue has not occurred.


rtrinque (approved):
Looks good

PR #2265, opened 2024-08-12
Rough draft of shared asset uploads

KludgeKML:
Rough change I think will be necessary to support 

- do we need to supply a .Values.uploadSharedAssets.path? Or do we just assume that's okay.


KludgeKML:
Closing because hopefully the actual solution will be even easier (we don't need to add an extra call to s3 sync, because we're just going to be putting gem and app assets in one big shared folder)

PR #2266, opened 2024-08-12
Remove GOVUK_NOTIFY_TEMPLATE_ID for frontend in integration

theseanything:
This doesn't appear to be used by frontend.


theseanything:
I couldn't find any references in the https://github.com/alphagov/frontend codebase that uses this env var.

beccapearce (approved):
Looks good to me! :)

PR #2281, opened 2024-08-21
Add new redis instance for contacts admin

callumknights:
Trello: https://trello.com/c/T2w1jKaF/1297-create-new-redis-instance-for-contacts-admin-and-move-contacts-admin-to-it-in-integration-and-staging


JonathanHallam (approved):
👍

PR #2282, opened 2024-08-21
Add email alert frontend to draft stack

KludgeKML:
Defines draft- apps for email-alert-frontend in all environments (draft-router-api was previously pointing to these non-existent apps, so no need to update it).

- Also removes an environment variable from the non-draft apps that is no longer used.

https://trello.com/c/KjFnLQwu/283-add-email-alert-frontend-to-draft-stack


ChrisBAshton (approved):
LGTM 👍

PR #2298, opened 2024-08-27
Add BigQuery env vars for GOV.UK Chat staging and production

chaecramb:
Adds env vars for connecting to the GCP BigQuery projects to be used for export by GOV.UK Chat staging and production environments. 

Note that staging uses the same GCP project as integration, hence the "gov-uk-chat-integration" project value in `charts/app-config/values-staging.yaml`. 

Credentials have already been added to Secrets Manager.


kevindew (approved):
Looks good 👍 

Only minor thing I'd suggest before merging is that you put the description about using integration in the commit message as well as the PR description.

PR #2300, opened 2024-08-28
Add gov.uk/chat as a routing rule in www-origin LB on integration

rtrinque:
We want to be able to access the govuk-chat app through the www.gov.uk/chat URL.
This modify the www-origin LB config to add a rule with higher priority that match on the www.gov.uk host and the /chat path.


ianhowell-gds (approved):
Looks good

PR #2301, opened 2024-08-28
Adds GOV.UK Chat crontask to export to BigQuery

chaecramb:
Automatically runs the `rake bigquery:export` task daily at 7 AM in the production server only.


kevindew (approved):
Looks good 👍

PR #2308, opened 2024-08-28
Fix Redis URL

brucebolt:
We are currently adding a newline character at the end of the string, which isn't needed.

This was based on the information at https://yaml-multiline.info/.

[Trello card](https://trello.com/c/USP5EwYf)


JonathanHallam (approved):
🤞

PR #2322, opened 2024-08-29
Provision standalone Redis for Content Publisher in staging

brucebolt:
[Trello card](https://trello.com/c/JDLHXKuN)

### DIFF
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -630,6 +630,12 @@ govukApplications:
       dbMigrationEnabled: true
       nginxClientMaxBodySize: &max-upload-size 500M
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &content-publisher-redis >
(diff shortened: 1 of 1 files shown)


charts/app-config/values-staging.yaml
```diff
@@ -630,6 +630,12 @@ govukApplications:
       dbMigrationEnabled: true
       nginxClientMaxBodySize: &max-upload-size 500M
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &content-publisher-redis >
```
Gweaton, line 636: I'm assuming we don't need the `-` here after all?

Gweaton (approved):

charts/app-config/values-staging.yaml
brucebolt: It seems to make no difference, the correct value gets populated.

My issue yesterday turned out to be completely unrelated to my changes.

PR #2343, opened 2024-08-30
Add GOV.UK Chat chrontask to promote from waiting list.

peteglondon:
Runs the `users:promote_waiting_list` task every hour.

Trello: https://trello.com/c/TELFc4LR/1853-write-a-rake-task-to-promote-waitinglistusers-to-earlyaccessusers-if-we-have-spaces


kevindew (approved):
Looks good

kevindew:
@peteglondon just clocked that this is only for production, we should also have this task in integration and staging.

PR #2356, opened 2024-09-02
Add site verification files to nginx configuration

KludgeKML:
Google and Bing verification files are currently served by static, and have to be on the root so can't be handled as normal assets. Replace this with direct serve from nginx, given they're small files.

https://trello.com/c/o55zYQeB/317-tidy-verification-paths-in-static

### DIFF
--- a/charts/app-config/templates/router-nginx-config.tpl
+++ b/charts/app-config/templates/router-nginx-config.tpl
@@ -184,6 +184,7 @@ http {
     # Google Search Console (gov.uk) verification files
     # See https://gov-uk.atlassian.net/wiki/spaces/GOVUK/pages/3585441793/Google+Search+Console for ownership of these verification tokens
 
+{{- if (ne .Values.govukEnvironment "production") }}
     location = /googlea6393a390aadfbaa.html {
       add_header Content-Type text/html;
       return 200 'google-site-verification: googlea6393a390aadfbaa.html';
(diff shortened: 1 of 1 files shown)


theseanything (approved):

charts/app-config/templates/router-nginx-config.tpl
```diff
@@ -181,6 +181,26 @@ http {
...
+      add_header Content-Type text/html;
+      return 200 'google-site-verification: googlec908b3bc32386239.html';
+    }
+
+    # Bing (gov.uk) verification files
+
+    location = /BingSiteAuth.xml {
+      add_header Content-Type application/xml;
+      return 200 '<?xml version="1.0"?><users><user>66F6ADB7C4481C94247D1E08FA04C6A4</user></users>';
+    }
```
theseanything, line 202:
```suggestion
        {{- if (ne .Values.govukEnvironment "production") }}
    location = /googlea6393a390aadfbaa.html {
      add_header Content-Type text/html;
      return 200 'google-site-verification: googlea6393a390aadfbaa.html';
    }

    location = /googlec908b3bc32386239.html {
      add_header Content-Type text/html;
      return 200 'google-site-verification: googlec908b3bc32386239.html';
    }

    # Bing (gov.uk) verification files

    location = /BingSiteAuth.xml {
      add_header Content-Type application/xml;
      return 200 '<?xml version="1.0"?><users><user>66F6ADB7C4481C94247D1E08FA04C6A4</user></users>';
    }
    {{- end }}
```

PR #2384, opened 2024-09-04
Provision standalone Redis for integration Short URL Manager

Gweaton:
[Trello card](https://trello.com/c/gs4w9T2A/1332-create-new-redis-instance-for-short-url-manager-and-move-short-url-manager-to-it-lemon-herb-%F0%9F%8D%8B%F0%9F%8C%BF)

This app doesn't have any workers, so this is less involved that the usual trilogy of PRs.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2870,6 +2870,10 @@ govukApplications:
         - name: organisations-import
           task: "organisations:import"
           schedule: "0 4 * * *"
+      redis:
+        enabled: true
+        redisUrlOverride:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2870,6 +2870,10 @@ govukApplications:
         - name: organisations-import
           task: "organisations:import"
           schedule: "0 4 * * *"
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: redis://shared-redis-govuk.eks.integration.govuk-internal.digital
```
yndajas, line 2876: Are there no workers for this one?

yndajas (approved):
Reading commit messages is useful

PR #2398, opened 2024-09-05
Turn off signon authentication for frontend of chat

kevindew:
This is needed to test out the integration of chat into the www hostname.


rtrinque (approved):
Looks good

PR #2399, opened 2024-09-05
Wrap true in quotes to ensure treated as string

kevindew:
This is being treated as a YAML boolean whereas we want it as a string


rtrinque (approved):
Looks good

PR #2418, opened 2024-09-06
Add datastore config env variable to `search-api-v2`

csutter:
This is already part of the AWS SM secret but not actually surfaced to the app as an environment variable.

See https://github.com/alphagov/search-v2-infrastructure/blob/main/terraform/environment/discovery_engine.tf#L71


nimalank7 (approved):
LGTM! (I'm assuming you've got the name correctly etc..)

PR #2421, opened 2024-09-09
Add notify reply to id for GOV.UK Chat

davidgisbey:
We want users to be able to reply to our emails. This adds a GOVUK_NOTIFY_REPLY_TO_ID env var which we use to set the reply to id in our application.


kevindew (approved):
Looks good - only UUID I've verified is prod one.

PR #2424, opened 2024-09-09
Fix duplicate query on Chat Dashboard chart

ianhowell-gds:
## What

Remove duplicate query on Chat Dashboard

## Why 

Duplicate is not required


rtrinque (approved):
Looks good

PR #2430, opened 2024-09-09
Provision Redis instance for staging Publishing API

yndajas:
[Trello](https://trello.com/c/piwScUqs/1334-create-new-redis-instance-for-publishing-api-and-move-publishing-api-to-it-lemon-herb-%F0%9F%8D%8B%F0%9F%8C%BF)

This is part of the work to have each app using its own Redis instance, which will eventually allow us to upgrade Sidekiq

### DIFF
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -1963,6 +1963,12 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &publishing-api-redis >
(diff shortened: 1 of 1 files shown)


charts/app-config/values-staging.yaml
```diff
@@ -1963,6 +1963,12 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &publishing-api-redis >
+            redis://shared-redis-govuk.eks.integration.govuk-internal.digital
```
Gweaton, line 1970: This should be `staging`

Gweaton (approved):

PR #2449, opened 2024-09-10
Point signon token slack alerts to #govuk-publishing-platform

KludgeKML:
These alerts are currently going to 2nd Line, but aren't necessarily actionable there. Reassign to owning team so they can decide whether the alerts are still necessary / useful.

https://trello.com/c/iVHdhusp/248-figure-out-next-steps-for-signon-permissions-token-alerts


JonathanHallam (approved):
Begrudgingly yes these belong to us now 🙈

PR #2463, opened 2024-09-10
Permit ingress to `robots.txt` for Asset Manager in production

brucebolt:
This will allow the `robots.txt` file added in https://github.com/alphagov/govuk-helm-charts/pull/2462 to be visible.


brucebolt:
> Good catch - can you update integration and staging too?

Sure, have added it to all three environments now.

PR #2480, opened 2024-09-12
Switch Support app to use new standalone Redis in integration

brucebolt:
Switch Support app to use new standalone Redis in integration

[Trello card](https://trello.com/c/8XxHPBgW)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2999,9 +2999,7 @@ govukApplications:
       redis:
         enabled: true
         redisUrlOverride:
-          app: &support-redis >
-            redis://shared-redis-govuk.eks.integration.govuk-internal.digital
-          workers: *support-redis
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2999,9 +2999,7 @@ govukApplications:
       redis:
         enabled: true
         redisUrlOverride:
-          app: &support-redis >
-            redis://shared-redis-govuk.eks.integration.govuk-internal.digital
-          workers: *support-redis
+          workers: redis://shared-redis-govuk.eks.integration.govuk-internal.digital*support-redis
```
yndajas, line 3002: `*support-redis` typo

brucebolt: Thanks!  Was caring more about trying to automate this, rather than looking at what I was copy and pasting 😬.

yndajas (approved):

PR #2482, opened 2024-09-12
Provision Redis instance for staging Email Alert API

yndajas:
[Trello](https://trello.com/c/hTtgKKyE/1362-create-new-redis-instance-for-email-alert-api-and-move-email-alert-api-to-it-lemon-and-herb%F0%9F%8D%8B%F0%9F%8C%BF)

This is part of the work to have each app using its own Redis instance, which will eventually allow us to upgrade Sidekiq

### DIFF
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -844,6 +844,12 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &email-alert-api-redis >
+            redis://shared-redis-govuk.eks.staging.govuk-internal.digital
+          workers: *email-alert-api-redis
       ingress:
         enabled: true
         annotations:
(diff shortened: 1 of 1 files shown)


brucebolt (approved):

charts/app-config/values-staging.yaml
```diff
@@ -844,6 +844,12 @@ govukApplications:
       uploadAssets:
         enabled: false
       workerEnabled: true
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &email-alert-api-redis >
+            redis://shared-redis-govuk.eks.integration.govuk-internal.digital
```
brucebolt, line 851:
```suggestion
            redis://shared-redis-govuk.eks.staging.govuk-internal.digital
```

brucebolt (changes requested):

charts/app-config/values-staging.yaml
yndajas: Oops - second time I've done that!

brucebolt (approved):

PR #2507, opened 2024-09-12
Pin clamav image at last working version

theseanything:
Newer images of clamav are unable to start correctly. This pins the configuration to that last working version. This is to allow us time debug what has broken in new images.


MuriloDalRi (approved):

steven-radford (approved):
Cannot claim to be a helm expert by any means but this seems legitimate, and is the matching tag as to what was used to revert production back to a working state.

PR #2514, opened 2024-09-13
Increase resource limits for Production Chat AI

ianhowell-gds:
## What

Increase the CPU & Memory limits & resources for Chat AI in Production

## Why

It has been  suggested we set resources to what we think our normal usage is and only use limits for extra headroom. This sets the number of CPUs in requests to match the number of processes we’ll have.


rtrinque (approved):
Looks good

PR #2515, opened 2024-09-13
Fix the url of the runbook in Chat alerts

ianhowell-gds:
## What

Update the runbook URL in the Chat AI Alerts
Update alerts `High5xxRate` and `LongRequestDuration` to go to pager

## Why

The current url is incorrect, so this should direct people to the right location for the runbook
The `High5xxRate` and `LongRequestDuration` alerts are high priority


rtrinque (approved):
Looks good

PR #2520, opened 2024-09-13
Add path condition on www-origin LB to route chat app in staging

rtrinque:
Add path condition on www-origin LB to route chat app in staging


ianhowell-gds (approved):
LGTM

PR #2521, opened 2024-09-13
Add path condition on www-origin LB to route chat app in production

rtrinque:
Add path condition on www-origin LB to route chat app in production


ianhowell-gds (approved):
LGTM

PR #2563, opened 2024-09-16
Provision new redis instance for Content data API in production

JonathanHallam:
[Trello](https://trello.com/c/n8NZZkcH/1365-create-new-redis-instance-for-content-data-api-and-move-content-data-api-to-it-peri-peri-%F0%9F%8C%B6%EF%B8%8F)

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -523,6 +523,12 @@ govukApplications:
 
   - name: content-data-api
     helmValues:
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &content-data-api-redis >
+            redis://shared-redis-govuk.eks.production.govuk-internal.digital
+          workers: *content-data-api-redis
       ingress:
         enabled: true
         annotations:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -523,6 +523,12 @@ govukApplications:
 
   - name: content-data-api
     helmValues:
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &content-data-api-redis >
+            redis://shared-redis-govuk.eks.staging.govuk-internal.digital
```
brucebolt, line 530:
```suggestion
            redis://shared-redis-govuk.eks.production.govuk-internal.digital
```

brucebolt (changes requested):

charts/app-config/values-production.yaml
JonathanHallam: curse you copy pate, have fixed it

brucebolt (approved):

PR #2567, opened 2024-09-17
Exclude 503 from alert and switch LongRequestDuration from Pagerduty to Slack

ianhowell-gds:
## What

Exclude 503 from the `High5xxRate` alert and switch `LongRequestDuration` alert to go to Slack instead of Pagerduty

## Why

The 503 error code will be used when the Chat service is disabled, so we do not want alerts to go to Pagerduty in that situation.

We have decided that alerts for `LongRequestDuration` should not go to Pagerduty, as there will likely be nothing an on-call engineer will be able to do to resolve the issue.


rtrinque (approved):
Looks good

PR #2569, opened 2024-09-17
Stagger GA processing of each CRON job

georges1996:
In integration we have noticed we run out of memory due to the large amount of data being passed into the queue. A way to reduce the load is to stagger the more memory intensive jobs, to not hit the max memory limit.

We have decided to change the CRON times to start at 10pm for all envs to have consistency.

### DIFF
--- a/charts/search-index-env-sync/values.yaml
+++ b/charts/search-index-env-sync/values.yaml
@@ -36,7 +36,7 @@ cronjobs:
 
   staging:
     blue-es6-cp-prod:  # This would be too long as blue-elasticsearch6-domain-cp-prod
-      schedule: "37 1 * * *"
+      schedule: "00 3 * * *"
       args: [restore_latest]
       extraEnv:
         - name: SNAPSHOT_REPO


kevindew (commented):
Do we need to do anything with the other update_popularity job?

Do we need to wait until 4:30 for the updated_popularity job? as if that was earlier our window would be less, but not sure if that is timed after something else?

Also shouldn't we also be doing same staggering to staging and prod? I imagine with the new Redis they'll have same memory in each environment?

theseanything (approved):

kevindew:
Production and Integration look good to me (i.e. to my knowledge there won't be any problems 🤞)

Staging looks like it will clash with the db sync: https://github.com/alphagov/govuk-helm-charts/blob/ef333196bb062584a691e9694c4fe132a9e6d8ea/charts/search-index-env-sync/values.yaml#L38-L45

Might be good to start it at 2am.

It'd also be good if the commit messages could include the rationale for the gaps between the jobs and rough estimates as to how long each job takes, as this is info that I imagine is only known to you at the moment.

PR #2570, opened 2024-09-17
Increase redis memory usage for search-api

theseanything:
The "search:update_popularity" task currently queues jobs with the full search information, which cause the task queue to use upto 6Gb of memory. We have worked planned to fix this memory issue.


kevindew (approved):
Thanks for this Sean.

Also had to Google what 500m meant for a cpu - glad I didn't accuse you of making a typo 😁

PR #2571, opened 2024-09-17
Increase redis memory for staging/production search-api

theseanything:
Missed staging and production from the previous PR - https://github.com/alphagov/govuk-helm-charts/pull/2570

The "search:update_popularity" task currently queues jobs with the full search information, which cause the task queue to use upto 6Gb of memory. We have worked planned to fix this memory issue.


kevindew (approved):
Great - I didn't realise we would ready to specify this yet. Thanks for sorting.

PR #2584, opened 2024-09-20
Use Europe/London Timezone on chat cron tasks

kevindew:
This allows setting the a time zone for cron jobs so they can run in a London timezone. 

I've also tweaked the config of chat cron job tasks following chats with team leads to avoid situations where we accumulate stuff overnight or have jobs clashing.

### DIFF
--- a/charts/generic-govuk-app/templates/cron-task.yaml
+++ b/charts/generic-govuk-app/templates/cron-task.yaml
@@ -12,6 +12,7 @@ metadata:
     app.kubernetes.io/component: "{{ .name }}"
 spec:
   schedule: "{{ .schedule }}"
+  {{ if .timeZone }}timeZone: "{{ .timeZone }}"{{- end }}
   suspend: {{ .suspend | default false }}
   jobTemplate:
     metadata:
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1343,12 +1343,15 @@ govukApplications:
         - name: promote-waiting-list
           task: "users:promote_waiting_list"
           schedule: "0 7-19 * * *"
+          timeZone: "Europe/London"
         - name: increment-instant-access-places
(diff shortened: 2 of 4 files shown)


charts/generic-govuk-app/templates/cron-task.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
     app.kubernetes.io/component: "{{ .name }}"
 spec:
   schedule: "{{ .schedule }}"
+  {{ if .timeZone }}timeZone: "{{ .timeZone }}"{{- end }}
```
davidgisbey, line 15: i'm not particularly familiar with the syntax here but do we need the string wrapping the .timeZone variable. i.e. could 

```
{{ if .timeZone }}timeZone: "{{ .timeZone }}"{{- end }}
```

be

```
{{ if .timeZone }}timeZone: {{ .timeZone }}{{- end }}
```

davidgisbey (approved):
looks good, i've just left the one optional/nitty and potentially wrong comment

charts/generic-govuk-app/templates/cron-task.yaml
kevindew: So if you do that you lose the previous wrapping.

E.g. with the former:

```
➜  app-config git:(cron-timezone) ✗ cat /tmp/test-2.helm | grep timeZone
  timeZone: "Europe/London"
  timeZone: "Europe/London"
  timeZone: "Europe/London"
```

with the latter:

```
➜  app-config git:(cron-timezone) ✗ cat /tmp/test-2.helm | grep timeZone
  timeZone: Europe/London
  timeZone: Europe/London
  timeZone: Europe/London
```

so it feels safer to do so.

davidgisbey: oh right thanks. I hadn't clocked it stripped the wrapping `"`s. Thanks for the explanation

PR #2601, opened 2024-09-25
Revert "Show chat entry banner on integration"

davidgisbey:
Reverts alphagov/govuk-helm-charts#2600. We're done with testing.


JamesCGDS (approved):
LGTM 👍

PR #2620, opened 2024-09-27
Update init containers to be PSS compliant in the apps namespace

nimalank7:
Description:
- Enforce initContainers in the `app` namespace to be compliant when PSS is set to [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- Tested in integration and observed that `content-data-admin` initContainers were starting properly
- `govuk-mirror-sync-cronjob` can't be tested in integration as this has no mirrors so the PR will have to be merged and we will have to manually inspect staging ot see if it succeeds.
- Paired with @MahmudH

### DIFF
--- a/charts/generic-govuk-app/templates/deployment.yaml
+++ b/charts/generic-govuk-app/templates/deployment.yaml
@@ -63,6 +63,15 @@ spec:
           volumeMounts:
             - name: assets
               mountPath: /assets
+          securityContext:
+            allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation | default "false" }}
+            runAsNonRoot: {{ .Values.securityContext.runAsNonRoot | default "true" }}
+            readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem | default "true" }}
+            seccompProfile:
+              type: RuntimeDefault
+            capabilities:
+              drop: ["ALL"]
+
       {{- end }}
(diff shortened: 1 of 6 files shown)


charts/generic-govuk-app/templates/deployment.yaml
```diff
@@ -127,9 +136,9 @@ spec:
           {{- end }}
           securityContext:
             allowPrivilegeEscalation: {{ .Values.securityContext.allowPrivilegeEscalation }}
-            readOnlyRootFilesystem: true
+            readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFileSystem }}
             capabilities:
-              drop: ["ALL"]
+              drop: {{ .Values.securityContext.capabilities.drop }}
```
theseanything, line 141: Why do we need to read these in as Values?

nimalank7: No particular reason. Mostly just being consistent with extracting everything out to values.

theseanything: If we're not changing it (and we wanna actually enforce this value anyway) - there's no need to have it as a variable. Just adds complexity to the configuration.

nimalank7: Yep that's fair. I've pushed up now!

theseanything (approved):

PR #2633, opened 2024-09-30
Use standalone Redis for emergency banner and Whitehall taxonomy cache in staging

brucebolt:
Whitehall uses Redis for three purposes:
- Taxonomy Cache
- Emergency Banner
- Sidekiq

We've created a new dedicated Redis instance for this to move it off the shared instance, and using different numerically-named databases which are accessed via the trailing `/1` or `/2` on that instance, rather than namespacing.

[Trello card](https://trello.com/c/6R3NueMz)


Gweaton (approved):
Looks good - as long as you're happy to go for the big bang approach of updating both at once!

PR #2669, opened 2024-10-03
Remove Chat Tech dashboard due to corruption

ianhowell-gds:
## What

Remove the GOV.UK Chat Technical Dashboard from Grafana

## Why

It looks like it has become corrupted, so the intention is to remove it and re-add it to resolve the issue


chaecramb (approved):

kevindew (approved):
Looks good. 

Do we have a good handle on when we'll know that this is synced to all envs so we know when we can revert?

ianhowell-gds:
I'm monitoring ArgoCD, so once deployed, I'll do the revert PR

PR #2671, opened 2024-10-03
Switch smoke tests to GOV.UK End to End tests

theseanything:
This removes Smokey tests in favor of the GOV.UK End to End test written with playwright. These tests have feature parity, run faster and more reliably.


MuriloDalRi (approved):

AgaDufrat:
@theseanything maybe worth posting in #govuk-developers about this

PR #2672, opened 2024-10-03
Switch search-api-v2 to new redis instance in integration

callumknights:
Provision new redis instance for search-api-v2 in integration and switch application and workers to use it.

Draining approach not necessary at Sidekiq/redis are not used for the workers.

Trello: https://trello.com/c/H4PW2hNC/1399-create-new-redis-instance-for-search-api-v2-and-move-search-api-v2-to-it-peri-peri-%F0%9F%8C%B6%EF%B8%8F

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2517,6 +2517,8 @@ govukApplications:
         types:
           - command: ['bin/rake', 'document_sync_worker:run']
             name: document-sync-worker
+      redis:
+        enabled: true
       cronTasks:
         - name: quality-monitoring-assert-invariants
           task: "quality_monitoring:assert_invariants"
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2517,6 +2517,8 @@ govukApplications:
         types:
           - command: ['bin/rake', 'document_sync_worker:run']
             name: document-sync-worker
+        redis:
+          enabled: true
```
brucebolt, line 2521: The indentation here needs to be one tab to the left, so it's not nested under `workers`.

callumknights: Sorted, thanks for spotting that 👍

brucebolt (approved):

PR #2678, opened 2024-10-04
Provision new redis instance for Draft email alert fe in integration

JonathanHallam:
[Trello](https://trello.com/c/lIGs2vnW/1417-create-new-redis-instance-for-draft-email-alert-frontend-and-move-draft-email-alert-frontend-to-it-lemon-and-herb-%F0%9F%8D%8B%F0%9F%8C%BF)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -975,6 +975,12 @@ govukApplications:
       arch: arm64
       rails:
         createKeyBaseSecret: false
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &email-alert-draft-frontend-redis >
+            redis://shared-redis-govuk.eks.integration.govuk-internal.digital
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -975,15 +975,19 @@ govukApplications:
       arch: arm64
       rails:
         createKeyBaseSecret: false
+      redis:
+        enabled: true
+        redisUrlOverride:
+          app: &email-alert-draft-frontend-redis >
+            redis://shared-redis-govuk.eks.integration.govuk-internal.digital
+          workers: *email-alert-draft-frontend-redis
```
brucebolt, line 983: This app doesn't have any workers (a value `workers: enabled: true`) in the helm chart would indicate the app having workers.

So each environment can be switched in one PR, with the change to just this and no draining required:
```yaml
      redis:
        enabled: true
```

PR #2687, opened 2024-10-07
Update smokey to be PSS compliant

MahmudH:
Description:
- Enforces this container to be compliant when PSS is set to [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- As part of #1883


nimalank7 (approved):
LGTM!

PR #2694, opened 2024-10-08
Reduce memory limit for Redis in search-api in int

davidgisbey:
This was bumped in this PR https://github.com/alphagov/govuk-helm-charts/pull/2570 to handle the overnight cron jobs running on Search API.

Now we've addressed the issue by removing giant payloads being passed in as params to workers we can revert the changes and test that the fix works as expected.


ianhowell-gds (approved):
LGTM

PR #2696, opened 2024-10-08
Refactor security context

nimalank7:
Description:
- Pod containers need to have the following fields to ensure compliance when PSS is set to [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- Refactored some fields into the podSecurityContext


MahmudH (approved):
LGTM

PR #2697, opened 2024-10-08
Refactor govuk-jobs to be compliant when PSS is set to restricted

nimalank7:
Description:
- Containers need additional fields to be compliant when PSS is set to [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)
- Part of https://github.com/alphagov/govuk-helm-charts/issues/1883


MahmudH (approved):
LGTM

MahmudH (approved):
LGTM

PR #2698, opened 2024-10-08
Refactor db-backup cronjobs securityContext

nimalank7:
Description:
- Some of the fields can be moved into `podSecurityContext` so that they apply to all containers instead of setting it multiple times in the `container` field `securityContext`
- Part of https://github.com/alphagov/govuk-helm-charts/issues/1883


MahmudH (approved):
LGTM

PR #2699, opened 2024-10-08
Enforce govuk-exporter to be compliant when PSS is set to restricted

nimalank7:
Description:
- Enforce `govuk-exporter` to be compliant when PSS is set to [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)


MahmudH (approved):
LGTM

PR #2701, opened 2024-10-08
Enforce renovate to be compliant when PSS is restricted

nimalank7:
Description:
- Run as UID > 1000 to avoid conflicts with host's user table. See [here](https://kubesec.io/basics/containers-securitycontext-runasuser/)
- Enforce `renote` to be compliant when PSS is said to [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)


MahmudH (approved):
LGTM

PR #2702, opened 2024-10-08
Enforce search-index-env-sync

nimalank7:
Description:
- Enforce `search-index-env-sync` to be compliant when PSS is saod to be [restricted](https://kubernetes.io/docs/concepts/security/pod-security-standards/)


MahmudH (approved):
LGTM

PR #2705, opened 2024-10-15
Add new internal ALB for search-api-v2

sihugh:
This is a new load balancer, because it behaves differently from the backend apps (this one is internal to the VPC), and will require different WAF settings because it will have a different traffic profile.

It allows traffic to be routed to search-api-v2 from applications outside the EKS cluster such as an API gateway.

We're doing this to allow publicly routed search API requests, while maintaining more control over what happens than usual. We don't want to adopt the previous approach of allowing open access to the API via a www.gov.uk route (eg www.gov.uk/api/search-v2.json) because of the cost implications. Because of that, this load balancer allows internal traffic only.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2504,6 +2504,18 @@ govukApplications:
     helmValues:
       arch: arm64
       dbMigrationEnabled: false
+      ingress:
+        enabled: true
+        annotations:
+          <<: [*alb-ingress-defaults]
+          alb.ingress.kubernetes.io/scheme: "internal"
+          alb.ingress.kubernetes.io/group.order: "172"
+          alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
+            [{"field": "host-header", "hostHeaderConfig": { "values": [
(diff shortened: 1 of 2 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2504,6 +2504,18 @@ govukApplications:
...
       dbMigrationEnabled: false
+      ingress:
+        enabled: true
+        annotations:
+          <<: [*alb-ingress-defaults]
+          alb.ingress.kubernetes.io/scheme: "internal"
+          alb.ingress.kubernetes.io/group.order: "172"
+          alb.ingress.kubernetes.io/conditions.{{ .Release.Name }}: >
+            [{"field": "host-header", "hostHeaderConfig": { "values": [
+                "search-api-v2.{{ .Values.publishingDomainSuffix }}"
```
theseanything, line 2515: Do we need to label it v2?

```suggestion
                "search-api.{{ .Values.publishingDomainSuffix }}"
```

sihugh: Possibly not - just trying to differentiate from search-api-not-v2, and reckon it might get confusing

theseanything (approved):

PR #2711, opened 2024-10-17
Deploy `govspeak-preview` to all environments

brucebolt:
This application is currently hosted on Heroku. We are moving it to Kubernetes.

[Trello card](https://trello.com/c/dhgDqvBa)

### DIFF
--- a/charts/app-config/image-tags/integration/govspeak-preview
+++ b/charts/app-config/image-tags/integration/govspeak-preview
@@ -1,0 +1,3 @@
+image_tag: latest
+automatic_deploys_enabled: true
+promote_deployment: false
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -1309,6 +1309,8 @@ govukApplications:
               name: signon-token-government-frontend-publishing-api
               key: bearer_token
 
+  - name: govspeak-preview
+
   - name: govuk-chat
     helmValues:
       dbMigrationEnabled: true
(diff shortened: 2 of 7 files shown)


charts/app-config/image-tags/integration/govspeak-preview
```diff
@@ -0,0 +1,3 @@
+image_tag: latest
+automatic_deploys_enabled: true
+promote_deployment: false
```
yndajas, line 3: **question:** promoting deployments

We don't want to promote deployment?

charts/app-config/values-staging.yaml
```diff
@@ -1308,6 +1308,7 @@ govukApplications:
             secretKeyRef:
               name: signon-token-government-frontend-publishing-api
               key: bearer_token
+  - name: govspeak-preview
```
yndajas, line 1311: **polish:** spacing

A line break before this entry would be good

yndajas (approved):

charts/app-config/image-tags/integration/govspeak-preview
brucebolt: This file gets updated by the `govuk-ci` user when we need a deployment to be promoted (along with the correct `image_tag`).  At the moment, it doesn't matter as I'll be manually deploying the first build anyway.

charts/app-config/values-staging.yaml
brucebolt: Thanks, will add this.

PR #2712, opened 2024-10-17
Allow ingress to `govspeak-preview`

brucebolt:
This sets up the ALB required to allow ingress to this application from the public.

[Trello card](https://trello.com/c/dhgDqvBa)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1309,6 +1309,18 @@ govukApplications:
               key: bearer_token
 
   - name: govspeak-preview
+    helmValues:
+      ingress:
+        enabled: true
+        annotations:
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1309,6 +1309,18 @@ govukApplications:
               key: bearer_token
 
   - name: govspeak-preview
+    helmValues:
```
yndajas, line 1312: **suggestion:** abbreviation

Might be worth spelling out what ALB stands for in the commit message

yndajas (approved):

charts/app-config/values-integration.yaml
brucebolt: Done.

PR #2716, opened 2024-10-21
Adding daily cronjob to run the rake task in publisher

baisa:
[Trello card](https://trello.com/c/FDV98XMH/489-ensure-scheduled-editions-are-published-mainstream)
[Rake task in mainstream publisher](https://github.com/alphagov/publisher/pull/2385)
This rake task should publish any missed scheduled editions.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1901,6 +1901,9 @@ govukApplications:
         - name: reschedule-pubs-after-db-restore  # non-prod only
           task: "editions:requeue_scheduled_for_publishing"
           schedule: "38 7 * * 1-5"
+        - name: publish-missed-scheduled-editions
+          task: "editions:publish_missed_scheduled_editions"
+          schedule: "27 5 * * *"
       ingress:
         enabled: true
         annotations:
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1937,6 +1937,9 @@ govukApplications:
         - name: mail-fetcher
           command: "script/mail_fetcher"
           schedule: "*/5 * * * *"
+        - name: publish-missed-scheduled-editions
+          task: "editions:publish_missed_scheduled_editions"
+          schedule: "11 1 * * *"
       ingress:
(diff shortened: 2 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1901,6 +1901,9 @@ govukApplications:
         - name: reschedule-pubs-after-db-restore  # non-prod only
           task: "editions:requeue_scheduled_for_publishing"
           schedule: "38 7 * * 1-5"
+        - name: publish-missed-scheduled-editions
+          task: "editions:publish_missed_scheduled_editions"
+          schedule: "15 5 * * * *"
```
mtaylorgds, line 1906: I think there's an extra asterisk in each of these schedules. I believe there should be 5 components, where these each have 6.

charts/app-config/values-production.yaml
```diff
@@ -1937,6 +1937,9 @@ govukApplications:
         - name: mail-fetcher
           command: "script/mail_fetcher"
           schedule: "*/5 * * * *"
+        - name: publish-missed-scheduled-editions
+          task: "editions:publish_missed_scheduled_editions"
+          schedule: "15 2 * * * *"
```
mtaylorgds, line 1942: What's the reason that production is scheduled to run at 0215, and integration & staging run at 5.15?

mtaylorgds (commented):
I think there's a syntax error with the schedules.

theseanything:
Can you add a bit of randomness into the schedules? - i.e. not on exactly quarter past
Helps distribute the resource usage of jobs.

charts/app-config/values-production.yaml
baisa: We want to run the job on production not too long after midnight, as this when publications are often scheduled and we do not want to have much of delay. For integration and staging we want to run them after the nightly environment sync jobs, so they reflect prod accordingly.

mtaylorgds (approved):
LGTM—might be worth confirming @theseanything is also happy before merging.

theseanything (approved):

PR #2718, opened 2024-10-22
Add smart survey env vars for chat

davidgisbey:
## Description

We're integrating with the SmartSurvey API to retrieve information from our survey for Chat. This commit adds the necessary environment variables for auth and retrieval of the survey data.

## Trello card

https://trello.com/c/IuRaZivo/2005-use-the-smartsurvey-api-to-export-to-bigquery-the-number-of-responses-weve-got

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1415,6 +1415,16 @@ govukApplications:
             secretKeyRef:
               name: signon-app-chat
               key: oauth_secret
+        - name: SMART_SURVEY_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: api_key
+        - name: SMART_SURVEY_API_KEY_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: api_key_secret
         - name: REDIS_URL
           value: redis://chat-redis.eks.integration.govuk-internal.digital
         - name: BIGQUERY_PROJECT
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1432,6 +1432,16 @@ govukApplications:
             secretKeyRef:
               name: signon-app-chat
               key: oauth_secret
+        - name: SMART_SURVEY_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: api_key
+        - name: SMART_SURVEY_API_KEY_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: api_key_secret
         - name: WEB_CONCURRENCY
           value: '4'
         - name: REDIS_URL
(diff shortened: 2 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -1415,6 +1415,21 @@ govukApplications:
             secretKeyRef:
               name: signon-app-chat
               key: oauth_secret
+        - name: SMART_SURVEY_URL
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: url
```
kevindew, line 1422: Does this actually need to be a secret? I'd wonder if we'd just have this set-up in our code base and not need an env var for it? is it: https://api.smartsurvey.io/ ?

charts/app-config/values-integration.yaml
```diff
@@ -1415,6 +1415,21 @@ govukApplications:
...
+        - name: SMART_SURVEY_BASIC_AUTH_USERNAME
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: username
+        - name: SMART_SURVEY_BASIC_AUTH_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: govuk-chat-smart-survey
+              key: password
```
kevindew, line 1432: The terminology of these seem a little different to what Smart Survey uses of API Key and API Key secret - even if they are ended up used as basic auth - so might be better to use their terminology.

kevindew (commented):
Just a few suggestions

charts/app-config/values-integration.yaml
davidgisbey: it has the id of our survey in it

charts/app-config/values-integration.yaml
davidgisbey: They sort of half use the terminology. I ended up using the username and pw due to potential confusion around stuff like this

image

charts/app-config/values-integration.yaml
kevindew: Right - hmm that seems to be a bit unusual as we'd be storing exactly one API endpoint specifically. I don't think we should do that.

I think we should have the survey id separate though so we can use it to form a url. I don't think it actually needs to be secret though it would be in [the URL of our survey](https://surveys.publishing.service.gov.uk/s/govuk-chat-beta/?user=b1f782ef-b3f6-4164-9689-9ecca83333c6) if we hadn't set a custom slug and it is available in the HTML source code of the page (1561650) - so I'd expect we just want it in our rails config unless we want to test with other surveys and thus have it as an env var.

charts/app-config/values-integration.yaml
kevindew: Ah I can understand that, though I assume they had to use that there because of the library they're using for modelling those API calls.

They also manage to go further inconsistent in their attempt to make it clear: https://docs.smartsurvey.io/docs/getting-started#getting-your-api-key:~:text=Basic%20Auth%2C%20Usernames,the%20Token%20Secret. by calling them token and token secret. I think API Key and API Key secret is the easiest to match to the credentials view though.

davidgisbey: okay cool i'll update them in the secrets manager and update them here 👍

davidgisbey:
@kevindew i've updated values in secrets manager and here

kevindew (approved):
Nice one, thanks

PR #2720, opened 2024-10-22
Update chat big query frequency export to run hourly

davidgisbey:
## Description

We want to run this hourly so we've got more up do date stats.

## Trello card

https://trello.com/c/EGoUT5e5/2004-work-with-dan-to-improve-frequency-and-data-exported-so-he-can-create-more-insightful-dashboards

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1368,7 +1368,7 @@ govukApplications:
       cronTasks:
         - name: bigquery-export
           task: "bigquery:export"
-          schedule: "0 7 * * *"
+          schedule: "5 * * * *"
           timeZone: "Europe/London"
         - name: promote-waiting-list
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -1368,7 +1368,7 @@ govukApplications:
       cronTasks:
         - name: bigquery-export
           task: "bigquery:export"
-          schedule: "0 7 * * *"
+          schedule: "0 * * * *"
```
theseanything, line 1371: Can you make this run a random minute (i.e. every 36 minute) past the hour to distribute the resource usage - as other cronjobs are likely to be set on the hr?

davidgisbey: sure 👍

kevindew (approved):
Nice one

PR #2729, opened 2024-10-28
Increase worker pods for gov.uk chat to 24 in Production

rtrinque:
Scaling up Production before we go live


ianhowell-gds (approved):
LGTM

PR #2731, opened 2024-10-29
Use AWS hostnames for chat memcache servers

kevindew:
Trello: https://trello.com/c/sZ9PIjmv/2060-set-up-dedicated-memcache-instances-for-chat

These AWS memcache servers are accessed via TLS, if we use the govuk-internal.digital hostnames then we can't verify the hostname as part of TLS verification. Switching to these AWS ones allows it.

While these all use the default port and it isn't necessary to include it, I thought I would anyway to verbatim copy the endpoint URL that AWS provide in their console.


ianhowell-gds (approved):
LGTM

PR #2732, opened 2024-10-30
Migrate Chat Service from `backend` load balancer to it's own one

ianhowell-gds:
## What

Migrate the Chat Service off of the shared `backend` load balancer and onto it's own load balancer

## Why

This is to reduce the risk on the other services attached to the shared load balancer being affected if the Chat Service gets overloaded when put live

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -48,6 +48,10 @@ _alb-ingress-waf-ruleset-backend: &alb-ingress-waf-ruleset-backend
   alb.ingress.kubernetes.io/wafv2-acl-arn: >-
     arn:aws:wafv2:eu-west-1:210287912431:regional/webacl/backend_public_web_acl/b5f616fd-699e-4b44-8ea2-7afd6626aa98
 
+_alb-ingress-waf-ruleset-chat: &alb-ingress-waf-ruleset-chat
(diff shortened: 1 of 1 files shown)


ianhowell-gds:
Working with Sam Simpson of the Platform Engineering Team & Roch Trinique we do not believe that having [GOV.UK](http://gov.uk/) Chat share a load balancer presents a realistic risk to the stability of [GOV.UK](http://gov.uk/) because:

- The load balancer should easily cope with the traffic levels we are expecting to see, as it will scale accordingly to handle any increase
- The load balancer will cope with traffic levels higher than our application is scaled for
- Fastly will filter out most of the nefarious requests to the service, limiting any DDoS attack

Having the load balancer as a shared component does mean there is an element of associated risk, however we have agreed that this risk is preferable to the complexity of setting up additional load balancers, because this would overly complicate the configuration of Fastly and AWS due to having to split traffic routing and thus make maintenance and decommissioning of [GOV.UK](http://gov.uk/) chat unnecessarily expensive. Similarly, we do not expect to have time to complete this task before chat goes live, and worry introducing it after going live risks an outage.

We also noted that this isn't the only aspect of shared infrastructure as Chat is served from the same Kubernetes cluster, which could similarly be affected - so we don't believe this is an outlier risk.

PR #2734, opened 2024-10-31
Define secret for Publishing API to access BigQuery

pezholio:
This defines a secret to be used in Publishing API to allow the team to access BigQuery for Analytics data (See https://github.com/alphagov/publishing-api/pull/2957)

The secrets are set up in all three AWS environments


richardTowers (approved):
Looks good to me

PR #2735, opened 2024-10-31
Disable automatic production deploys for GOV.UK Chat

kevindew:
This is a cautionary measure that we want to take on Monday when there is going to be a filmed demo of chat for TV and live demo's for journalists.

This shouldn't be merged before Monday morning (4th November)


ianhowell-gds (approved):
LGTM

PR #2736, opened 2024-11-01
Use BigQuery secrets in Publishing API

pezholio:
Now we have the secrets defined (see https://github.com/alphagov/govuk-helm-charts/pull/2734), we can use them and set the relevant environment variables in Publishing API.


richardTowers (approved):
Looks good, names match

PR #2741, opened 2024-11-05
Add env var for GOV.UK Chat Slack integration

jackbot:
Extracts a secret from Secrets Manager and sets it as an env var in the
GOV.UK Chat app.

We only need this in production as we don't want to trigger Slack
notifications from other environments.


ianhowell-gds (approved):
LGTM

kevindew (commented):
Looks good to me. Do wonder if we also want something in integration so we can test code that performs a notification without needing to deploy all the way to prod? Then again may not be too big a deal if the Slack integration is simple.

kevindew (approved):

PR #2743, opened 2024-11-05
Increase webpod to 16 and reduce worker to 4 for chat.

rtrinque:
We want to increase the number of webpod to absorb the onslaught of new users signing up to the service.
Worker are not involved in the signup process and were increased previously by mistake.


ianhowell-gds (approved):
LGTM

PR #2745, opened 2024-11-05
Correct scaling up chat for launch

kevindew:
This resets the value of 24 worker processes back to 4, since scaling this up would have not the desired effect of making the system more resilient for sign-ups as the worker processes are not involved in them.

It does double the number of web pods so that there are more of these available as this will allow handling more web requests and thus increase ability to handle signups.


kevindew:
Dupe of https://github.com/alphagov/govuk-helm-charts/pull/2743/files

PR #2747, opened 2024-11-05
Match production resources values in staging for web pods

rtrinque:
We temporarily bump up staging resources to the level of production for the purpose of load testing.


ianhowell-gds (approved):
LGTM

PR #2764, opened 2024-11-13
Redirect chat.eks domain to publishing.service domain in staging and …

rtrinque:
…integration

We don't want people to be able to access the chat client directly through the load balancer.


ianhowell-gds (approved):
Looks good

PR #2765, opened 2024-11-13
Redirect chat.eks domain to www.gov.uk in production

rtrinque:
We don't want people to be able to access the chat client directly through the load balancer.


ianhowell-gds (approved):
LGTM

PR #2767, opened 2024-11-14
Finder Frontend: Remove autocomplete URL env config

csutter:
This will now be a route directly on Finder Frontend itself rather than an external API, so it no longer needs configuring.


kevindew (approved):
Looks good

PR #2773, opened 2024-11-15
Mount `emptyDir` volume into the `script` template of the `parse-failures` workflow

nimalank7:
Description:
- In order to `runAsNonRoot` we can't have containers writing to the base layer. So mount an `emptyDir` volume to write `message.txt` to - see guidance [here](https://argo-workflows.readthedocs.io/en/latest/empty-dir/)


nimalank7:
Unnecessary. `Emissary` is the default workflow executor and can extract artifacts from `/tmp`. See https://argo-workflows.readthedocs.io/en/latest/workflow-executors/

PR #2778, opened 2024-11-18
Add to integration support for Frontend to render the static error pages

KludgeKML:
We want to test that frontend can render static error pages in the same way that Static does (so that we can remove that from Static on the path to retiring it entirely) - this adds a similar job to the equivalent one for static. Both will run in integration, but static deployments are fairly rare so that shouldn't cause any problems with testing.

- Add in some temporary guard clauses so that this won't cause versions of frontend without the error pages code to fail.

(works with https://github.com/alphagov/frontend/pull/4382, but guard clause should mean it doesn't fail with other branches of Frontend)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1179,6 +1179,7 @@ govukApplications:
   - name: frontend
     helmValues:
       arch: arm64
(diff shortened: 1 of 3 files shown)


nimalank7 (approved):
LGTM!

PR #2785, opened 2024-11-20
Remove REDIS_URL env var from draft-email-alert-frontend

samsimpson1:
This was left in by mistake when the work to replace the shared redis instance was done

https://github.com/alphagov/govuk-infrastructure/issues/1477


yndajas (approved):
Thanks!

JonathanHallam (approved):
good find, thanks

PR #2792, opened 2024-11-25
Set `GIT_CONFIG_GLOBAL` to `/tmp` folder to prevent writing to root filesystem

nimalank7:
Description:
- Currently the `update-image-tag` workflow writes to `/root/.gitconfig` when `git config --global user.email` and `git config --global user.name` is called
- Setting `GIT_GLOBAL_CONFIG: tmp` will set the `.gitconfig` file in `/tmp` which is an `emptyDir` volume that is mounted into our container
- This will allow `readOnlyRootFileSystem: true` to be set on Argo Workflows to further harden our containers


theseanything (approved):
Alternatively, if this doesn't work we can just mount config map as the git config - so it doesn't have to write one.

PR #2804, opened 2024-11-29
Remove chat promo banner env var from Smart answers

davidgisbey:
We're no longer rendering the chat promo banner on any pages in Smart answers, so we can remove the related environment variable.


AgaDufrat (approved):
Probably worth removing from fronted repo as well: https://github.com/alphagov/frontend/blob/2bcb6c112948724fb42b7ec28ed1f1559471cc0d/app/helpers/govuk_chat_promo_helper.rb

kevindew:
Thanks @AgaDufrat - we're still using the banner in the frontend app.

PR #2807, opened 2024-12-02
Give chat 800 places a day

kevindew:
This increases the frequency of place allocation to once an hour and lowers the amount to 100. This task will run 8 times a day allocating 800 places.

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -1385,7 +1385,7 @@ govukApplications:
           timeZone: "Europe/London"
         - name: increment-instant-access-places
           task: "settings:increment_pilot_places[instant_access]"
-          schedule: "0 9,11,13,15 * * *"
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -1385,7 +1385,7 @@ govukApplications:
           timeZone: "Europe/London"
         - name: increment-instant-access-places
           task: "settings:increment_pilot_places[instant_access]"
-          schedule: "0 9,11,13,15 * * *"
+          schedule: "0 9,10,11,12,13,14,15,16 * * *"
```
jackbot, line 1388: We could use `0 9-16 * * *` here to make it a little terser?

kevindew: Good shout

jackbot (approved):

PR #2819, opened 2024-12-04
Revert "Load routes from Content Store in integration/staging"

theseanything:
Reverts alphagov/govuk-helm-charts#2816


theseanything:
Not needed anymore

PR #2823, opened 2024-12-06
[WIP] Add a global variable to enable search autocomplete

leenagupte:
The new search autocomplete is due to be rolled out across GOV.UK

Autocomplete uses a denylist to stop certain terms from serving suggestions. The process to update the denylist is very manual and developer heavy.

If might be difficult to update the denylist over the holiday period. In that situation the decision might be taken to temporarily turn off autocomplete entirely.

A global environment variable has been added rather than individual environment variables for each affect app as the search bar exists in the header on all pages on GOV.UK. This means that it needs to be accessed by the [layout_super_navigation_header](https://github.com/alphagov/govuk_publishing_components/pull/4451/files) which is pulled in by static for use in 6 other rendering apps.

### DIFF
--- a/charts/app-config/templates/env-configmap.yaml
+++ b/charts/app-config/templates/env-configmap.yaml
@@ -16,6 +16,7 @@ data:
   GOVUK_APP_DOMAIN_EXTERNAL: {{ .Values.publishingDomainSuffix }}
   GOVUK_ASSET_ROOT: https://{{ .Values.assetsDomain }}
   GOVUK_CSP_REPORT_URI: {{ .Values.cspReportURI | quote }}
+  GOVUK_ENABLE_SEARCH_AUTOCOMPLETE: {{ .Values.enableSearchAutocomplete }}
   GOVUK_ENVIRONMENT: {{ .Values.govukEnvironment }}
   GOVUK_ENVIRONMENT_NAME: {{ .Values.govukEnvironment }}
   # TODO: remove GOVUK_PROMETHEUS_EXPORTER once govuk_app_config >=9.10 is everywhere.
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -60,6 +60,8 @@ _alb-ingress-group-backend: &alb-ingress-group-backend
 emergency-banner-redis:
   - &emergency-banner-redis redis://whitehall-admin-redis/1
 
+enableSearchAutocomplete: "true"
+
 # Apps for Argo CD to deploy, along with any app-specific Helm values.
 
 govukApplications:
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -51,6 +51,8 @@ _alb-ingress-group-backend: &alb-ingress-group-backend
 emergency-banner-redis:
   - &emergency-banner-redis redis://whitehall-admin-redis/1
 
+enableSearchAutocomplete: "true"
+
 # Apps for Argo CD to deploy, along with any app-specific Helm values.
 
 govukApplications:
--- a/charts/app-config/values-staging.yaml
+++ b/charts/app-config/values-staging.yaml
@@ -58,6 +58,8 @@ _alb-ingress-group-backend: &alb-ingress-group-backend
 emergency-banner-redis:
   - &emergency-banner-redis redis://whitehall-admin-redis/1
 
+enableSearchAutocomplete: "true"
+
 # Apps for Argo CD to deploy, along with any app-specific Helm values.
 
 govukApplications:


theseanything (approved):

theseanything:
This will set the value to "true" for all environments at the same time. If you'd like to have per environment control consider using Helm value and then setting it in the environment helm values file.

theseanything:
> > This will set the value to "true" for all environments at the same time. If you'd like to have per environment control consider using Helm value and then setting it in the environment helm values file.
> 
> Can you still make it global per environment?

As in it will be applied to all apps in that environment - yes!

charts/app-config/templates/env-configmap.yaml
```diff
@@ -16,6 +16,7 @@ data:
   GOVUK_APP_DOMAIN_EXTERNAL: {{ .Values.publishingDomainSuffix }}
   GOVUK_ASSET_ROOT: https://{{ .Values.assetsDomain }}
   GOVUK_CSP_REPORT_URI: {{ .Values.cspReportURI | quote }}
+  GOVUK_ENABLE_SEARCH_AUTOCOMPLETE: "true"
```
theseanything, line 19:
```suggestion
  GOVUK_ENABLE_SEARCH_AUTOCOMPLETE: {{ .Values.enableSearchAutocomplete }}
```

Then add the associated value in:
https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-integration.yaml
https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-staging.yaml
https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-production.yaml

e.g.
```
enableSearchAutocomplete: "true"
```

leenagupte: Oh! That's much nicer! Thanks ❤️

charts/app-config/templates/env-configmap.yaml
```diff
@@ -16,6 +16,7 @@ data:
   GOVUK_APP_DOMAIN_EXTERNAL: {{ .Values.publishingDomainSuffix }}
   GOVUK_ASSET_ROOT: https://{{ .Values.assetsDomain }}
   GOVUK_CSP_REPORT_URI: {{ .Values.cspReportURI | quote }}
+  GOVUK_ENABLE_SEARCH_AUTOCOMPLETE: {{ .Values.enableSearchAutocomplete }}
```
kevindew, line 19:
```suggestion
  {{- if .Values.enableSearchAutocomplete }}
  GOVUK_ENABLE_SEARCH_AUTOCOMPLETE: "true"
  {{- end }} 
```

Otherwise I believe that even if we remove the `enableSearchAutocomplete` declaration an env var will still be created so we won't be able to do a simple check of `if ENV["GOVUK_ENABLE_SEARCH_AUTOCOMPLETE"]` and instead need to check the contents of the env var - which given there's a bunch of apps could be a bit tedious (and relates to the eternal struggle of configuring the boolean env var).


(slight caveat that I'm not 100% sure of exactly helm works in this regard but expect a configmap of "GOVUK_ENABLE_SEARCH_AUTOCOMPLETE:` produces `ENV["GOVUK_ENABLE_SEARCH_AUTOCOMPLETE"] = ""`)

charts/app-config/values-integration.yaml
```diff
@@ -60,6 +60,8 @@ _alb-ingress-group-backend: &alb-ingress-group-backend
 emergency-banner-redis:
   - &emergency-banner-redis redis://whitehall-admin-redis/1
 
+enableSearchAutocomplete: "true"
```
kevindew, line 63:
```suggestion
enableSearchAutocomplete: true
```

As per suggestion above about using this as a boolean rather than the contents of the env var

kevindew (commented):
Nice one picking this up @leenagupte, lets talk about the other bits as I was volunteering my services on this so happy to share load / pains of setting this stuff up however you'd like.

For this I do wonder if this is better flipped to be an env var that disables rather than enables (e.g. GOVUK_DISABLE_SEARCH_AUTOCOMPLETE). My logic is that given the normal state will be for this to be enabled then having a env var to disable it will mean that by default we need to find ways to set the env var for devs to experience autocomplete locally and thus will need extra set-up.

charts/app-config/templates/env-configmap.yaml
leenagupte: I'm not sure about booleans either. I wasn't sure if boolean would be preserved.

Regarding the name of the ENV VAR, I made it "truthy" because that hurts my brain less when it came to using the env var in the rendering apps.

This change was just trying to figure out how it "could" be done, so I'm happy for you to take over and flip this.

leenagupte:
Closing in favour of: https://github.com/alphagov/govuk-helm-charts/pull/2832

PR #2836, opened 2024-12-10
Upload 503.html page to mirror bucket as part of static error page creation

KludgeKML:


charts/generic-govuk-app/upload-frontend-error-pages.sh
```diff
@@ -9,4 +9,6 @@ mkdir -p "${OUTPUT_PATH}"
 cd "${OUTPUT_PATH}"
 curl --fail-early -fo '#1.html' "${SERVICE}/static-error-pages/{${ERROR_PAGES_COMMA_SEPARATED}}.html"
 eval ls "{$ERROR_PAGES_COMMA_SEPARATED}.html" || (echo Failed to download one or more files.; exit 1)
+
 aws s3 sync . "s3://govuk-app-assets-${GOVUK_ENVIRONMENT}/error_pages/"
+aws s3 sync 503.html "s3://govuk-${GOVUK_ENVIRONMENT}-mirror/error/503.html"
```
theseanything, line 14: Ah ideally these shouldn't be in the mirror bucket at all. We probably want to CloudFront to just pull the HTML from the frontend assets S3 bucket.

theseanything (approved):

PR #2837, opened 2024-12-10
Add GOVUK_CHAT_PROMO_ENABLED env var for finder-frontend

kevindew:
There is a increased promotional push for GOV.UK Chat which involves adding the promotional banner to finder-frontend in addition to the 3 other apps.

It is expected that these banners will be switched off on 20th December (by these env vars) before the code for the banners is removed and these env vars are removed altogether.


davidgisbey (approved):
lgtm!

PR #2848, opened 2024-12-12
Remove Router API and Router Mongo

theseanything:
These are no longer used as Router now populates routes via the Content Store.


nimalank7 (approved):
LGTM!

PR #2854, opened 2024-12-16
Remove scaling up of staging GOV.UK Chat infrastructure

kevindew:
This looks to be scaling that was left behind when load testing GOV.UK Chat in staging and is not necessary for the limited traffic this environment receives.


ianhowell-gds (approved):
LGTM

PR #2858, opened 2024-12-18
Add nginx redirect for /favicon.ico from static

KludgeKML:
- Static now no longer serves the favicon on gov.uk (that's done by frontend), but it was also serving the favicon for assets.publishing.service.gov.uk (which appear in search results), so while static still exists we need to add a redirect to the frontend version of the file. When static is removed entirely, we'll have to find a new solution to this (possibly in the fastly configuration, or maybe in another place in the nginx configuration in this repo - the google site verification and the zendesk logo will also need this).

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2841,6 +2841,10 @@ govukApplications:
           location = /googlec908b3bc32386239.html {
             return 200 'google-site-verification: googlec908b3bc32386239.html';
           }
+          # Favicon now that static isn't serving it
+          location = /favicon.ico {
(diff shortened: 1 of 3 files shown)


kevindew (commented):
This looks good to me. Only minor thing I wondered if there's any scenario where we want to maintain a query string where we'd want to replace the redirect with:

```
return 301 https://www.integration.publishing.service.gov.uk$request_uri;
```

but I can't really think why we'd need that.

Other thing is I imagine it'd be better to first push this out for integration and then have a follow up PR for staging and prod once that is verified - just feels a little delicate given lack of tests to go straight to prod.

kevindew (commented):
Nice one - assume just needs a rebase.

kevindew (approved):

PR #2863, opened 2024-12-20
Define signon-api-token

pezholio:
This defines a new secret for Whitehall to access the newly minted user details endpoint in Signon. 

The secrets are set up in all three AWS environments


pezholio:
Closing this as these are automatically generated by the `sync_token_secrets` task in Whitehall.

PR #2864, opened 2024-12-20
Use the generated Signon API key in Whitehall

pezholio:
I’ve given the “Whitehall” user access to the newly-minted API endpoint in Signon, and run the `signon-sync-token-secrets-to-k8s` job, so this secret should be ready for use in each app.


theseanything:
Why Whitehall needs to access Signon's APIs? What feature does this relate too?

pezholio:
This relates to the newly-minted API endpoint https://github.com/alphagov/signon/pull/3426 that allows us to get users from a central source if truth. This requires a new API "user" that is set up for that endpoint only. It doesn't relate to the OAuth side of signon at all

theseanything (approved):

theseanything (changes requested):
Can you roll this out to integration/staging first.

theseanything (approved):

PR #2860, opened 2024-12-18
Increase Prod Replica Counts for 3 services

dj-maisy:
## What?
This increases the replica counts in Production for the following apps/services:

* Frontend (4 to 6)
* Local Links Manager (3 to 6)
* Locations API - App & Workers (3 to 6)

These increases are expected to be temporary. Requested by @johnpauldickie

PR #2882, opened 2025-01-06
Revert "Increase Prod Replica Counts for 3 services"

dj-maisy:
## What?
This PR reverts alphagov/govuk-helm-charts#2860 which was merged in anticipation of additional traffic/load. We no longer need this surplus capacity so should revert the values to the previous defaults.


KludgeKML (approved):
LGTM

PR #2886, opened 2025-01-09
Change disabling of search autocomplete

csutter:
> [!CAUTION]
> This requires the following PRs to be merged first, or autocomplete will temporarily stop working: 
> * https://github.com/alphagov/govuk_publishing_components/pull/4533
> * https://github.com/alphagov/finder-frontend/pull/3624
> * https://github.com/alphagov/frontend/pull/4569

This was previously managed by rendering a different component in _every_ single rendering application (due to the autocomplete functionality being present in the global header).

This changes it to be a single feature flag on `search-api-v2` instead that controls whether any results are returned, and changes it to be a positive flag instead.


kevindew (commented):
Looks good, but would suggest we probably don't want the superfluous ENABLE_AUTOCORRECT env vars and just want the instructions to tell someone to add it.

kevindew (approved):
Nah, not strong opinions on it. Your points are valid so happy to keep it.

PR #2888, opened 2025-01-09
Remove Kubescape helm chart

theseanything:
We currently have this only deployed in integration and haven't been using it. Removing for now, until we revisit work.


nimalank7 (approved):
LGTM!

PR #2909, opened 2025-01-16
Heatmap for request rate / durations by doc type (government-frontend)

richardTowers:
This is useful for getting a feel for which document types get the most traffic, and which ones are fast / slow.

Dashboard in experimental form here: https://grafana.eks.production.govuk.digital/d/fea5k4nydlhc0b/richard-towersat-government-frontend-heatmaps?folderUid=VE8lzSyVz&orgId=1&from=now-24h&to=now&timezone=browser&var-namespace=apps&var-app=government-frontend&var-controller=content_items&var-action=show

image

### DIFF
--- a/charts/monitoring-config/dashboards/government-frontend-heatmaps.json
+++ b/charts/monitoring-config/dashboards/government-frontend-heatmaps.json
@@ -1,0 +1,261 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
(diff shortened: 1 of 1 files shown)


theseanything:
Do you wanna leave it in the experiments folder for a bit and see how much use/how useful it becomes before adding it to our main collection?

theseanything:
Nit: Wasn't super obvious to me that these are requests to the Frontend applications (as opposed to CDN) - dunno if it's possible to add a title or description somewhere?

richardTowers:
Yeah, I'll pop this in to draft and leave the dash in experiments for now. We can add it to the main collection once it's proved its worth / been improved sufficiently.

PR #2927, opened 2025-01-23
Create config for publisher-on-postgres branch

syed-ali-tw:
[Trello](https://trello.com/c/7TgXoJSL/586-enable-eks-for-postgres-branch-of-publisher)

As an step towards moving Mainstream publisher from Mongo database to Postgres database, we want to be able to spin up a separate app integration from a branch that runs postgres database.
We have created the database [in this PR](https://github.com/alphagov/govuk-infrastructure/pull/1569).

A step pending is to create secrets in aws secrets manager for the database arn, which will be completed before merge.

These changes will only live till the time we need to run postgres in parallel to mongo. As a first, we want to run the app in integration for all our migrations and testing.

### DIFF
--- a/charts/app-config/image-tags/integration/publisher-on-postgres-branch
+++ b/charts/app-config/image-tags/integration/publisher-on-postgres-branch
@@ -1,0 +1,3 @@
+image_tag: 1
+automatic_deploys_enabled: false
+promote_deployment: false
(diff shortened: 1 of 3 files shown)


Tetrino (approved):

dj-maisy (approved):
I've had a quick look at this and can't spot any obvious issues with it. I've also had a look at the ExternalSecret and that looks fine too. 👍 

Plus, it's Integration. What's the worst that could happen? XD

PR #2934, opened 2025-01-23
Update notification settings for Chat

ianhowell-gds:
## What

Update notification settings to reduce the wait times

## Why

The wait times were too long previously


rtrinque (approved):
Looks good

PR #2938, opened 2025-01-24
Update test cronjob alerts for search api

ianhowell-gds:
## What

Update cronjob alerts with additional information and add alert matcher & receiver

## Why

The additional information will help identify the cronjob with the issue by adding a link to the job in ArgoCD.
The new alert matcher & receiver are to make use of this new information.


rtrinque (approved):
looks good

PR #2939, opened 2025-01-24
Set trigger to correct evaluation for cronjob alerts

ianhowell-gds:
## What

Set the evaluation to the correct expression for the cronjobs alert

## Why

The previous setting was a test to ensure the alert triggers. Now we are happy with the alert, this sets it to the correct value.


rtrinque (approved):
Looks good

PR #2940, opened 2025-01-24
Add the repo name to the pub-on-pg app

dj-maisy:
## What?
Argo cannot find the app image for the `publisher-on-postgres` app. This adds the missing `repoName` property so Argo/k8s can find the correct repo.


syed-ali-tw (approved):

dj-maisy:
If we're creating a new ECR Repo for these images, this PR is probably no longer necessary.

PR #2966, opened 2025-02-03
Update cronjob alerts to fire at any time

ianhowell-gds:
## What

Remove alert inhours restriction

## Why

We would like the alert to go to the Slack channel at any time


rtrinque (approved):
Looks good

PR #2977, opened 2025-02-06
De-integrate GOV.UK Chat from www.gov.uk in integration

kevindew:
Trello: https://github.com/alphagov/govuk-helm-charts/pull/2977

Now that we have configured a gone route for www.gov.uk/chat [1] this removes the additional configuration to serve the /chat route with the www loadbalancer.

This removes the additional ALB conditions for router with extra paths for /chat. It removes the nginx redirects to keep people on www rather than the chat.publishing hostname, and turns on signon authentication for chat again, so that it can't be accessed without a signon account.

I've applied this only to integration so far so that I can verify it all works correctly.

I've left the ingress configuration [2] as it seemed every other app with public access has it, however I was a little confused as we didn't seem to add it until we configured the app to be served on www.

[1]: https://github.com/alphagov/publishing-api/pull/3122
[2]: https://github.com/alphagov/govuk-helm-charts/commit/06f61038f801fbee1ad2c4790ee9fb81ac92e8e2

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1617,9 +1617,6 @@ govukApplications:
           schedule: "15 9,11,13,15 * * 1-5"
           timeZone: "Europe/London"
       extraEnv:
-        # remove the need for signon to access frontend of chat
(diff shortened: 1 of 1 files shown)


rtrinque (approved):
Looks good

PR #2985, opened 2025-02-07
Fix chart linting

kentsanggds:


govuk-app-linter.sh
```diff
@@ -0,0 +1,39 @@
...
+  yq 'explode(.)' "./charts/app-config/values-$env.yaml" > "lint-tmp/$env-exploded.yaml"
+
+  for app_index in $(seq 0 "$array_length");do
+    filename="./lint-tmp/$(yq e ".govukApplications[$app_index].name" "./charts/app-config/values-$env.yaml")-chart.yaml"
+
+    echo "$app_index/$array_length: linting $filename";
+
+    yq e ".govukApplications[$app_index].helmValues" "lint-tmp/$env-exploded.yaml" > "$filename";
+
+    output=$(helm lint --quiet -f "$filename" ./charts/generic-govuk-app)
```
samsimpson1, line 27: Bare in mind there are apps in the list which use different charts, like asset-manager

samsimpson1: Maybe something like this
```suggestion
    CHART_PATH=$(yq -r '.chartPath // "charts/generic-govuk-app"' "${filename}")
    output=$(helm lint --quiet -f "$filename" ${CHART_PATH)
```

kentsanggds: I'll try that out

samsimpson1:
One other thing: we're also running kubeconform in that pipeline, and it'd be good if the rendered charts were passed to that too

govuk-app-linter.sh
kentsanggds: Testing it out, doesn't quite work so I'm fixing it

kentsanggds:
I fixed the issue with the chart path, lets get this reviewed again, I'll put the kubeconform check in another PR

samsimpson1 (approved):

PR #2991, opened 2025-02-10
Add service account to Chat in Staging & Production for Bedrock access

ianhowell-gds:
## What 

Add a service account to the Chat pods in Staging & Production 

## Why

This will provide them with permission to access the Bedrock service in AWS


rtrinque (approved):
Looks good

PR #2995, opened 2025-02-12
Add govuk-graphql

richardTowers:
This is an experiment, so only going to integration for now.

The app exposes an API and a user interface, both authenticated behind GDS SSO.

The API includes draft content, which is potentially sensitive, but any risk of information being exposed unnecessarily is adequately mitigated:

- The app uses gds-sso, like other apps. So only users with access to the app in Signon can see the UI, and API calls require bearer tokens (or session cookies)
- Sensitive data (like access limited drafts) is sanitized in integration in any case.

The other main weirdness here is that I'm reusing publishing-api's database credentials. I think this is okay for a test in integration, but for a higher environment we'd at the very least want a read-only user, and probably a whole database read replica so we don't have to worry so much about performance impacts on the primary.

I've already added the secrets to Signon in integration, and sync'd them to k8s.

### DIFF
--- a/charts/app-config/image-tags/integration/govuk-graphql
+++ b/charts/app-config/image-tags/integration/govuk-graphql
@@ -1,0 +1,3 @@
+image_tag: latest
+automatic_deploys_enabled: true
+promote_deployment: false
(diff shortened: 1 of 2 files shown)


jasonBirchall (approved):
LGTM 🚀

PR #3007, opened 2025-02-24
Revert "Set publishing-api app to use serverless valkey"

samsimpson1:
Turns out Sidekiq doesn't support Redis in cluster mode, so serverless ElastiCache can't work

Reverts alphagov/govuk-helm-charts#3006


AgaDufrat (approved):
😞

PR #3015, opened 2025-02-26
Bump search-api's redis memory limit to 3GB in integration

samsimpson1:
It's running out of memory due to the republishing job that's running


ryanb-gds (approved):
Thanks Sam. It would be really neat if the number of workers autoscaled when Redis is getting a bit full so that we could avoid these sorts of issues - one for the platform backlog 😉

PR #3030, opened 2025-03-05
Resume RabbitMQ message processing now that the govuk publishing queue is cleared

ryanb-gds:
The government indexing queue processor is much slower because it writes to stdout for each message, so we need our 8 replicas to get good throughput.


dj-maisy (approved):
🚀

kevindew (approved):
Nice

PR #3031, opened 2025-03-05
Reduce number of Search API worker replicas to 2

ryanb-gds:
Now that we have cleared our backlog of search indexing jobs, we can reduce the number of worker pod replicas to 2 again. This should be more than enough to handle our regular indexing load.


kevindew (approved):
Looks good

PR #3051, opened 2025-03-12
Ephemeral monitoring config

samsimpson1:
This PR includes a lot of changes to monitoring-config. It is all so the chart will work on ephemeral clusters.

It does the following:

* Disables AlertManger config since we don't care about alerting on an ephemeral cluster
* Creates a new templated values file for kube-prometheus-stack
* Creates a new templated values file for tempo
* Adds a conditional that makes ephemeral envs use production helm chart versions
* Disables the Fastly exporter for Prometheus since we aren't using Fastly services here

https://github.com/alphagov/govuk-infrastructure/issues/1744


dj-maisy (approved):
Looks good, cap'n. 🚢

PR #3074, opened 2025-03-17
Add read replica for Publishing API in integration

brucebolt:
This will be a parallel deployment of the Publishing API application in intgegration, used solely for serving GraphQL queries from frontend applications.

As this will not be receiving any write requests, it does not need access to Redis or any workers.

The `hostReplica` secret has already been added through [the AWS Console](https://eu-west-1.console.aws.amazon.com/secretsmanager/secret?name=govuk%2Fpublishing-api%2Fpostgres&region=eu-west-1). The other credentials remain the same as the main version of the database.

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2650,6 +2650,38 @@ govukApplications:
               name: publishing-api-bigquery
               key: client-secret
 
+  - name: publishing-api-read-replica
+    repoName: publishing-api
+    helmValues:
+      arch: arm64
+      appResources:
+        limits:
+          cpu: 2
+          memory: 1.5Gi
+        requests:
+          cpu: 10m
+          memory: 512Mi
+      nginxClientMaxBodySize: 2M
+      dbMigrationEnabled: false
+      extraEnv:
+        - name: GDS_SSO_OAUTH_ID
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-publishing-api
+              key: oauth_id
+        - name: GDS_SSO_OAUTH_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-publishing-api
+              key: oauth_secret
+        - name: GOVUK_CONTENT_SCHEMAS_PATH
+          value: /app/content_schemas
+        - name: DATABASE_URL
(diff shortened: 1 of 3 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -2652,6 +2652,40 @@ govukApplications:
...
+            secretKeyRef:
+              name: signon-app-publishing-api
+              key: oauth_id
+        - name: GDS_SSO_OAUTH_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-publishing-api
+              key: oauth_secret
+        - name: EVENT_LOG_AWS_BUCKETNAME
+          value: govuk-publishing-api-replica-event-log-integration
```
yndajas, line 2680: **question:** does this need removing after [this change](https://github.com/alphagov/publishing-api/pull/3216)?

charts/app-config/values-integration.yaml
```diff
@@ -2652,6 +2652,40 @@ govukApplications:
...
+        - name: GDS_SSO_OAUTH_ID
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-publishing-api
+              key: oauth_id
+        - name: GDS_SSO_OAUTH_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: signon-app-publishing-api
+              key: oauth_secret
```
yndajas, line 2678: **question:** also curious if we need these and the schemas environment variable above. I don't think they're used by GraphQL, but maybe the app needs them for basic functionality

charts/app-config/values-integration.yaml
brucebolt: I'd removed that line, but never pushed the change 🤦‍♂️

charts/app-config/values-integration.yaml
brucebolt: I was wondering this too.

I left the SSO OAUTH secret as we'll still be using Signon to authenticate API users, so assumed this would be required.

For the content schemas path, I wasn't sure whether it was needed, but could see it being used in the `govuk_schemas` gem, which may or may not be required for core Publishing API functionality.

yndajas (approved):

PR #3076, opened 2025-03-17
Disable `uploadAssets` for Publishing API read replica

brucebolt:
This is a API application, so there are no assets to upload.

Adding this configuration to match the main Publishing API deployment, as this was missed in https://github.com/alphagov/govuk-helm-charts/pull/3074.


yndajas (approved):
Oops!

PR #3077, opened 2025-03-18
Configure Publishing API read replica environment variables

brucebolt:
The replica of this application needs access to both the primary and replica databases, so providing both through environment varaibles.

Additionally, it needs to have the `RAILS_ENV` overwritten, to tell the application to operate in replica mode.

This will allow https://github.com/alphagov/publishing-api/pull/3219 to be completed and merged.


yndajas (approved):
Might be useful to mention the use case in the commit message briefly, so we can work out if it's still needed in future, but makes sense based on your explanation in Slack!

PR #3089, opened 2025-03-25
Modify Grafana queries on Sidekiq dashboard

MuriloDalRi:
This dashboard was showing `no data` on the memory and disk usage panels when a specific application
was selected.

The issue arose because app instances are named with a `-worker` suffix (e.g., `asset-manager-worker`),
while their corresponding Redis pods follow a different pattern: `<app>-redis-<random>`.
This mismatch meant that filtering by `$app` would not correctly match the associated Redis instances.

https://github.com/alphagov/govuk-infrastructure/issues/1816

### DIFF
--- a/charts/monitoring-config/dashboards/sidekiq-queues.json
+++ b/charts/monitoring-config/dashboards/sidekiq-queues.json
@@ -24,15 +24,14 @@
       }
     ]
   },
-  "description": "",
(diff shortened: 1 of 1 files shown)


dj-maisy (approved):
Looks good to me! 🚀

PR #3091, opened 2025-03-25
Emphasise app and environment in deploy failure alerts

yndajas:
It's useful for developers to monitor the deploy alerts channel in Slack to be aware of any failures to promote deployments for their apps. This channel can sometimes get a little crowded, and it can be a challenge to see which app and environment a failed workflow run relates to at a glance

This bolds these two variables to make the alerts easier to parse both in:
- the post-sync/post-deploy workflow, which is for promoting deployment to the next environment - failures usually come from end-to-end tests
- the deploy-image workflow, which has a similar template, so it would seem useful to keep them consistent

Longer term, it might be better to route these alerts to the relevant team channel, but this is a quicker win in the short term

### DIFF
--- a/charts/argo-services/templates/workflows/deploy-image/workflow.yaml
+++ b/charts/argo-services/templates/workflows/deploy-image/workflow.yaml
@@ -62,7 +62,7 @@ spec:
                       "type": "section",
                       "text": {
                           "type": "mrkdwn",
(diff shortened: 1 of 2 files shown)


nicholsj:
Relates to https://github.com/alphagov/govuk-infrastructure/issues/1715

AgaDufrat (approved):

PR #3096, opened 2025-03-26
Search API v2: New recording rules; remove alerts

csutter:
- Remove broken and not quite working right alerts for invariants
- Add recording rules and rework success rate alerts modelled after
  Whitehall


ianhowell-gds (approved):
Looks good

PR #3101, opened 2025-04-01
Fix cluster-autoscaler helm template error in ephemeral bootstrap chart

samsimpson1:
The `helm-release` template needs `deepCopy` because it uses `.Files.Get` from the global state, and it won't be available unless its copied into the arguments passed to the template.

This issue is causing the bootstrap chart to fail to install at the moment.


MuriloDalRi (approved):

nimalank7 (approved):
LGTM!

PR #3106, opened 2025-04-02
Configure Prometheus to monitor DGU in ephemeral

nimalank7:
Description:
- Currently Prometheus doesn't scrape DGU because podMonitor and serviceMonitor isn't configured properly
- As part of https://github.com/alphagov/govuk-infrastructure/issues/1744


kentsanggds (approved):
LGTM

PR #3109, opened 2025-04-03
Turn off default alerting rules for API server on ephemeral Grafana

nimalank7:
Description:
- Copy the configuration over from production configuration file
- `kube-prometheus-stack` configures some default alerting rules for the API server. We've disabled them in AlertManager so turn them off in Grafana as well
- As part of https://github.com/alphagov/govuk-infrastructure/issues/1744


kentsanggds (approved):
LGTM

PR #3112, opened 2025-04-03
Add step to warn about govuk_content_block_tools failures

MuriloDalRi:
This is a new step that will send slack messages about E2E test failures to another channel. The channel will be a test one for now and later updated to the be the Content Modelling channel.


AgaDufrat (approved):
🤞

PR #3117, opened 2025-04-04
Add config for ephemeral grafana to set the timezone and date formats

kentsanggds:
This should bring the ephemeral grafana more in line with the production grafana

We've decided not to create the grafana admin user as it's not really used by anyone as changes to grafana are managed in charts.

https://github.com/alphagov/govuk-infrastructure/issues/1744


nimalank7 (approved):
LGTM!

PR #3119, opened 2025-04-04
Search API v2: Improve rate recording rules

csutter:
These use `clamp_min` to avoid division by zero, but that has the side
effect of distorting values when the actual total request rate is below
1 (by obviously clamping it up to 1).

Instead, this avoids division by zero by declaring that if there are no
requests at all, the success rate is 1 by definition.

This also replaces the use of derived metrics as input series with the
original raw metrics to avoid dreaded `duplicate sample for timestamp`
errors with the new tests.


ianhowell-gds (approved):
Looks good

PR #3120, opened 2025-04-04
Add read replica for Publishing API in production

yndajas:
[Trello](https://trello.com/c/ZwWoo7Mg/1640-set-up-the-read-replica-in-production)

This will be a parallel deployment of the Publishing API application in production, used solely for serving GraphQL queries from frontend applications

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2469,6 +2469,49 @@ govukApplications:
               name: publishing-api-bigquery
               key: client-secret
 
+  - name: publishing-api-read-replica
+    repoName: publishing-api
+    helmValues:
+      arch: arm64
+      appResources:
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -2469,6 +2469,49 @@ govukApplications:
...
+      arch: arm64
+      appResources:
+        limits:
+          cpu: 2
+          memory: 1.5Gi
+        requests:
+          cpu: 20m
+          memory: 768Mi
+      nginxClientMaxBodySize: 2M
+      dbMigrationEnabled: true
```
brucebolt, line 2484: Does this need to be disabled since the migrations will run in the main app deployment?

brucebolt (approved):

PR #3123, opened 2025-04-07
Search API v2: Add mid-term alerts

csutter:
These are only for search, not autocomplete (as that is less important). Fire an alert if search success rate is less than 99.9% in a rolling 1h period for more than 2h.

- Add recording rules for mid-term (`search_api_v2:search_requests:rate1h` and `search_api_v2:search_successful_requests:ratio_rate1h`)
- Add alerting rule for mid-term (`SearchDegradedMid`)


ianhowell-gds (approved):
Looks good

PR #3124, opened 2025-04-07
Search API v2: Simplify alerting tests

csutter:
- Use more naturally divisible numbers and remove rounding
- Make labelling consistent between tests
- Consolidate some positive and negative tests into one and check at different times for different conditions


ianhowell-gds (approved):
Looks ok

PR #3126, opened 2025-04-09
Add sanity check to post-sync Argo Workflow to ensure only versioned releases are promoted

samsimpson1:
This should prevent branch deploys from being unintentionally promoted beyond the environments they were originally deployed in to.

https://github.com/alphagov/govuk-infrastructure/issues/1965


nimalank7 (approved):
LGTM!

PR #3127, opened 2025-04-09
Strengthen image tag format check

richardTowers:
This uses bash's built in support for regular expressions to check the full IMAGE_TAG pattern, instead of just checking that the first character is v.

Only checking the first character still leaves some risk that a branch like "very-sensitive-change-do-not-deploy-to-production" could be accidentally deployed.

Checking that this does the right thing in bash:

    % IMAGE_TAG=v38274193
    % if [[ ! "$IMAGE_TAG" =~ ^v[0-9]+$ ]]; then echo "$0 - not promoting image tag ($IMAGE_TAG) as it doesn't match the required pattern (^v[0-9]+$)"; fi
    % IMAGE_TAG=vanilla
    % if [[ ! "$IMAGE_TAG" =~ ^v[0-9]+$ ]]; then echo "$0 - not promoting image tag ($IMAGE_TAG) as it doesn't match the required pattern (^v[0-9]+$)"; fi
    bash - not promoting image tag (vanilla) as it doesn't match the required pattern (^v[0-9]+$)

### DIFF
--- a/charts/argo-services/scripts/check-for-promotion.sh
+++ b/charts/argo-services/scripts/check-for-promotion.sh
@@ -1,9 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Ensure the image tag starts with a v
-if [ "${IMAGE_TAG:0:1}" != "v" ]; then
-  echo "false"
+if [[ ! "$IMAGE_TAG" =~ ^v[0-9]+$ ]]; then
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):

richardTowers:
Nim pointed out that IMAGE_TAG is a hash, not a v123 git tag. So this won't work. I'm not sure how the "starts with v" thing works, but maybe it's a hash prefixed with v?

nimalank7 (approved):
LGTM!

PR #3129, opened 2025-04-10
Add cascading delete finaliser to all Application resources

samsimpson1:
This adds a finaliser to all ArgoCD Application resources that causes ArgoCD to cascade delete child resources of these applications. This will allow us to easily remove all resources from the cluster when shutting down an ephemeral cluster.

https://github.com/alphagov/govuk-infrastructure/issues/1972


nimalank7 (approved):
LGTM!

PR #3131, opened 2025-04-14
Add new Graphql dashboard

callumknights:
Used for monitoring the performance of Graphqlified pages.

Important to preserve here to prevent loss of this monitoring while testing.

Trello: https://trello.com/c/YgjscovT/1606-visualise-response-times


JonathanHallam (approved):
👍

PR #3133, opened 2025-04-15
Update publish-api to use test redis instances in Integration

AgaDufrat:
This was missed causing 
`Unknown URL scheme: "publishing-api-valkey.integration.govuk-internal.digital:6379" (ArgumentError) raise ArgumentError, "Unknown URL scheme: #{url.inspect}" `


ianhowell-gds (approved):
LGTM

PR #3140, opened 2025-04-16
Bump version of cluster-secret-store chart

samsimpson1:
0.1.1 seems to not have been deployed to the chart repo properly as it still contains a v1alpha1 ClusterSecretStore resource, and that API version has been removed from external-secrets


nimalank7 (approved):
LGTM!

PR #3153, opened 2025-04-25
Support indivudual document types in graphql dashboard

callumknights:
Add individual document types to visualisations for both graphql and content-store

Also split the visualisations for request times out into individual averages to reduce visual clutter, and nested them inside titled rows.

Before: 
image

After:
image

^Heavily zoomed out so I can capture the differences in one screenshot


mike3985 (approved):
📈

PR #3154, opened 2025-04-28
Set PrunePropagationPolicy=orphan on kube-prometheus-stack resources

samsimpson1:
These resources are going to be moved to their own chart

https://github.com/alphagov/govuk-infrastructure/issues/2044


sihugh (approved):
This should not delete child resources 👍

PR #3155, opened 2025-04-28
Search API v2: Purge old user events daily

csutter:
This triggers the daily task to clear out all user events from the final week of the retention period.

We're calling this task daily even though it deals with a week's worth of data so we have some breathing room if it fails at a really inconvenient time (say, in the middle of long holidays) without violating our data protection agreements.

see https://github.com/alphagov/search-api-v2/pull/424 for the implementation.


davidgisbey (approved):
lgtm!

PR #3164, opened 2025-04-29
Switch `search-api` over to new Elasticache instance

ianhowell-gds:
## What

Switch `search-api` service over to new Elasticache instance `search-api-valkey.integration.govuk-internal.digital`

## Why

So that testing can be carried out for this service


rtrinque (approved):
Looks good

PR #3165, opened 2025-04-29
Reduce `replicaCount` for publishing-api in integration 

pezholio:
We can now safely reduce this number (changed in https://github.com/alphagov/govuk-helm-charts/pull/3142) in integration now Publishing API is less busy


nicholsj (approved):
LGTM thanks

PR #3169, opened 2025-04-29
Add table for measuring slow responses

JonathanHallam:
I made a new table that takes the slowest ten document types and shows them to you. I like it! 

Screenshot 2025-04-29 at 14 33 19

PR #3170, opened 2025-04-29
Same thing as the last commit but for graphQL/Publishing API

JonathanHallam:
Adds a similar graph than in http://github.com/alphagov/govuk-helm-charts/pull/3169 but for graphql

### DIFF
--- a/charts/monitoring-config/dashboards/graphql-stats.json
+++ b/charts/monitoring-config/dashboards/graphql-stats.json
@@ -76,6 +76,30 @@
                 "value": true
               }
             ]
+          },
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "document_type"
+            },
+            "properties": [
+              {
+                "id": "displayName",
+                "value": "Document Type"
+              }
+            ]
+          },
+          {
+            "matcher": {
+              "id": "byName",
+              "options": "Value"
+            },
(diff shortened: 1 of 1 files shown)


charts/monitoring-config/dashboards/graphql-stats.json
```diff
@@ -118,7 +118,98 @@
...
-      "title": "10 Slowest document types (Last 6 hours)",
+      "title": "10 Slowest Content Store document types (Last 6 hours)",
+      "type": "table"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "prometheus"
+      },
+      "description": "10 slowest average response times to content store, broken up by document type",
```
callumknights, line 129: Description and title 8 lines above need changing, both refer to content store.

JonathanHallam: oops, good spot. Fixed!

JonathanHallam: The title you can see above is for the other panel, so that ones right

callumknights: Yep my bad, silly Json ordering

yndajas (approved):

PR #3176, opened 2025-05-01
Increase upload file size on Licensify from 10MB to 50MB

dgm34:
This fixes an issue raised by @steven-radford at today's GDS-APTO catch up. The file size limit on Licensify's nginx is 200MB so we could even consider making this 200MB, but i've erred on the side of caution and gone a bit lower out of an abundance of caution.

### DIFF
--- a/charts/app-config/templates/router-nginx-config.tpl
+++ b/charts/app-config/templates/router-nginx-config.tpl
@@ -138,7 +138,7 @@ http {
       proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
 
       # Allow large uploads to licensify-frontend
-      client_max_body_size 10M;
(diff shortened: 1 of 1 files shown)


richardTowers (approved):
Looks good to me, as far as I can tell. Might be worth someone from platform engineering just confirming that a 50MB upload isn't going to cause their servers to catch filre or anything.

dgm34:
> Looks good to me, as far as I can tell. Might be worth someone from platform engineering just confirming that a 50MB upload isn't going to cause their servers to catch filre or anything.

Valid. Paging @dj-maisy and @samsimpson1 for their 2c

dj-maisy (approved):
A 50MB file sounds impressive for the needs of licensify, but if it's needed, then it's needed! 🚀

PR #3186, opened 2025-05-07
Search API v2: Add long term success rate alert

csutter:
This adds a long term (24h) rate, ratio, and alert to the existing short (5m) and medium (1h) term ones.


ianhowell-gds (approved):
Looks good

PR #3189, opened 2025-05-08
Combine the graphql and content store panels and add a percenterator

JonathanHallam:
Added a new panel that combines two old ones, and a new column that shows a percentage difference between the two things
Screenshot 2025-05-08 at 14 56 41

### DIFF
--- a/charts/monitoring-config/dashboards/graphql-stats.json
+++ b/charts/monitoring-config/dashboards/graphql-stats.json
@@ -27,7 +27,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 6726,
+  "id": 6815,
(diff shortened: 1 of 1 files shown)


charts/monitoring-config/dashboards/graphql-stats.json
```diff
@@ -1310,5 +1223,5 @@
   "timezone": "browser",
   "title": "Graphql stats",
   "uid": "aeh00wtwwg8owc",
-  "version": 5
+  "version": 1
```
yndajas, line 1226: **issue:** should this version be 6?

JonathanHallam: weird, fixed!

brucebolt (approved):

PR #3193, opened 2025-05-09
Add request numbers to graphql stats

JonathanHallam:
Adds in number of requests by content type to graphql dashboard

Screenshot 2025-05-09 at 16 27 36

### DIFF
--- a/charts/monitoring-config/dashboards/graphql-stats.json
+++ b/charts/monitoring-config/dashboards/graphql-stats.json
@@ -74,6 +74,45 @@
                 "value": "percent"
               }
             ]
+          },
+          {
+            "matcher": {
+              "id": "byName",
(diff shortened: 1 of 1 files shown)


brucebolt (approved):
Looks good to me.

Not for this PR, but maybe something to think about: I wonder if there's a way we can make the table match the time period specified in the selector at the top right of the dashboard, instead of always being the last 6 hours?

JonathanHallam:
> Looks good to me.
> 
> Not for this PR, but maybe something to think about: I wonder if there's a way we can make the table match the time period specified in the selector at the top right of the dashboard, instead of always being the last 6 hours?

Yeah i did look at this when I first made it and I couldn't get it to work. Probably should revisit it though

PR #3200, opened 2025-05-12
Only remove the e2e test cronjob and config settings

kentsanggds:
Previous PR to remove the e2e tests caused a problem with the post sync e2e tests, this will just target the e2e test cronjobs.

https://github.com/alphagov/govuk-infrastructure/issues/2082


nimalank7:
You can also remove https://github.com/alphagov/govuk-helm-charts/blob/377eb1d38a42aec0d3992da01d3f1ba6be60e5e9/charts/govuk-e2e-tests/values.yaml#L6

PR #3203, opened 2025-05-14
Add ServiceMonitor for Argo Workflows

samsimpson1:
This will allow us to monitor the state of Argo Workflows, such as how many workflows are failing

The new service is required so that all metrics endpoints are scraped if there are multiple workflow controllers running (we have this in prod)

https://github.com/alphagov/govuk-infrastructure/issues/2138


dj-maisy (approved):
LGTM. 👍

PR #3224, opened 2025-05-22
Search API v2: Another Alertmanager template fix

csutter:
This fixes unbalanced/plain wrong escaping and removes a stray `]`.


nimalank7 (approved):
LGTM!

PR #3243, opened 2025-06-05
Remove Smart Survey env vars for GOV.UK Chat

chaecramb:
Smart Survey is no longer used in GOV.UK Chat following the completion of the private beta.


davidgisbey (approved):
looks good 👍

PR #3248, opened 2025-06-09
Remove Notify related env vars and secret for GOV.UK Chat

davidgisbey:
## Description

We've removed out Notify integration from GOV.UK Chat, so we no longer need the associated environment variables and secret.

## Trello card

https://trello.com/c/RxzQTX8Z/2534-remove-notify-integration


jackbot (approved):
LGTM :+1:

Do we need to delete the `GOVUK_NOTIFY_API_KEY` secret from secrets manager too? And from Heroku.

PR #3255, opened 2025-06-11
Add behaviour to HPA config for Chat

ianhowell-gds:
## What

Add behaviours for scaling up and scaling down to the HPA configuration

## Why

This is to slow the scaling process while we test it's functionality


charts/app-config/values-integration.yaml
```diff
@@ -1682,6 +1682,19 @@ govukApplications:
                   target:
                     type: Value
                     value: 10
+            behaviour:
```
samsimpson1, line 1685:
```suggestion
            behavior:
```

samsimpson1 (approved):

PR #3256, opened 2025-06-12
Make bars 1 hour wide

JonathanHallam:
This makes the bars on the per hour graphql bar chart 1 hour wide to more easily visualise the number of requests

## Before
Screenshot 2025-06-12 at 10 22 41

## After
Screenshot 2025-06-12 at 10 22 29


marianasthiago-gds:
Hey there, I had a think about this, and even though the chart with bars provide a clearer view of the number of requests that happen in an hour, it might also mask smaller trends during that time. I think the should probably keep the previous suggestion.

PR #3258, opened 2025-06-12
Make ArgoCD ignore differences in replicas value for resources with HPA enabled

samsimpson1:


ianhowell-gds (approved):
That looks ok to me

PR #3264, opened 2025-06-13
Add repoName to the generic-govuk-app selector labels to enable release app to find matching apps.

kentsanggds:
Some apps are deployed on the cluster that fo not use their repo name as their release name, so publish the label to enable the k8s api to select it.

https://github.com/alphagov/govuk-infrastructure/issues/2237

### DIFF
--- a/charts/generic-govuk-app/templates/_helpers.tpl
+++ b/charts/generic-govuk-app/templates/_helpers.tpl
@@ -48,6 +48,7 @@ Selector labels
 */}}
 {{- define "generic-govuk-app.selectorLabels" -}}
 app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/repoName: {{ .repoName | default .Release.Name  }}
 {{- end }}
 
 {{/*
--- a/charts/generic-govuk-app/templates/_helpers.tpl
+++ b/charts/generic-govuk-app/templates/_helpers.tpl
@@ -41,6 +41,7 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/arch: {{ default "amd64" .Values.arch }}
+app.kubernetes.io/repoName: {{ .repoName | default .Release.Name  }}
 {{- end }}
 
 {{/*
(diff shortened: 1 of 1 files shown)


charts/generic-govuk-app/templates/_helpers.tpl
```diff
@@ -48,6 +48,7 @@ Selector labels
 */}}
 {{- define "generic-govuk-app.selectorLabels" -}}
 app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/repoName: {{ .repoName | default .Release.Name  }}
```
samsimpson1, line 51: This belongs in the normal labels, not the selector ones

kentsanggds: ok I'll move it up

charts/generic-govuk-app/templates/_helpers.tpl
```diff
@@ -41,6 +41,7 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/arch: {{ default "amd64" .Values.arch }}
+app.kubernetes.io/repoName: {{ .repoName | default .Release.Name  }}
```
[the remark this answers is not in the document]
samsimpson1: The `app.kubernetes.io` prefix on labels is for [well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/), and shouldn't be used for custom ones

AgaDufrat:
The Release app still needs to know the k8s name to generate links to app's ArgoCD page, you are planning to add. It maybe better to store k8s name in Release DB instead (assuming we want to move away from [/repos.json](https://docs.publishing.service.gov.uk/repos.json)). 
Unless there're other use cases.

kentsanggds:
The k8s app name should be available after the chart is found when the `repository-name` label is available as it is being exposed as a label - https://github.com/alphagov/govuk-helm-charts/blob/597bd1d7fdc8653a051995ce1104a0b69e5fc334/charts/generic-govuk-app/templates/deployment.yaml#L9-L10

AP-Hunt (approved):

PR #3275, opened 2025-06-19
♻️ Use externalDomainSuffix instead of incorrect externalDomainSuffix

jasonBirchall:
I misinterpreted the difference between environments. One is the publishing domain, one is gov.uk


dj-maisy (approved):
🚀

PR #3283, opened 2025-06-23
✨ Add Slack template and trigger on failure

jasonBirchall:
If an argocd sync fails or is degraded this notification, after the slack webhook is approved, will allow us to customise our triggered alarms. I thought of sending the alerts to the low-priority channel as we don't actually know how this will appear.


jasonBirchall:
I can't get a slack webhook approved by our sysadmins so I'll close this off. 😢

PR #3304, opened 2025-07-03
Sum total GraphQL requests by schema name

richardTowers:
| Before | After |
|---|---|
| image | image |

### DIFF
--- a/charts/monitoring-config/dashboards/graphql-stats.json
+++ b/charts/monitoring-config/dashboards/graphql-stats.json
@@ -1371,10 +1371,10 @@
           },
           "editorMode": "code",
           "exemplar": true,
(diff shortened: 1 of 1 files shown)


charts/monitoring-config/dashboards/graphql-stats.json
```diff
@@ -1490,12 +1490,12 @@
           "disableTextWrap": false,
           "editorMode": "code",
           "exemplar": true,
-          "expr": "sum by(job) (increase(http_requests_total{namespace=\"apps\", job=~\"publishing-api-read-replica\", controller=\"graphql\", action=\"execute\"}[1h]))",
+          "expr": "sum by(document_type) (increase(http_requests_total{namespace=\"apps\", job=~\"publishing-api-read-replica\", controller=\"graphql\", action=\"execute\"}[1h]))",
```
JonathanHallam, line 1493: I think this should probably be schema name rather than document type so that news articles are lumped together

PR #3311, opened 2025-07-07
publishing-api: revert to using in-cluster Redis in integration, staging

AP-Hunt:
We want to make all of the Redis usage across all of the apps consistent once again. We previously moved some integration and staging deployments to using Valkey from AWS Elasticache.

Part of https://github.com/alphagov/govuk-infrastructure/issues/2331


AP-Hunt:
I went into the integration and staging clusters and cleared out the existing Redis instances by running `FLUSHALL`

PR #3320, opened 2025-07-08
Add separator --- to fix the release rolebinding

kentsanggds:
Without the separator only the last block of yaml will be applied when iterating through the range of values. 
This is why the rolebinding was created for the `licensify` namespace but not the `apps` namespace.

https://github.com/alphagov/govuk-infrastructure/issues/2237


jfharden (approved):
That's an easy one to miss in a templated chart! At least the fix was simple :)

jfharden:
Hmm the lint check is failing, feel free to ping me when you need a re-review

kentsanggds:
Seems to be something weird with this branch, so I'm going to create another PR instead

PR #3325, opened 2025-07-08
Bring in Fastly Dashboard from previous PR

dj-maisy:
## What?
This Dashboard was lost in a reverted PR - let's pull it back in because it could actually be useful. And I want to use it as an example Dashboard for the `HighRequestRate` alert.


AP-Hunt (approved):
This will probably be useful for what I'm working on too!

PR #3330, opened 2025-07-09
Add Publisher's postgres databases to backup chronjobs

Tetrino:
As we're migrating to postgres we need to enable backups. This commit does not include the removal of the current mongodb backup job as we wish for that to stay in place for the moment until migration is confirmed stable.

https://trello.com/c/omm4fVel/


mtaylorgds (approved):
LGTM

PR #3337, opened 2025-07-11
Chat: disable OpenSearch syncing to test environment

jackbot:
We're having some data issues with our OpenSearch cluster for GOV.UK
Chat.

The production/staging/integration environments aren't being used at the
moment but the test environment is. So we want to temporarily disable
the nightly syncing of the production -> test env while we sort things
out.


kevindew (approved):
Nice

PR #3339, opened 2025-07-14
[PP-7253] Add `publishing-api` to Plek unprefixable hosts

brucebolt:
Plek adds `draft-` to the start of the hostname for Publising API when running in the draft stack.

There is only one instance of Publishing API for both draft and live, so we need to specify this application as 'unprefixable' so this prefix is no longer added.


charts/app-config/templates/env-configmap.yaml
```diff
@@ -30,7 +30,7 @@ data:
   {{- end }}
   GOVUK_SIDEKIQ_JSON_LOGGING: "1"
   GOVUK_WEBSITE_ROOT: https://www.{{ .Values.externalDomainSuffix }}
-  PLEK_UNPREFIXABLE_HOSTS: account-api,asset-manager,feedback,licensify,local-links-manager,locations-api,places-manager,search-api,signon
+  PLEK_UNPREFIXABLE_HOSTS: account-api,asset-manager,feedback,licensify,local-links-manager,locations-api,places-manager,publishing-api,search-api,signon
```
richardTowers, line 33: Do we need `publishing-api-read-replica` in here too now?

brucebolt: Yes, I think we do.  I'll add that now.

richardTowers: Yeah, I was going to say the same thing... Feels like a bit of a yak shave to fix that though.

PR #3340, opened 2025-07-14
Reenable OpenSearch syncing to test environment

jackbot:
We disabled this recently while we sorted out the data in the production
OpenSearch instance. This has been done now, so we can resume the daily
syncing.


davidgisbey (approved):
lgtm!

PR #3347, opened 2025-07-14
Change Filter String Builder, $first was invalid.

dj-maisy:
## What?
So the previous template was apparently not valid - even though it passed `helm template .`, Alertmanager did not like the use of $first := true in the template, so have attempted to simplify and use the range $index which Alertmanager's documentation claims to be valid.

### DIFF
--- a/charts/kube-prometheus-stack-bootstrap/templates/_slack-helpers.tpl
+++ b/charts/kube-prometheus-stack-bootstrap/templates/_slack-helpers.tpl
@@ -8,12 +8,13 @@
 
 {{- define "slack.filterstring" -}}
 {{`{{- if .Labels.SortedPairs -}}`}}
-{{`{`}}{{`- $first := true -}`}}
-{{`{{- range .Labels.SortedPairs -}}`}}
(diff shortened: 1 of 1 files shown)


charts/kube-prometheus-stack-bootstrap/templates/_slack-helpers.tpl
```diff
@@ -8,12 +8,13 @@
 
 {{- define "slack.filterstring" -}}
 {{`{{- if .Labels.SortedPairs -}}`}}
-{{`{`}}{{`- $first := true -}`}}
-{{`{{- range .Labels.SortedPairs -}}`}}
-  {{`{{- if not $first -}},{{- end -}}`}}
-  {{`{{- $first = false -}}`}}
-  {{`{{- .Name -}}={{- printf "%q" .Value -}}`}}
-{{`{{- end -}}`}}{{`}`}}
+%7B
```
AP-Hunt, line 11: Did you intend to include `%7B`?

AP-Hunt (approved):

PR #3365, opened 2025-07-24
Add scaleup & scaledown behaviour for Chat Autoscaling

ianhowell-gds:
## What

Add behaviour characteristics for Chat autoscaling

## Why

To smooth out the scaling in and out of govuk-chat pods


rtrinque (approved):
Looks good to me

PR #3367, opened 2025-07-28
Add Publisher-on-PG app to Production and Staging

Tetrino:
Mirrors the config from Integration to enable migration on the main environements.

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2330,6 +2330,122 @@ govukApplications:
         - name: GOOGLE_TAG_MANAGER_ID
           value: *publishing-bootstrap-gtm-id
 
+  - name: publisher-on-pg
+    helmValues:
+      arch: arm64
+      appResources:
+        limits:
+          cpu: 1
+          memory: 1.5Gi
+        requests:
+          cpu: 10m
+          memory: 512Mi
+      dbMigrationEnabled: true
+      workers:
+        enabled: true
+      workerResources:
+        limits:
+          cpu: 1
+          memory: 1Gi
+        requests:
+          cpu: 10m
+          memory: 200Mi
+      redis:
+        enabled: true
+      cronTasks:
+        - name: reports-generate
+          task: "reports:generate"
+          schedule: "0 * * * *"
+        - name: publish-missed-scheduled-editions
+          task: "editions:publish_missed_scheduled_editions"
+          schedule: "11 1 * * *"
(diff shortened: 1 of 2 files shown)


charts/app-config/values-production.yaml
```diff
@@ -2330,6 +2330,127 @@ govukApplications:
...
+      workerResources:
+        limits:
+          cpu: 1
+          memory: 1Gi
+        requests:
+          cpu: 10m
+          memory: 200Mi
+      redis:
+        enabled: true
+      cronTasks:
```
mtaylorgds, line 2355: I see that this config doesn't have a `mail-fetcher` cron task—is that correct?

charts/app-config/values-production.yaml
```diff
@@ -2330,6 +2330,127 @@ govukApplications:
...
+          valueFrom:
+            secretKeyRef:
+              name: signon-token-publisher-link-checker-api
+              key: bearer_token
+        - name: LINK_CHECKER_API_SECRET_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: publisher-link-checker-api-callback-token
+              key: LINK_CHECKER_API_SECRET_TOKEN
+        - name: MONGODB_URI
```
mtaylorgds, line 2425: Does this app need the `MONGODB_URI`?

charts/app-config/values-production.yaml
```diff
@@ -2330,6 +2330,127 @@ govukApplications:
...
+        - name: LINK_CHECKER_API_BEARER_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: signon-token-publisher-link-checker-api
+              key: bearer_token
+        - name: LINK_CHECKER_API_SECRET_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: publisher-link-checker-api-callback-token
+              key: LINK_CHECKER_API_SECRET_TOKEN
```
mtaylorgds, line 2424: Is there a danger in having both `publisher` and `publisher-on-pg` able to talk to the various integrations? Is this necessary (especially in prod)?

charts/app-config/values-production.yaml
Tetrino: This is correct. The branch is only there for replication/migration and I didn't want the job to clash. When migration is complete the original job will continue as this branch gets retired.

samsimpson1 (approved):

PR #3395, opened 2025-08-05
Make `activeDeadlineSeconds` configurable for db migration job

ChrisBAshton:
## Why

The strong_migrations gem recommends setting a high statement_timeout to allow long-running migrations (e.g. adding indexes concurrently). In Whitehall, [we’ve configured](https://github.com/alphagov/whitehall/pull/10298):

```
StrongMigrations.statement_timeout = 1.hour
StrongMigrations.lock_timeout = 10.seconds
```

However, the current Kubernetes Job definition for database migrations has a hardcoded `activeDeadlineSeconds: 900` (15 minutes), which means the job may be terminated before the migration completes — even if MySQL is still executing the long-running statement. This risks cutting off in-progress migrations and leaving the database in an inconsistent state.

## What

This commit:

- Adds a new configurable value: `dbMigrationActiveDeadlineSeconds` (default: 900)
- Updates dbmigration-job.yaml to use the value from `.Values.dbMigrationActiveDeadlineSeconds`
- Allows individual applications (e.g. whitehall-admin) to override this timeout via environment-specific values files (e.g. values-integration.yaml)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -4108,6 +4108,7 @@ govukApplications:
           task: "taxonomy:rebuild_cache"
           schedule: "*/10 7-20 * * *"
       dbMigrationEnabled: true
+      dbMigrationActiveDeadlineSeconds: 3700 # 1 hour + 10 min buffer to match StrongMigrations.statement_timeout
       workers:
         enabled: true
       workerResources:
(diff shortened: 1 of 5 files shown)


ryanb-gds (approved):
Seems like a necessary evil to me. Let's see what the platform folks think

PR #3397, opened 2025-08-07
Test disabling fact-check processing

mtaylorgds:
Enable maintenance mode for the `publisher-on-pg` app in integration and also enable the `mail-fetcher` cronjob so that we can test that fact check email processing does not occur when maintenance mode is enabled.


AgaDufrat:
Do you still need to do this? Would you mind converting the PR to Draft until it's ready for review/merge?

mtaylorgds:
We didn't need this in the end; closing PR.

PR #3409, opened 2025-08-12
GOV.UK Chat: switch answer strategy to use OpenAI

jackbot:
We're having a pen test performed on GOV.UK Chat and one of the
requirements is that we need to test the OpenAI side of things.

This changes the `ANSWER_STRATEGY` env var, which will make GOV.UK Chat
use OpenAI for all its LLM calls.

We'll remove this (and so switch back to the default of using Anthropic)
after the pen test is complete.


davidgisbey (approved):
lgtm

PR #3416, opened 2025-08-13
feat: 🎸 add titles and descriptions to argo workflows

jaskaransarkaria:
- add title and description to:
  - deploy image
  - update-image-tag
  - post-sync

This change will surface key vars making it easier to search and indentify workflow runs as an application dev.

image

Changes:

- add title and descriptions to the workflows with the relevant vars
- pass the application var from the `post-sync` workflow to the `deploy-image` workflow where it can then be passed on to `update-image-tag`

Testing:

- I tested the logic of adding titles and descriptions on a test workflow that I created in the ui

closes: https://github.com/alphagov/govuk-infrastructure/issues/1967

### DIFF
--- a/charts/argo-services/templates/workflows/post-sync/workflow.yaml
+++ b/charts/argo-services/templates/workflows/post-sync/workflow.yaml
@@ -81,7 +81,7 @@ spec:
           curl -s "${WEBHOOK_URL}/update-image-tag" \
             -H "Authorization: Bearer ${WEBHOOK_TOKEN}" \
             --json '{
-              "application": "{{"inputs.parameters.application"}}"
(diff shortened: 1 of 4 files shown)


charts/argo-services/templates/workflows/post-sync/workflow.yaml
```diff
@@ -75,6 +81,7 @@ spec:
           curl -s "${WEBHOOK_URL}/update-image-tag" \
             -H "Authorization: Bearer ${WEBHOOK_TOKEN}" \
             --json '{
+              "application": "{{"inputs.parameters.application"}}"
```
samsimpson1, line 84: Looks like there are missing opening brackets here

samsimpson1 (approved):

PR #3440, opened 2025-08-20
[PP-7316] Add anonymous_user_id_secret to external secrets

richardTowers:
This secret has been added manually in integration, staging and production. [More context for the secret here](https://github.com/alphagov/signon/blob/main/docs/arch/adr-001-pass-anonymised-user-ids-to-publishing-apps-and-analytics.md).

The secret needs to be shared between different applications (i.e. whitehall should get the same secret value as mainstream etc.). Consequently, I haven't put it in any of the app-specific directories, instead choosing to put it at the top level, alongside basic-auth.yaml which is also shared.

I'm interested in a discussion as to whether this file / directory is the right place for it - it's only going to be used in publishing apps (whitehall, travel advice, publisher, content-tagger, signon etc.), but that's not _all_ apps.

### DIFF
--- a/charts/external-secrets/templates/anonymous-user-id.yaml
+++ b/charts/external-secrets/templates/anonymous-user-id.yaml
@@ -1,0 +1,20 @@
+apiVersion: external-secrets.io/v1beta1
+kind: ExternalSecret
+metadata:
+  name: anonymous-user-id
+  labels:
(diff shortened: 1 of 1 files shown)


richardTowers:
Thanks! Let's go with this for now then.

PR #3448, opened 2025-08-21
Add emergency banner redis url access to smartanswers in integration

KludgeKML:
- Also update k8s startup probe to use the ready endpoint


AgaDufrat (approved):
_It would be nice if there was a bit more context in the PR so we don't have to click through and read the Trello card._

PR #3450, opened 2025-08-21
Add startup probe failure alertmanager rules

jfharden:
Add alerting to #govuk-deploy-alerts which informs us when the startProbes for an app are failing for an extended time


jaskaransarkaria (approved):
✨ LGTM

PR #3457, opened 2025-08-21
Purposefully try and deploy a bad emergency banner redis url for draft-static

jfharden:
Follow on to test again after https://github.com/alphagov/govuk-helm-charts/pull/3454 and https://github.com/alphagov/govuk-helm-charts/pull/3452 both didn't trigger the alert


AP-Hunt (approved):
🎠

PR #3463, opened 2025-08-26
Update the requested memory for frontend in all envs

jfharden:
Since we enabled ruby JIT compilation frontend is using more memory and keeps getting it's app container OOMKilled. It has plenty of headroom in the limit, but the requested is a touch too low. Update this in prod and staging to 1.5GiB, and in integration to 500MiB


kentsanggds (approved):
LGTM

PR #3471, opened 2025-08-27
Add additional sidekiq processes to GOV.UK Chat

davidgisbey:
We're making changes to GOV.UK Chat to ensure user questions are prioritised over non-time-sensitive tasks.

We're introducing a bunch of jobs that will run after an answer is composed for analysis. These need to run on a separate Sidekiq process to ensure we don't block users from getting their questions answered.

We can remove the original worker config because the `default-worker` implements a :default queue which will be used in the meantime prior to https://github.com/alphagov/govuk-chat/pull/499 being merged

We're also scaling up the process which generates answers since that does the bulk of the work and is time critical on production ahead of the experiment we're running with the app.

The plan is:

1. merge the PR on GOV.UK Chat that adds the sidekiq config ✅ 
2. merge this PR
3. merge the PR on GOV.UK Docker that adds the new Sidekiq processes
4. merge the PR on GOV.UK Chat that uses the new Sidekiq processes
6. remove unused config on GOV.UK Chat

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1566,8 +1566,10 @@ govukApplications:
       workers:
         enabled: true
         types:
-          - command: ["sidekiq", "-C", "config/sidekiq.yml"]
-            name: worker
+          - command: ["sidekiq", "-C", "config/sidekiq_answer.yml"]
+            name: answer-worker
+          - command: ["sidekiq", "-C", "config/sidekiq_default.yml"]
+            name: default-worker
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Looks good to me, but will need a Platform Engineering review.

I also wonder - if we already have the new sidekiq config files in the app when this is merged (which I imagine we want to avoid errors) then I'd have thought this PR could also remove the process for config/sidekiq.yml and not need a second PR?

MuriloDalRi (approved):

PR #3501, opened 2025-09-05
Add HPA config for Chat Staging

ianhowell-gds:
## What

Update pod specs and apply horizontal pod autoscaling for Chat service in Staging

## Why

So we can carry out load testing emulating the config we want to deploy in Production


rtrinque (approved):
Looks good

PR #3507, opened 2025-09-09
Add graph for Bedrock invocation count to Chat dashboard

ianhowell-gds:
## What

Add a graph to count the number of invocations per minute of the Bedrock service

## Why

There is a rate limit of 200 requests per minute, so we would like to monitor the usage


rtrinque (approved):
Looks good

PR #3515, opened 2025-09-11
Remove publisher-on-pg

mtaylorgds:
With Mainstream Publisher now migrated to use a PostgreSQL—instead of MongoDB—database, there is no longer a need for the publisher-on-pg application.


AgaDufrat (approved):
LGTM

AP-Hunt (approved):

PR #3548, opened 2025-09-18
Split token graphs by ModelId (Titan & Sonnet)

ianhowell-gds:
## What

Separate out the Tokens Used graphs so that they show the different tokens by ModelId

## Why

There are different limits for the two types of token (Titan & Sonnet) we are using, so it makes sense to separate them on the graphs in order to have full visualisation of their usage


rtrinque (approved):
Looks good

PR #3499, opened 2025-09-04
Add alerting for search-api-v2 search quality degradation

emmalowe:
Search quality alerting thresholds have been set as:
- Binary recall top 3 < 90% (warning)
- Binary recall top 3 < 80% (critical)
- Clickstream NDCG top 10 < 85% (warning)
- Clickstream NDCG top 10 < 75% (critical)

Metrics are checked every 15mins, and an alert is raised immediately if metrics are found to be below the relevant threshold.

Edits to https://docs.publishing.service.gov.uk/manual/search-alerts-and-monitoring.html#alertmanager to reflect these changes coming soon!

Jira ticket for this PR: https://gov-uk.atlassian.net/browse/SCH-1423

See Jira ticket for relevant decisions taken for this PR, and further justification here: https://docs.google.com/document/d/1uXpBrc8eE2D-hyTv0hygPnkOMqahUpaARBOtZ9NmWyg/edit?tab=t.0#heading=h.lsh5cbl1kvlj

PR #3559, opened 2025-09-23
Suppress warning notifications when critical alerts are firing

emmalowe:
The search team have set up two separate alert thresholds for metrics related to search quality evaluations results, to distinguish between a "warning" level drop in quality, and a "critical" level drop in quality. When the critical alert is firing, it would be overly noisy and possibly confusing to also notify for the warning alert so this change suppresses the warning notifications when the critical alert is firing.

See Alert Manager docs: https://prometheus.io/docs/alerting/latest/configuration/#inhibition-related-settings

Alerts that will be affected by this change are being added here: https://github.com/alphagov/govuk-helm-charts/pull/3499

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1465

⚠️ Please don't assume I know what I'm doing with this change - it's a bit new to me! 😅

### DIFF
--- a/charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
+++ b/charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
@@ -180,6 +180,16 @@ spec:
       {{- include "slack.template" . | nindent 6 }}
       text: |-
         {{- include "slack.text" . | nindent 8 }}
+  inhibitRules:
+  - sourceMatch:
+      severity: 'critical'
+      destination: 'slack-search-team'
+    targetMatch:
+      severity: 'warning'
+      destination: 'slack-search-team'
+    equal:
+      - environment
+      - alertname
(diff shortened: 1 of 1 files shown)


AP-Hunt:
I think we'll want to discuss this as a platform engineering team, because this is behaviour change that will affect everybody

charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
```diff
@@ -180,6 +180,16 @@ spec:
       {{- include "slack.template" . | nindent 6 }}
       text: |-
         {{- include "slack.text" . | nindent 8 }}
+  inhibitRules:
+  - sourceMatch:
```
hannako, line 184: In the docs linked to in the description, the syntax looks different. ie source_match etc. Do you know why it's camel case here? Presume camel case is correct, as that matches the rest of the keys in this file, just not sure I understand why!

PR #3571, opened 2025-09-25
Update inhibit rules definitions for alert manager config

emmalowe:
Inhibit rules is not working as it should (warning notifications are not being suppressed when they should be). One reason could be that the definition is written in a style that has been deprecated, so this updates the definition to the new format.

See Alertmanager docs: https://prometheus.io/docs/alerting/latest/configuration/#inhibit_rule

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1465

### DIFF
--- a/charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
+++ b/charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
@@ -181,19 +181,13 @@ spec:
       text: |-
         {{- include "slack.text" . | nindent 8 }}
   inhibitRules:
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):
I've read the docs, and I think agree that's what it should look like AFAICT.

emmalowe:
This has already been addressed in this PR: https://github.com/alphagov/govuk-helm-charts/pull/3568

We're not using native alertmanager, so the relevant docs for this are actually here:  
https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1alpha1.InhibitRule

PR #3573, opened 2025-09-25
Add last_over_time recording rule to search quality alerts

emmalowe:
Our alerts are flapping off and on because of data dropouts in Prometheus. Prometheus treats a lack of data as “everything is ok” by default, so when we’re having an incident and then there’s a data dropout, the alert stops firing (and is marked as resolved). When we start getting data again and it’s below the threshold, the alert becomes active again.
    
Adding a last_over_time recording rule allows us to smooth over data gaps by looking at the last metric received over a specified period. In this case we've chosen 26 hours because data drops outs that are roughly 24 hours in length appear to be quite common for the clickstream query set.

See Prometheus docs on \<aggregation\>_over_time() recording rules:
https://prometheus.io/docs/prometheus/latest/querying/functions/#aggregation_over_time

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1477

Example of data drop outs in Prometheus: 
Screenshot 2025-09-29 at 17 49 52

### DIFF
--- a/charts/monitoring-config/rules/search_api_v2.yaml
+++ b/charts/monitoring-config/rules/search_api_v2.yaml
@@ -146,11 +146,18 @@ groups:
   - name: SearchApiV2QualityMetrics
     interval: 15m
     rules:
+      - record: search_api_v2:quality_metrics:last_binary_recall
+        expr: last_over_time(search_api_v2_evaluation_monitoring_recall{top="3", dataset="binary", month="last_month"}[26h])
+      - record: search_api_v2:quality_metrics:last_clickstream_ndcg
+        expr: last_over_time(search_api_v2_evaluation_monitoring_ndcg{dataset="clickstream", top="10", month="last_month"}[26h])
       - alert: SearchQualityDegradedBinaryRecall
-        expr: search_api_v2_evaluation_monitoring_recall{top="3", dataset="binary", month="last_month"} < 0.90
+        expr: search_api_v2:quality_metrics:last_binary_recall < 0.90
         labels:
           severity: warning
           destination: slack-search-team
+          top: 3
+          dataset: binary
+          month: last_month
         annotations:
           summary: "Search API v2: Degraded search result quality (binary recall top 3 has dropped)"
           grafana_path: >-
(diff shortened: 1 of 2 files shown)


charts/monitoring-config/rules/search_api_v2_tests.yaml
```diff
@@ -300,15 +300,12 @@ tests:
...
               description: "24 hour rolling success rate for search requests has dropped below 99.99% for more than 24 hours."
 
-    # Tests for SearchQualityDegradedBinaryRecall
+    # Tests for SearchQualityDegradedBinaryRecall (data below threshold)
   - interval: 15m
     input_series:
       - series: 'search_api_v2_evaluation_monitoring_recall{top="3", dataset="binary", month="last_month"}'
         values: '0.95x2 0.89 0.79'
     alert_rule_test:
-      - eval_time: 30m
```
emmalowe, line 309: Since I was adding even more new tests, I took this one out because it didn't seem really necessary. Same for clickstream below

dj-maisy (approved):

hannako (commented):
Excellent sleuthing! Given none of us are really experts in this field, it would be really helpful if you could break this PR up a bit and add links to the docs in the commit. 

I can see where you've added a `last_over_time` recording rule, but then you've also added another one that uses the `absent()` syntax. Would it makes sense to split those out into separate commits, and then you could link to the documentation for each change.

emmalowe:
It took me a few goes, but I've split the commit into two as you suggested @hannako !

charts/monitoring-config/rules/search_api_v2.yaml
```diff
@@ -146,11 +146,18 @@ groups:
...
+        expr: last_over_time(search_api_v2_evaluation_monitoring_recall{top="3", dataset="binary", month="last_month"}[26h])
+      - record: search_api_v2:quality_metrics:last_clickstream_ndcg
+        expr: last_over_time(search_api_v2_evaluation_monitoring_ndcg{dataset="clickstream", top="10", month="last_month"}[26h])
       - alert: SearchQualityDegradedBinaryRecall
-        expr: search_api_v2_evaluation_monitoring_recall{top="3", dataset="binary", month="last_month"} < 0.90
+        expr: search_api_v2:quality_metrics:last_binary_recall < 0.90
         labels:
           severity: warning
           destination: slack-search-team
+          top: 3
```
emmalowe, line 158: I'm having to add these labels back in manually because (as far as I understand it) by using a recording rule, it means the labels from the raw metric don't flow through automatically anymore.

PR #3581, opened 2025-09-30
Update startupProbeFailure alert to only alert on increases over 10 minutes

jfharden:
We essentially alert forever on any container which fails its startup probe 30 times, but then actually launches. This is because it's probe failure metrics lives on if the container lives on, and it's always then over 30. So change this to only alert on increases over 10 minutes.

It also meant slightly adjusting the existing tests input series since `30x10` means `30+0x10` which expands to `30 30 30 30 30 30 30 30 30 30`, so making it `0+10*10` makes it `0 10 20 30 40 50 60 70 80 90 100`

I did a tiny bit of cleaning up too (removing some comments which are just duplicating the name of the rule

### DIFF
--- a/charts/monitoring-config/rules/startup_probe_failure.yaml
+++ b/charts/monitoring-config/rules/startup_probe_failure.yaml
@@ -4,10 +4,12 @@ groups:
       - alert: StartupProbeFailure
         expr: >-
           label_replace(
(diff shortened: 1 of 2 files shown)


AP-Hunt (approved):
I like this

PR #3593, opened 2025-10-07
Add pod count graphs to Chat dashboard

ianhowell-gds:
## What

Add graphs to the Chat Grafana dashboard showing how many pods are running for frontend and worker jobs.

## Why

This is to help monitor how well the service is coping with workload and if the autoscaling is working as we want and expect.


rtrinque (approved):
Looks good

PR #3619, opened 2025-10-17
[SCH-1553] Add search api v2 beta features to staging and production

leenagupte:
Related to: https://github.com/alphagov/govuk-helm-charts/pull/3625
[Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/SCH/boards/994?selectedIssue=SCH-1492)

# What's changed?

Adds environment variables and cron tasks for [search-api-beta-features](https://github.com/alphagov/search-api-v2-beta-features) to staging and production.

# Why?

A new [search-api-beta-features] has been created to use features from in the `google-cloud-discovery_engine-v1beta` client gem. These features can't be added to the existing [search-api-v2](https://github.com/alphagov/search-api-v2) because the beta client gem is not compatible with the `google-cloud-discovery_engine` main client gem.

The app will be used to run the evaluations for GOV.UK site search that are currently run by [search-api-v2].

[search-api-beta-features]: https://github.com/alphagov/search-api-v2-beta-features
[search-api-v2]: https://github.com/alphagov/search-api-v2

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -3101,21 +3101,6 @@ govukApplications:
       redis:
         enabled: true
       cronTasks:
-        - name: report-clickstream-metrics
-          task: "quality:report_quality_metrics[clickstream]"
-          schedule: "0 6 * * *" # 06:00 GMT daily
-        - name: report-explicit-metrics
-          task: "quality:report_quality_metrics[explicit]"
-          schedule: "0 7 * * *" # 07:00 GMT daily
-        - name: report-binary-metrics
-          task: "quality:report_quality_metrics[binary]"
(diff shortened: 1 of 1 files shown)


charts/app-config/values-integration.yaml
```diff
@@ -3179,6 +3164,51 @@ govukApplications:
               name: search-api-v2-google-cloud-discovery-engine-configuration
               key: DISCOVERY_ENGINE_DEFAULT_LOCATION_NAME
 
+  - name: search-api-v2-beta-features
+    helmValues:
```
emmalowe: Do we need any additional stuff from `search_api_v2` here? I guess I'm surprised to not see anything about `appResources` or similar, but then again I don't know a lot about how this works :)

emmalowe (approved):

PR #3624, opened 2025-10-20
Add crontask for govuk-synthetic-test-app-runner and allow sh command

kentsanggds:
This crontask will create a cronjob which will run on the cluster in the apps namespace.

At the moment I'm deploying the cronjob via the crontask in order to allow the cronjob to be deployed with different ENVIRONMENT_NAME per environment. 

I'm also just pulling the image from GHCR.

So a few questions: 

- should the cronjob be in the `apps` namespace as it is testing the release pipeline?
- should it be deployed like other govuk apps?
- should we be pulling the image from GHCR? or use ECR?

https://github.com/alphagov/govuk-infrastructure/issues/2736

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -4010,6 +4010,9 @@ govukApplications:
 
   - name: govuk-synthetic-test-app
     helmValues:
+      appImage:
+        repository: ghcr.io/alphagov/govuk/govuk-synthetic-test-app-runner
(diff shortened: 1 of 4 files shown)


kentsanggds:
Will pivot away from this approach because we only need a cronjob and not a deployment kind for the Ginkgo tests to run. 
After a team chat following standup we decided that we should leave it running in the `apps` namespace but should pull the image from `ECR`

PR #3625, opened 2025-10-20
[SCH-1492] Add search-api-v2-beta-features app to integration

leenagupte:
Related to: https://github.com/alphagov/govuk-helm-charts/pull/3619
[Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/SCH/boards/994?selectedIssue=SCH-1492)

# What's changed?

Adds environment variables and cron tasks for [search-api-beta-features](https://github.com/alphagov/search-api-v2-beta-features) to Integration.

# Why?

A new [search-api-beta-features] has been created to use features from in the `google-cloud-discovery_engine-v1beta` client gem. These features can't be added to the existing [search-api-v2](https://github.com/alphagov/search-api-v2) because the beta client gem is not compatible with the `google-cloud-discovery_engine` main client gem.

The app will be used to run the evaluations for GOV.UK site search that are currently run by [search-api-v2].

[search-api-beta-features]: https://github.com/alphagov/search-api-v2-beta-features
[search-api-v2]: https://github.com/alphagov/search-api-v2

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -3110,21 +3110,6 @@ govukApplications:
       redis:
         enabled: true
       cronTasks:
-        - name: report-clickstream-metrics
-          task: "quality:report_quality_metrics[clickstream]"
-          schedule: "0 6 * * *" # 06:00 GMT daily
-        - name: report-explicit-metrics
-          task: "quality:report_quality_metrics[explicit]"
-          schedule: "0 7 * * *" # 07:00 GMT daily
-        - name: report-binary-metrics
-          task: "quality:report_quality_metrics[binary]"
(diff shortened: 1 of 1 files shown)


nimalank7:
This is fine (assuming secrets are already in Secret Manager). Note that the default `appResource` values are:

```
appResources:
  limits:
    cpu: 1
    memory: 1500Mi
  requests:
    cpu: 500m
    memory: 800Mi
```

There are more default values located in https://github.com/alphagov/govuk-helm-charts/blob/main/charts/generic-govuk-app/values.yaml so it's worth checking if you're happy with those.

At the moment ingress isn't enabled so you won't be able to access the app from outside the cluster if that's the desired behaviour?

PR #3642, opened 2025-10-23
Use `search-api-v2` secret key base for `search-api-v2-beta-features`

nimalank7:
Description:
- This is just testing out beta features for `search-api-v2` and is only used to run Rake tasks so I think it's fine to share the secret
- https://gov-uk.atlassian.net/jira/software/c/projects/SCH/boards/994?selectedIssue=SCH-1492


AgaDufrat (approved):
Makes sense to share it in this case although it would be good to mention in the commit message why those apps need to share sessions.

PR #3768, opened 2025-11-18
Increase memory limit for link-checker-api workers to 4GB

samsimpson1:
2GB isn't enough. If this doesn't work, I'll let the team know and leave it


nicholsj (approved):
LGTM thanks

PR #3786, opened 2025-11-24
Add app management scripts for maintenance

samsimpson1:
This includes a set of python scripts that manipulate the values file for a given environment.

Each script takes a list of apps to perform a given operation on, and the path to a values file.

Each operation either enables or disables app deployments, worker deployments or cron jobs.

This is done via string manipulation of the yaml file, instead of parsing it and manipulating it as a data structure. This is to preserve formatting in the file, as well as various things like comments and yaml anchors.

These scripts are for use during app maintenance. They are not intended to be used long-term, and as such may be flaky or difficult to maintain.

https://github.com/alphagov/govuk-infrastructure/issues/3222


AP-Hunt:
This is really hard to review as it stands. There's a _lot_ of code, and I don't have as much context as I'd like. Can you provide any test cases I can verify?

PR #3817, opened 2025-11-27
Add JWT secret to Content Block Manager (Integration)

pezholio:
This adds a new secret to generate JWT tokens for non-signon users to access our new preview endpoint for factchecking purposes.


dj-maisy (approved):
I'm assuming `authenticating-proxy-jwt-auth-secret` was being used as a placeholder/tester. 👍

PR #3837, opened 2025-12-04
Fact Check Manager integration

AlexRyanShep:
This PR sets up images tags and values for Fact Check Manager to get it hosted on Integration for development. As we're only targeting Integration for now, `promote_deploys` is disabled.


jfharden (commented):
Approving with a caveat, I suspect early in your app development it might be a particularly annoying one too

PR #3848, opened 2025-12-06
Disable publishing apps in production

samsimpson1:


charts/app-config/values-production.yaml
```diff
@@ -1808,6 +1819,7 @@ govukApplications:
         - name: PUMA_TIMEOUT
           value: "300"
       cronTasks: []
+      appEnabled: false
   - name: licensify
```
AgaDufrat: Could you double check `licensify`? I can't see it being disabled with `appEnabled: false` but I seem to remember it had a different key (it may have changed)

AgaDufrat (approved):

PR #3896, opened 2025-12-17
[SCH - 1406] - Update GOVUK Search dashboard

hannako:
Add Vertex metrics, and reorder rows.

See https://grafana.eks.production.govuk.digital/d/jep6t8d/gov-uk-search-with-vertex-metrics-2?orgId=1&from=now-12h&to=now&timezone=browser

Screenshot 2025-12-22 at 15 48 57

Jifa ticket: https://gov-uk.atlassian.net/browse/SCH-1406

### DIFF
--- a/charts/monitoring-config/dashboards/govuk-search.json
+++ b/charts/monitoring-config/dashboards/govuk-search.json
@@ -128,8 +128,8 @@
         ]
       },
       "gridPos": {
-        "h": 8,
-        "w": 14,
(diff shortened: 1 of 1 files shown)


leenagupte (approved):
The changes look good. 🎉 

I think perhaps the reorder commit could be fixed, but that's not a blocker.

I think a later change could be to line up the response times and the latency so they sit on top of each other.

PR #3908, opened 2025-12-23
[PP-7503] Enable graphql for transactions

callumknights:
Add an environment variable that will direct 1% of traffic for transactions to graphql.

This will be increase later on when we have monitored the effect of this change.


brucebolt (approved):
Change is fine with me.

Could you add the JIRA work item number to the PR title, so it links back (makes it easier to find the PRs from the work item that way).

PR #3909, opened 2025-12-23
[SCH-1636] Enable continuous deployment for search-api

stringiest:
Enable continuous deployment for search-api.

I confirm that the application meets the requirements set out in [RFC-128](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-128-continuous-deployment.md):
- [x] only people with production access can deploy to the live site
- [x] code coverage for the application exceeds 95%
- [x] application has [contract tests](https://docs.publishing.service.gov.uk/manual/pact-testing.html) where needed
- [x] application is healthy (e.g. has tests that check for connectivity to any read/write dependencies

### DIFF
--- a/charts/app-config/image-tags/integration/account-api
+++ b/charts/app-config/image-tags/integration/account-api
@@ -1,3 +1,3 @@
-image_tag: v630
+image_tag: v633
 promote_deployment: true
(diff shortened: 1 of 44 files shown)


hannako (approved):
🥳 WOOHOO!

PR #3910, opened 2025-12-24
[PP-7484] Increase some GraphQL rates to 50%

brucebolt:
These pages have been running at 10% traffic rendered by GraphQL for the last few days with an acceptable response time.

Therefore increasing the traffic to 50%.


callumknights (approved):

brucebolt:
We're holding off merging this until Monday, as we don't want to leave the additional traffic unattended for so many days.

PR #3916, opened 2026-01-05
Alert on Crawler Errors Threshold hitting 10%

dj-maisy:
## What?
This adds an alert to fire when the percentage of Crawl Errors on govuk-mirror exceeds 10% in a 1 hour window. This will imply that something is very wrong with the Crawler that is preventing it from doing its job, i.e. a rate limit, connection error, timeout or some other kind of outage.

It will currently notify Platform Engineering via Slack in the event of a failure, rather than Raise to PagerDuty. If we think the threshold is too high, we can adjust it. It is normal to get a low rate of errors (currently about 2,000 per run) due to pages being `HTTP 410 "Gone"` which is acceptable to us. The metrics exposed do not currently allow us to differentiate between 410 and 5XX, however.

### DIFF
--- a/charts/monitoring-config/rules/mirrors_tests.yaml
+++ b/charts/monitoring-config/rules/mirrors_tests.yaml
@@ -30,3 +30,37 @@ tests:
             exp_annotations:
               summary: Mirror freshness has exceeded two days
               description: The mirror hasn't been updated in more than two days
+
+  - name: MirrorCrawlErrorThresholdExceeded - Alert when error rate exceeds 10%
+    interval: 1m
+    input_series:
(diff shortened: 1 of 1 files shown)


charts/monitoring-config/rules/mirrors.yaml
```diff
@@ -14,3 +14,16 @@ groups:
           summary: Mirror freshness has exceeded two days
           description: >-
             The mirror hasn't been updated in more than two days
+
+      - alert: MirrorCrawlErrorThresholdExceeded
```
AP-Hunt: Let's add a test for this :D

PR #3918, opened 2026-01-06
Set min: 0 on Average GraphQL Response Time gauge

richardTowers:
... Otherwise it calculates the minimum from the data, and the gauge ends up being really confusing.

| Before | After |
|---|---|
| image | image |


AgaDufrat (approved):
✨

PR #3925, opened 2026-01-06
[PP-7515] Add Zendesk secret template for support API

callumknights:
Add external secrets template for new Support API zendesk secret

### DIFF
--- a/charts/external-secrets/templates/support-api/zendesk.yaml
+++ b/charts/external-secrets/templates/support-api/zendesk.yaml
@@ -1,0 +1,23 @@
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: support-api-zendesk
+  labels:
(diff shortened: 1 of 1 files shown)


charts/external-secrets/templates/support-api/zendesk.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: support-api-zendesk
+  labels:
+    {{- include "external-secrets.labels" . | nindent 4 }}
+  annotations:
+    kubernetes.io/description: >
+      Zendesk Oauth access token for Supoort API
```
brucebolt:
```suggestion
      Zendesk Oauth access token for Support API
```

brucebolt:
Could you also add the JIRA work item number, so the link gets created to this PR from there (for easier navigation between the two).

PR #3983, opened 2026-01-13
feat: 🎸 pass irsa to renovate pod

jaskaransarkaria:
This is the 2/3 prs to enable renovate to watch eks addons for updates:

1. ~~[create the new renovate irsa](https://github.com/alphagov/govuk-infrastructure/pull/3471)~~
1. pass the new service account with correct aws permissions to the renovate cron job
1. [configure the `renovate.json` to check for any updated eks addons and add gha to ensure that the hardcoded k8s version in `renovate.json` is up to date](https://github.com/alphagov/govuk-infrastructure/pull/3473)

related:  https://github.com/alphagov/govuk-infrastructure/issues/3423


samsimpson1 (approved):

charts/app-config/values-production.yaml
```diff
@@ -2633,6 +2640,15 @@ govukApplications:
         enabled: true
         create: false
         name: release-assumer
+  - name: renovate
```
jfharden: Probably want to update https://github.com/alphagov/govuk-helm-charts/blob/7e131419ee57e43b894dd5f4ee1184077e70a7a5/charts/app-config/templates/renovate.yaml instead of this

PR #4007, opened 2026-01-19
Add places-manager ExternalSecret only

jfharden:
The previous attempt at this (https://github.com/alphagov/govuk-helm-charts/pull/4003) failed and caused the db migration jobs to try and create a new database within the postgres instance named places-manager_production.

I realised that the DATABASE_URL has a hardcoded database name on the end, and in creating the previous external secret I had changed it to places-manager_production, so this PR creates only the secret and sets it to imminence_production with a comment for future readers to understand why it's called that.

It's on the list of @KludgeKML's team to rename the database, but it's quite the pain, so for now lets stick with it being called imminence.

### DIFF
--- a/charts/external-secrets/templates/places-manager/postgres.yaml
+++ b/charts/external-secrets/templates/places-manager/postgres.yaml
@@ -1,0 +1,28 @@
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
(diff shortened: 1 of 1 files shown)


AP-Hunt (approved):
Confirm the `postgis-conn-string.tpl` file exists. I wondered if that was a typo at first.

PR #4059, opened 2026-02-05
Add Database Backups & Syncs grafana dashboard

jfharden:
Add a grafana dashboard to visualise the current state of the database backups and syncs, as well as show some history of filesizes, durations, and state changes

You can see this currently deployed in staging (I picked staging since it is the environment that does all of the operations (restore, transform, backup), and it does them every weekday meaning the stats I added yesterday have been populated overnight: https://grafana.eks.staging.govuk.digital/d/jfc4fnp/database-backups-and-syncs

There will be a follow on PR to add alerting for these, and then another follow on where I will add a display of current alerts to this dashboard.

A couple of screenshots:

The table at the top with summarised information about the latest backup
Screenshot 2026-02-05 at 09 27 01

An enlarged version (just by choosing view in the dashboard) version of the file sizes over time for shared-documentdb
Screenshot 2026-02-05 at 09 29 25

Durations over time showing publisher (which doesn't do a transform) and publishing-api (which does)
Screenshot 2026-02-05 at 09 28 17

State transition view for publishing-api

Screenshot 2026-02-05 at 09 30 10

### DIFF
--- a/charts/monitoring-config/dashboards/database-backups-and-syncs.json
+++ b/charts/monitoring-config/dashboards/database-backups-and-syncs.json
@@ -1,0 +1,987 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "Showing the state of the database backup and sync processes",
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 0,
+  "id": 0,
+  "links": [],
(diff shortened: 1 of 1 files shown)


charts/monitoring-config/dashboards/database-backups-and-syncs.json
```diff
@@ -0,0 +1,987 @@
...
+    ]
+  },
+  "time": {
+    "from": "now-2d",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "browser",
+  "title": "Database Backups & Syncs",
+  "uid": "jfc4fnp",
```
jaskaransarkaria: How was this uid generated? I've previously had issues when a dashboard skeleton was copy and pasted and then grafana couldn't load subsequent dashboards because of the clashing uid.

jfharden: It's generated by grafana. I made this 100% in the UI.

I'll delete the one I created by hand which will allow this to be created :)

PR #4067, opened 2026-02-10
Run the Mirror drift checker at midday on Mondays, Wednesdays and Fridays

AP-Hunt:
## What
We wanted to run in every 3 days at midday, but that's a hard (if not impossible) to express in Crontab syntax. Instead we settle for Mondays, Wednesdays, and Fridays.

## How to review
Put the expression I've used into something like https://crontab.guru and see that it comes out how I intended.


nicholsj (approved):
LGTM (I've double checked the cron config)

PR #4102, opened 2026-02-19
Reinstate the chat cron task to export to BigQuery

kevindew:
Chat is going to start having a few production users of the web application soon and we want their data to be exported hourly.


davidgisbey (approved):
lgtm!

PR #4162, opened 2026-03-23
Demo 2: revert mispelling

AlexRyanShep:
[As agreed earlier](https://gds.slack.com/archives/C013F737737/p1774268452987959), this is to support a training session, introducing a fault on purpose to give a live demo of interrogating the resulting logs. This PR reverts the first of the errors introduced


AlexRyanShep:
Note: force push was just a rebase with main so that this will correctly show the change to be merged/reverted

PR #4188, opened 2026-04-07
Revert "Change pagerduty drill to run on Tuesday"

KludgeKML:
Reverts alphagov/govuk-helm-charts#4177 (This is actually a re-re-revert due to the scheduling mistake I made when I initially merged, but it does what we want and puts the drill back on the Monday - we'll have to do this again in May).


nicholsj (approved):
LGTM thanks

PR #4201, opened 2026-04-09
[PNP-9923] Update startup probe endpoint for account-api on Integration

deborahchua:
We have a `/healthcheck/ready` endpoint in account-api that carries out some healthchecks,
but it has not been configured in this repo (the default is `/healthcheck/live`).

We will test this endpoint in Integration before proceeding to Staging and Production.

Refs:
- https://github.com/alphagov/account-api/blob/main/config/routes.rb#L3-L6

[Jira card](https://gov-uk.atlassian.net/browse/PNP-9923)


KludgeKML (approved):
LGTM

PR #4202, opened 2026-04-09
[PNP-9924] Update startup probe endpoint for email-alert-frontend on Integration

deborahchua:
We have a `healthcheck` endpoint in `email-alert-frontend` that carries out some healthchecks, but it has not been configured in this repo (the default is `/healthcheck/live`).

We will test this endpoint on Integration before proceeding to Staging and Production.

Refs:
- https://github.com/alphagov/email-alert-frontend/blob/main/config/routes.rb#L5-L7

[Jira card](https://gov-uk.atlassian.net/browse/PNP-9924)

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -1071,6 +1071,8 @@ govukApplications:
           memory: 320Mi
       redis:
         enabled: true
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM

PR #4206, opened 2026-04-10
[PNP-9925] Update startup probe endpoint for places-manager on Integration

deborahchua:
We have a `/healthcheck/ready` endpoint in `places-manager` that carries out some healthchecks, but it has not been configured in this repo (the default is `healthcheck/live`).

We will test this on Integration before proceeding to Staging and Production.

Refs:
- https://github.com/alphagov/places-manager/blob/main/config/routes.rb#L5-L7

[Jira card](https://gov-uk.atlassian.net/browse/PNP-9925)

Related PRs:
- https://github.com/alphagov/places-manager/pull/1577

### DIFF
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -2303,6 +2303,8 @@ govukApplications:
           memory: 200Mi
       redis:
         enabled: true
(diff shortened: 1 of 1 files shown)


sairamya93 (approved):
Looks good 👍

PR #4211, opened 2026-04-13
Add ability to run GitHub action after successful deploy

pezholio:
In Content Block Manager, we have a [suite of E2E tests](https://github.com/alphagov/content-modelling-e2e) that are more in depth than the GOV.UK E2E tests, and are only designed to be run in the integration environment. 

At the moment, these are only run on demand, and generally drift when changes are made. I'd like to run these after every successful integration deploy, so we can have confidence everything is still working as expected.

After some investigation, I found there wasn't an out of the box way to do this easily, and I had a bit of time to experiment how easy it would be to add this functionality.

This PR adds the ability to run a Github actions workflow for any application/environment during the post-sync workflow in Argo, which covers our use case, but also offers up the flexibility for any users to do the same.

I'm not super au-fait with Helm Charts / Kubernetes, but this seems like I have all the bases covered. Feedback appreciated! (even if it's just "no, this won't work" 😆 )

### DIFF
--- a/charts/app-config/templates/govuk-application.yaml
+++ b/charts/app-config/templates/govuk-application.yaml
@@ -19,6 +19,8 @@ metadata:
     imageTag: "{{ $imageTag }}"
     promoteDeployment: "{{ $promoteDeployment }}"
     automaticDeploysEnabled: "{{ $autoDeploysEnabled }}"
+    postSyncGithubActionRepo: "{{ .postSyncGithubAction.repo | default "" }}"
+    postSyncGithubActionWorkflow: "{{ .postSyncGithubAction.workflow }}"
     notifications.argoproj.io/subscribe.on-deployed.argo_events: ""
     notifications.argoproj.io/subscribe.deployment.grafana: "deployment|{{ .name }}"
     postSyncWorkflowEnabled: "{{ .postSyncWorkflowEnabled | default "true" }}"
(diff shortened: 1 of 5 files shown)


jfharden (approved):
So this looks ok to me (I'm also not super familiar with the nitty gritty of how the workflows hang together and work, but I asked the rest of the Platform Engineering team too), but with a few caveats:


* If this breaks the workflows it will break all deploys of all apps, so we should be ready to roll it back asap
* It's in our roadmap to move to concourse-ci to replace existing CI/CD solutions, but that's minimum months away (at least 6 I would assume). When we do it's possible we will lose this feature in the initial migration
* We don't know if the github token has permission to launch workflows or not, but if that breaks it only breaks for your app and you can easily turn it off
* You further tie your deploys into github availability, but specifically this time into github actions availability which generally has a lower availability than general git pulls, but this choice is with you and you can always turn it off if it gets too annoying :)

jfharden:
As an aside, an alternative might be to just run your e2e tests on a cron rather than tying into an argo workflow (the GHA workflow failing wont stop deployments anyway, so this just makes it a different kind of trigger that will run regularly, rather than triggered by deploys)

pezholio:
@jfharden Good idea - I'll give cron a go first - this might be overkill. Will revisit this if we think it's useful, but will close for now. Thanks for the prompt feedback and consideration 👍

PR #4226, opened 2026-04-15
Update fact-check-manager (integration) JWT_AUTH_SECRET to shared secret

young-rob:
We need to generate preview links to the draft origin with a JWT from fact-check-manager so this secret needs to be the same as `authenticating-proxy-jwt-auth-secret`. We can remove the existing secret defined specifically for fact-check-manager.

Once this is merged, we will delete the old secret from AWS secrets manager.


nimalank7 (approved):
LGTM!

PR #4230, opened 2026-04-17
Create regular snapshot of integration

stringiest:
The [govuk-docker/bin/replicate-elasticsearch.sh](https://github.com/alphagov/govuk-docker/blob/main/bin/replicate-elasticsearch.sh) script allows a user to replicate integration data locally. We do not currently run a regular job to create a snapshot of integration, so the latest available snapshots are from November 2025. 

Adding a cronTask to create a weekly snapshot of integration, for use locally by developers.

### DIFF
--- a/charts/search-index-env-sync/values.yaml
+++ b/charts/search-index-env-sync/values.yaml
@@ -80,6 +80,12 @@ cronjobs:
           value: govuk-staging
         - name: SUBDOMAIN
           value: elasticsearch6
+    green-es6-snapshot: # This would be too long as green-elasticsearch6-domain-snapshot
+      schedule: "37 5 * * 1"
+      args: [create_and_clean_up]
+      extraEnv:
+        - name: SUBDOMAIN
+          value: elasticsearch6
     chat-engine-cp-prod:
       schedule: "12 4 * * *"
       args: [restore_latest]


emmalowe:
Couple of questions here:

1. The order of snapshots and restore (in both staging and integration) seem back to front to me. Wouldn't you want to get the latest data before taking a new snapshot? Otherwise we're effectively introducing a lag in production data getting to integration.
2. Might it be confusing to take a weekly snapshot of integration, when the integration sync is happening daily? On the other hand, I see there might not be a huge value in having a daily snapshot. To go the other way, @ChrisBAshton raised the point that it could be confusing to have integration syncing daily for elasticsearch, while everything else in govuk seems to sync weekly for integration.

stringiest:
> 1. The order of snapshots and restore (in both staging and integration) seem back to front to me. Wouldn't you want to get the latest data before taking a new snapshot? Otherwise we're effectively introducing a lag in production data getting to integration.

Yes, I noticed that for staging & deliberately replicated it for integration. I think I was thinking that maybe we'd made a conscious decision to be a day behind on staging, and that there seemed a similar time gap available to mimic that for integration. (Though I'm not sure how I thought a day's delay would help on a weekly job.)

> 2. Might it be confusing to take a weekly snapshot of integration, when the integration sync is happening daily? On the other hand, I see there might not be a huge value in having a daily snapshot. To go the other way, @ChrisBAshton raised the point that it could be confusing to have integration syncing daily for elasticsearch, while everything else in govuk seems to sync weekly for integration.

Agreed. I couldn't quite bring myself to take a daily snapshot, as we really don't need to update our local data that often, but also didn't feel bold enough to take us down to a weekly sync of integration.

It feels like maybe we need to wait for @hannako to be back on Monday to talk about scheduling for this. (In that case we might end up with the create_and_clear job running on a Tuesday, at least initially, otherwise I'll have to wait yet another week before there's an up-to-date snapshot to pull from integration! )

Either that or this could go in as-is to give us an up-to-date snapshot for integration on Monday, and then we could discuss the correct scheduling for it all after that?

What do you think?

emmalowe:
> Either that or this could go in as-is to give us an up-to-date snapshot for integration on Monday, and then we could discuss the correct scheduling for it all after that?

Good point that we need a recent snapshot soon to unblock you, so I agree - I think this can go in as-is and then we can chat about refining it next week.

emmalowe (approved):

emmalowe (approved):

AP-Hunt (approved):

PR #4232, opened 2026-04-21
Account API doesn't use Notify, so doesn't need any config for it

KludgeKML:
To-do after merging:
- [ ] Delete the keys from Secrets Manager.


AgaDufrat (approved):
Checked that it's not used, so good to bin.

dj-maisy (approved):

PR #4247, opened 2026-04-23
Add more models to GOV.UK Chat technical dashboard

jackbot:
https://gdsgovukagents.atlassian.net/browse/CHAT-521

We're expanding the list of models we use on GOV.UK Chat to include:

* Claude Sonnet 4.5
* Claude Haiku 4.5

This means we'll want to monitor rate limits and token usage on this
dashboard, just like we do with the current models (Sonnet 4, Titan,
gpt-oss-120b).

This commit consolidates some graphs we have at the moment. Currently
each model has 3 graphs: percentage tokens used, number of tokens used
and invocation count. Adding 2 more models means another 6 graphs, to
bring us up to 15 graphs in total. It's becoming quite a mess.

So instead, this commit consolidates the percentage token used and
invocation count graphs, with each model having its own line on
the graph.

Number of tokens used graphs are tricky to consolidate, as models can have
different data attributes represented. For example with the Titan models
we only show `InputTokenCount`, whereas for GPT OSS we show
`InputTokenCount`, `OutputTokenCount` and `TotalTokenCount`, and for
Sonnet/Haiku we show `InputTokenCount`, `OutputTokenCount`,
`TotalTokenCount` and `CacheWriteInputTokenCount`. So showing all these
on the same graph would be confusing and just a total visual overload.

So we'll keep 5 separate graphs for the token count for the models.

You can see the new version in [this copy of the dashboard](https://grafana.eks.production.govuk.digital/d/jandz6k/gov-uk-chat-technical-copy?orgId=1&from=now-2d&to=now&timezone=browser&refresh=15m)

## Before

Screenshot 2026-04-23 at 16 03 25

## After

Screenshot 2026-04-23 at 16 03 17

### DIFF
--- a/charts/monitoring-config/dashboards/govuk-chat-technical.json
+++ b/charts/monitoring-config/dashboards/govuk-chat-technical.json
@@ -18,7 +18,6 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 1528,
   "links": [],
   "panels": [
     {
@@ -102,7 +101,7 @@
           "sort": "none"
         }
       },
-      "pluginVersion": "12.3.3",
+      "pluginVersion": "12.4.1",
       "targets": [
         {
           "editorMode": "code",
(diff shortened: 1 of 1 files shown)


ianhowell-gds (approved):
LGTM

PR #4248, opened 2026-04-23
Fix redirect location in router-probe-backend and add test for it

jfharden:
The redirect was adding the server hostname and scheme, so it was redirecting to `http://router-probe-backend/redirected` which is incorrect, so this PR:

* Turns off absolute redirects
* Fixes the actual location to be `/__probe__/redirected`
* Adds a test which explicitly checks the location header returns is as expected


jfharden:
Closes https://github.com/alphagov/govuk-infrastructure/issues/4020

samsimpson1 (approved):

PR #4255, opened 2026-04-28
Adds ALLOWED_GEOJSON_SLUGS value in integration

KludgeKML:
- This controls the /api/places endpoint, which will return 404 for all slugs unless they're listed in this variable.
- This means we'll be able to merge the geojson code without causing any data to be exposed until we're ready.


sairamya93 (approved):
LGTM!

PR #4277, opened 2026-05-19
Change PD drill to be on Tues

nicholsj:
Until we fix this properly we need to switch this to running on Tuesday next week to avoid the Bank Holiday.


KludgeKML (approved):
LGTM

PR #4291, opened 2026-06-02
[PP-7768] Add Slack alert for old Sidekiq locks

brucebolt:
The Content APIs team would like to be informed when the `sidekiq-unique-jobs` gem fails to release locks after Sidekiq jobs have completed.

Adding a Slack notification to inform our team when this occurs.

Depends on https://github.com/alphagov/publishing-api/pull/4077.

### DIFF
--- a/charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
+++ b/charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
@@ -124,9 +124,9 @@ spec:
       groupInterval: 30m
     - matchers:
         - name: destination
-          value: slack-graphql-notifications
+          value: slack-content-apis-team
           matchType: =
-      receiver: 'slack-graphql-notifications'
+      receiver: 'slack-content-apis-team'
       repeatInterval: 1h
       groupWait: 5m
       groupInterval: 30m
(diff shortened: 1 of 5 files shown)


charts/monitoring-config/rules/publishing_api.yaml
```diff
@@ -0,0 +1,16 @@
...
+        for: 20m
+        labels:
+          severity: warning
+          destination: slack-publishing-api-notifications
+        annotations:
+          summary: Sidekiq jobs locked for too long in Publishing API.
+          description: >-
+            Sidekiq jobs have been locked for more than one hour in Publishing API.
+          runbook_url: >-
+            https://docs.publishing.service.gov.uk/manual/alerts/publishing-api-old-locks-present.md
```
AgaDufrat, line 16: ❤️

charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
```diff
@@ -220,6 +220,12 @@ spec:
       {{- include "slack.template" . | nindent 6 }}
       text: |-
         {{- include "slack.text" . | nindent 8 }}
+  - name: 'slack-publishing-api-notifications'
```
AgaDufrat: This is fine but I wonder if we could rename and reuse `slack-graphql-notifications` given it has the same template configuration.

AgaDufrat (approved):

charts/kube-prometheus-stack-bootstrap/templates/alertmanagerconfig.yaml
brucebolt: Great idea.  I've renamed this and removed the new destination that I'd added.

brucebolt:
Also realised I missed adding a test for the alert.  I've added a couple of those in the second commit.

PR #4307, opened 2026-06-16
Sch 2143 add alert for number of autocomplete suggestions

emmalowe:
- Removes old AutocompleteDegradedAcute alert, which is no longer helpful now that we're handling GCP errors within search-api-v2, so these return a 200 status. (See https://github.com/alphagov/search-api-v2/pull/677)
- Adds new autocomplete alerting that checks the maximum number of autocomplete suggestions we return from a request to Google Cloud Discovery Engine, and raises when this drops to zero.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2143

### DIFF
--- a/charts/monitoring-config/rules/search_api_v2.yaml
+++ b/charts/monitoring-config/rules/search_api_v2.yaml
@@ -79,6 +79,13 @@ groups:
             /
             sum (search_api_v2:search_requests:rate1h) > 0
           ) or vector(1)
+      - record: search_api_v2:discovery_engine_autocomplete_suggestions_response:p100_increase1h
+        expr: |
+          histogram_quantile(1, sum by(le) (
+            increase(
+              search_api_v2_discovery_engine_autocomplete_suggestions_response_bucket[1h]
+            )
+          )) and sum(increase(search_api_v2_discovery_engine_autocomplete_suggestions_response_bucket{le="+Inf"}[1h])) > 500
       - alert: SearchDegradedMid
         expr: search_api_v2:search_successful_requests:ratio_rate1h < 0.999
         for: 2h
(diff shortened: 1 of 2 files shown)


charts/monitoring-config/rules/search_api_v2.yaml
```diff
@@ -127,6 +101,21 @@ groups:
           description: >-
             1 hour rolling success rate for search requests has dropped below 99.9% for more than 2
             hours.
+      - alert: AutocompleteDegradedMid
```
hannako, line 104: Do we need the "Mid"?

emmalowe: Maybe not. I added it to be in line with the other searchy alerts that look at metrics over a period of time. We seem to be using "Acute" for 5mins, "Mid" for 1hr and "Long" for 24hrs.

charts/monitoring-config/rules/search_api_v2.yaml
```diff
@@ -127,6 +101,21 @@ groups:
...
+        labels:
+          severity: warning
+          destination: slack-search-team
+        annotations:
+          summary: "Search API v2: Degraded autocomplete performance (mid)"
+          grafana_path: >-
+            d/govuk-search/gov-uk-search?orgId=1&from=now-24h&to=now&timezone=browser
+          runbook_url: >-
+            https://docs.publishing.service.gov.uk/manual/search-alerts-and-monitoring.html#alertmanager
+          description: >-
```
emmalowe, line 116: Is this any better @hannako ?

emmalowe:
I added some more details on the threshold to the commit message @hannako

hannako (approved):

PR #4308, opened 2026-06-16
Remove HighVertexP90Latency alert

emmalowe:
Now that we have a 2-second timeout for search requests [1], there is no need for this alert because it will never be triggered.

[1] https://github.com/alphagov/search-api-v2/blob/main/app/services/discovery_engine/clients.rb#L17


emmalowe:
Closing because we will likely want to tweak this rather than deleting it. 

I've left a note on https://gov-uk.atlassian.net/browse/SCH-2070

PR #4312, opened 2026-06-17
[PP-7776] Stop frontend apps from using Publishing API read replica in production

AgaDufrat:
We’re removing Publishing API read replica in staging and production to reduce
cost as we’re not actively using it at the moment.

Depends on:
- https://github.com/alphagov/frontend/pull/5614
- https://github.com/alphagov/collections/pull/4549
- https://github.com/alphagov/feedback/pull/2482


brucebolt (approved):

nimalank7 (approved):
LGTM!

PR #4313, opened 2026-06-17
[PP-7776] Remove publishing-api-read-replica deployment in production 

AgaDufrat:
We’re removing Publishing API read replica in staging and production to reduce
cost as we’re not actively using it at the moment.

Depends on (and includes commits):
- https://github.com/alphagov/govuk-helm-charts/pull/4312


brucebolt (approved):

nimalank7 (approved):
LGTM!

PR #4317, opened 2026-06-23
feat: add CKAN anonymisation script for integration restore (DGUK-688)

amar-gds-ndl:
I have now added ckan.sql which will masks PII after restoring staging data to integration it includes user: name → user_{id}, fullname → ***, email → user_{id}@example.com, apikey → *** package: author/maintainer names → ***, emails → masked@example.com and package_extra: masks values for 14 PII-bearing keys (emails, names, phones)

the transform step into the integration ckan-postgres cronjob in values.yaml.

### DIFF
--- a/charts/db-backup/scripts/ckan.sql
+++ b/charts/db-backup/scripts/ckan.sql
@@ -19,19 +19,17 @@
 -- Excludes active sysadmin accounts with a @dsit.gov.uk email — these are
 -- real admin accounts that must remain functional after restore.
 
-UPDATE public."user"
+UPDATE public.user
 SET
-  name     = 'user_' || id,
+  name = 'user_' || id,
   fullname = CASE
-               WHEN fullname IS NOT NULL THEN repeat('*', length(fullname))
-               ELSE NULL
-             END,
-  email    = 'user_' || id || '@example.com',
-  apikey   = CASE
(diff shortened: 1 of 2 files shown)


ashimali (approved):

charts/db-backup/scripts/ckan.sql
```diff
@@ -0,0 +1,96 @@
...
+
+-- ---------------------------------------------------------------------------
+-- user
+-- ---------------------------------------------------------------------------
+-- email:    user_{id}@example.com  — unique per user, traceable via UUID
+-- name:     user_{id}              — replaces username slug (often email-derived)
+-- fullname: repeat '*' to match original length
+-- apikey:   repeat '*' to match original length
+
+UPDATE public."user"
```
ashimali: As discussed we should exclude masking any of the sysadmin user accounts.

Leave users where users where sysadmin == true and state == ‘active’ and email ilike '%@dsit.gov.uk' unmodified.

amar-gds-ndl: As requested I have now skipped masking for users where sysadmin is true,
state is active, and email matches %@dsit.gov.uk

ashimali (approved):

jaskaransarkaria (approved):

PR #4321, opened 2026-06-26
feat: More complete anonymisation of user accounts. 

ashimali:
The backup produced after this anonymisation process is intended for wider use than just the datagovuk integration environment.

Therefore we decided to go further and effectively decommission user accounts.

This is instead of  deletion which would break relational integrity.

This PR removes the exclusion of sysadmin accounts, sets passwords to null (passwords are a nullable field) and lastly sets all users to status 'deleted'.

### DIFF
--- a/charts/db-backup/scripts/ckan.sql
+++ b/charts/db-backup/scripts/ckan.sql
@@ -4,9 +4,12 @@
 -- Runs as part of the db-backup transform step after restore from staging.
 --
 -- Tables covered:
---   user          (name, fullname, email, apikey)
+--   user          (name, fullname, email, apikey, password)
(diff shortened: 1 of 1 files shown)


ashimali:
Going to create another process on destination side to create dev usable backup rather than accomplish this here.

ashimali:
@kentsanggds  and @amar-gds-ndl  I think best to do this to be 100% honest. I'd rather have the friction of having to create test admin from cli on integration than carry over these account details from prod.

Let me know what you both think and if on-board, we can ask platform eng to approve

kentsanggds (approved):
LGTM

jfharden (approved):

PR #4329, opened 2026-07-01
Remove assisted digital feedback in integration

sairamya93:
- Assisted digital feedback experiment is no longer running in `feedback` app and hence the config to store the feedback information in google spreadsheets is removed.

[JIRA ticket](https://gov-uk.atlassian.net/browse/PNP-10059)


KludgeKML (approved):
LGTM

PR #4346, opened 2026-07-08
Add /search/opensearch.xml as static path served by nginx

KludgeKML:
- This was served by finder-frontend, but it hasn't changed, so is a good candidate for being served directly from nginx, with a cache of 30 minutes.


sihugh (approved):
🚀

AP-Hunt (approved):

PR #4348, opened 2026-07-13
Warning level alert if the elasticsearch cluster status unknown

hannako:
The Elasticsearch HealthAPI can return a status of UNKNOWN or UNAVAILABLE https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-health-report#operation-health-report-200-body-application-json-status

We need to add a link to up to date documentation to explain debugging steps in the event of this alert firing. Will be added in follow up work (See Jira [#SCH-2014](https://gov-uk.atlassian.net/browse/SCH-2014))

Jira https://gov-uk.atlassian.net/browse/SCH-2013

### DIFF
--- a/charts/monitoring-config/rules/search_api_alerts.yaml
+++ b/charts/monitoring-config/rules/search_api_alerts.yaml
@@ -2,6 +2,16 @@ groups:
   - name: SearchApiElasticsearchAlerts
     rules:
       - alert: SearchApiElasticsearchClusterStatus
+        expr: search_api_elasticsearch_status == 3
+        for: 30m
+        labels:
+          severity: warning
+          destination: slack-search-team
+        annotations:
+          summary: Elasticsearch cluster status is UNKNOWN or UNAVAILABLE
+          description: >-
+            The state of the Elasticsearch cluster used by Search API is unknown.
+      - alert: SearchApiElasticsearchClusterStatus
         expr: search_api_elasticsearch_status == 2
         for: 5m
         labels:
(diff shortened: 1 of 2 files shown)


charts/monitoring-config/rules/search_api_alerts.yaml
```diff
@@ -1,6 +1,16 @@
 groups:
   - name: SearchApiElasticsearchAlerts
     rules:
+      - alert: SearchApiElasticsearchClusterStatus
+        expr: search_api_elasticsearch_status == 3
+        for: 5m
```
emmalowe: What's the rationale behind making this 5m? Just wondering because the yellow warning is 30m.

charts/monitoring-config/rules/search_api_alerts_tests.yaml
```diff
@@ -49,6 +49,21 @@ tests:
                 The Elasticsearch cluster used by Search API is in YELLOW state.
                 Replica shards are not allocated. The cluster is still functional
                 but resilience is reduced.
+  - interval: 1m
+    input_series:
+      - series: "search_api_elasticsearch_status"
+        values: "3x40"
+    alert_rule_test:
+      - alertname: SearchApiElasticsearchClusterStatus
+        eval_time: 30m
```
emmalowe: This should probably match the time the "for" time in the alert for a tighter check. I'd also move it higher up in this test section so that the tests are in the same order as the alerts in the main file, so they're easier to find.

charts/monitoring-config/rules/search_api_alerts.yaml
hannako: The rationale wasssssss..... cut and paste errors 😂 FIXED!

emmalowe (approved):

PR #4354, opened 2026-07-16
Define GOV.UK Notify API key external secret for Fact Check Manager

young-rob:
### Changes

Define an external secret for GOV.UK Notify API key for Fact Check Manager.


PeterHattyar (dismissed):
Ignore this pls, reviewed in error.

samsimpson1 (approved):

PR #4364, opened 2026-07-20
[PP-7782] Set ClamAV `maxFileSize` as an environment variable for Asset Manager

brucebolt:
We would like the Asset Manager application to be able to know ClamAV's file size limit, so it can handle requests containing large assets better.

Therefore moving the file size limit into a variable, which is then made available through an environment variable on the app.

### DIFF
--- a/charts/asset-manager/values.yaml
+++ b/charts/asset-manager/values.yaml
@@ -146,3 +146,6 @@ nfs:
   assetStorage: 15Gi
 
 appEnabled: true
+
+clamav:
+  maxFileSizeMB: 750
--- a/charts/app-config/values-integration.yaml
+++ b/charts/app-config/values-integration.yaml
@@ -238,6 +238,8 @@ govukApplications:
             secretKeyRef:
               name: authenticating-proxy-jwt-auth-secret
               key: JWT_AUTH_SECRET
+        - name: MAX_FILE_SIZE_MB
+          value: "{{ .Values.clamav.maxFileSizeMB }}"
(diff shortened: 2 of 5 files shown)


charts/asset-manager/values.yaml
```diff
@@ -146,3 +146,6 @@ nfs:
   assetStorage: 15Gi
 
 appEnabled: true
+
+clamav:
+  maxFileSize: 750
```
yndajas: **suggestion:** would it be worth moving the unit over here too so you don't need to go looking for `charts/asset-manager/templates/clamav-configmap.yaml` to work it out? Other values here seem to include it

yndajas (dismissed):

charts/asset-manager/values.yaml
brucebolt: I decided not to include the unit, as I'd then need to parse it in the Rails app to determine which unit is being used.

yndajas: Is it worth including the unit in the variable/value name in that case?

brucebolt: Maybe I could rename it to `maxFileSizeMB`?

yndajas: Yeah I reckon that would be helpful

brucebolt: Done.

yndajas: Potentially the environment variable too?

jaskaransarkaria (approved):

PR #4367, opened 2026-07-23
Asset Manager Improvements: Integration/Staging

dj-maisy:
## What?
This applies some changes for asset manager in Integration and Staging only for now, with the aim of:

* Reducing our Bandwidth usage by using mild gzip compression (assets-origin is one of our main offenders for this)
* Setting some generous (but not too generous) keepalive limits to reduce constant cycling of TLS handshakes.
* Setting the TCP nopush and TCP nodelay options to stop Fastly having to wait longer than necessary.
* Let resolver ignore IPv6 and cache kube hosts for up to 5 seconds to reduce the number of lookups (help avoid DNS storms).
* Turn off proxy buffering for anything coming from the cloud storage proxy location - this is wasting cycles writing stuff into /tmp when it doesn't need to.

...ultimately reducing the number of peak LCUs we need to pay for ($$$). And hopefully cutting our Bandwidth needs ($$$$$).

This is currently only being rolled out to Integration and Staging so we can give it all a test before rolling anything into Prod.

## Related

* https://github.com/alphagov/govuk-fastly-secrets/pull/112
* https://github.com/alphagov/govuk-fastly/pull/428

### DIFF
--- a/charts/asset-manager/templates/nginx-configmap.yaml
+++ b/charts/asset-manager/templates/nginx-configmap.yaml
@@ -30,9 +30,40 @@ data:
       server_tokens off;
 
       sendfile        on;
+
+      {{- if or (eq .Values.govukEnvironment "integration") (eq .Values.govukEnvironment "staging") }}
+      tcp_nopush      on;
+      tcp_nodelay     on;
+
+      keepalive_timeout  75;
+      keepalive_requests 10000;
+
+      # Gzip Compression
+      gzip on;
+      gzip_comp_level 5;
+      gzip_min_length  256;
+      gzip_proxied any;
+      gzip_vary on;
+      gzip_types
+        text/plain
+        text/css
+        text/xml
+        text/javascript
+        application/javascript
+        application/x-javascript
+        application/json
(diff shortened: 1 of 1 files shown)


charts/asset-manager/templates/nginx-configmap.yaml
```diff
@@ -30,9 +30,40 @@ data:
       server_tokens off;
 
       sendfile        on;
+
+      {{- if or (eq .Values.govukEnvironment "integration") (eq .Values.govukEnvironment "staging") }}
```
AP-Hunt, line 34: We'll need to keep an eye on CPU usage across the cluster once this goes out, because we're trading bandwidth for CPU time

AP-Hunt (approved):

charts/asset-manager/templates/nginx-configmap.yaml
dj-maisy: Agreed - it's why I kept it at a slightly conservative value of 5. Most configs seem to settle at 6.

PR #4369, opened 2026-07-27
Add Asset Manager bearer token

callumknights:
Add environment variables in all environments for the new Asset Manager bearer token for HMRC Manuals API


brucebolt (approved):

nimalank7 (approved):
LGTM!

PR #4374, opened 2026-07-30
[SCH-2037] Restore search ES cluster from environment above before taking snapshot

stringiest:
Amend scheduling of cronjobs for `search-index-env-sync` so that restoring from a snapshot of the environment above happens before a snapshot of the current environment is taken. 

Also move to weekly syncing of search integration cluster, to match the way that data is synced between environments for the rest of GOV.UK.

See [Jira ticket SCH-2037](https://gov-uk.atlassian.net/browse/SCH-2037)


koetsier (approved):

nimalank7 (approved):
LGTM!

PR #4383, opened 2026-08-03
[SCH-2052] Retry snapshot creation if gateway times out

stringiest:
Snapshot creation has been failing sporadically with 504 Gateway Time-out. I've added up to three retries to see if that resolves the problem.

See [Argo logs](https://argo.eks.production.govuk.digital/applications/cluster-services/search-index-env-sync?view=tree&resource=&orphaned=false&node=batch%2FJob%2Fapps%2Fsearch-index-env-sync-green-es6-snapshot-29760591%2F0&tab=logs) or screenshot from them showing error:
20260803 gateway timeout on snapshot creation

Relates to [Jira ticket SCH-2052](https://github.com/alphagov/govuk-helm-charts/pull/4383)

### DIFF
--- a/charts/search-index-env-sync/snapshot.sh
+++ b/charts/search-index-env-sync/snapshot.sh
@@ -39,6 +39,7 @@ create () {
   local deadline
   local current_time
   local state
+  local snapshot_retry_count
 
   start_time=$(date +%s)
   readonly start_time
@@ -47,6 +48,15 @@ create () {
   
   set +e
   result=$(mycurl -XPUT "$ES_URL/_snapshot/$SNAPSHOT_REPO/$SNAPSHOT_NAME?wait_for_completion=true")
+
+  snapshot_retry_count=1
+  while [[ "$result" == *"504 Gateway Time-out"* ]] && [[ "$snapshot_retry_count" -le 3 ]]; do
+    echo "Gateway timed-out, retrying snapshot creation"
+    sleep 5
+    result=$(mycurl -XPUT "$ES_URL/_snapshot/$SNAPSHOT_REPO/$SNAPSHOT_NAME?wait_for_completion=true")
+    ((snapshot_retry_count++))
+  done
+
   set -e
   state=$(jq -r .snapshot.state <<<"$result")
   echo "$result" | jq


charts/search-index-env-sync/snapshot.sh
```diff
@@ -48,6 +49,14 @@ create () {
   set +e
   result=$(mycurl -XPUT "$ES_URL/_snapshot/$SNAPSHOT_REPO/$SNAPSHOT_NAME?wait_for_completion=true")
   set -e
+  snapshot_retry_count=1
+  while [[ "$result" == *"504 Gateway Time-out"* ]] && [[ "$snapshot_retry_count" -le 3 ]]; do
+    echo "Gateway timed-out, retrying snapshot creation"
+    sleep 5
+    result=$(mycurl -XPUT "$ES_URL/_snapshot/$SNAPSHOT_REPO/$SNAPSHOT_NAME?wait_for_completion=true")
+    snapshot_retry_count++
```
emmalowe: I'm not sure this works. (I'm getting an error when I try it out in my terminal.)

I think you need to wrap it in brackets:
`((snapshot_retry_count++))`

Or try direct assignment:
`snapshot_retry_count=$((snapshot_retry_count + 1))`

charts/search-index-env-sync/snapshot.sh
```diff
@@ -48,6 +49,14 @@ create () {
   set +e
   result=$(mycurl -XPUT "$ES_URL/_snapshot/$SNAPSHOT_REPO/$SNAPSHOT_NAME?wait_for_completion=true")
   set -e
```
emmalowe: I think you want to `set -e` after the retry. That way (as far as I understand it!) we wouldn't exit immediately if we get an error in the while loop, but would keep retrying and then save and output the result if it failed 3 times.

charts/search-index-env-sync/snapshot.sh
```diff
@@ -48,6 +49,14 @@ create () {
...
+  snapshot_retry_count=1
+  while [[ "$result" == *"504 Gateway Time-out"* ]] && [[ "$snapshot_retry_count" -le 3 ]]; do
```
emmalowe, line 53: Would it be better for the condition to be based on an http status?

charts/search-index-env-sync/snapshot.sh
stringiest: Thanks! Now changed

charts/search-index-env-sync/snapshot.sh
stringiest: I don't think we get the http status in the result, just the html of the 504 page.

charts/search-index-env-sync/snapshot.sh
stringiest: Thanks! Now changed.

emmalowe (approved):
Looks good to me!

dj-maisy (approved):

PR #4386, opened 2026-08-04
Make a flag to easily toggle it in Production

dj-maisy:
## What?
The Nginx config seems to be valid and work - there are no errors (that we can see) in Integration or Staging. I want to now make this a "feature flag" style switch so we can easily turn it on in Production.


AP-Hunt (approved):
+1 feature flagging

PR #4387, opened 2026-08-04
Enable Assets Nginx ALB Tweaks in Production.

dj-maisy:
## What?
This toggles the new "ALB Optimisation" feature flag to `true` in Production now that we're confident that the changes are valid. With this flipped to "true", we can keen an eye out for any errors and see if these changes have a positive trend on our LCU and Processed Bytes.


AP-Hunt (approved):
Keep an eye on CPU usage of the nodes as well

PR #4393, opened 2026-08-06
Move restore from prod back by 5 minutes

stringiest:
Currently this is coinciding with a regular uptick in traffic that happens between 15 & 35 minutes past the hour each hour, meaning that we've seen an increase in the number of errors since the job was rescheduled on 30th July from 3am to 2:30am:
Screenshot 2026-08-06 at 13 40 31

It also coincides with the cluster health dropping to `yellow` status:
Screenshot 2026-08-06 at 13 59 56

Moving the job back by 5 minutes to 2:35am so that it starts after the rush. The next uptick in traffic on staging happens at 2:45am, so it'd be good for this to finish before that:

Screenshot 2026-08-06 at 13 34 56

Also making a similar change to the timing of the restore backup from staging job running on integration, as there's a similar spike in traffic between 15 & 35 minutes past the hour each hour:

Screenshot 2026-08-06 at 13 57 11

This also coincides with the integration ES cluster having `yellow` status, though not every day on integration. E.g. 3rd August 2026:
Screenshot 2026-08-06 at 14 04 09

Moving this job to 4:35am to avoid the current spike.

### DIFF
--- a/charts/search-index-env-sync/values.yaml
+++ b/charts/search-index-env-sync/values.yaml
@@ -39,7 +39,7 @@ cronjobs:
 
   staging:
     green-es6-cp-prod:  # This would be too long as green-elasticsearch6-domain-cp-prod
-      schedule: "30 2 * * *"
+      schedule: "35 2 * * *"
       args: [restore_latest]
       extraEnv:
         - name: SNAPSHOT_REPO
--- a/charts/search-index-env-sync/values.yaml
+++ b/charts/search-index-env-sync/values.yaml
@@ -73,7 +73,7 @@ cronjobs:
 
   integration:
     green-es6-cp-stag:  # This would be too long as green-elasticsearch6-domain-cp-stag
-      schedule: "17 4 * * 1"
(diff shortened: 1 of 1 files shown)


emmalowe (dismissed):

charts/search-index-env-sync/values.yaml
```diff
@@ -73,7 +73,7 @@ cronjobs:
 
   integration:
     green-es6-cp-stag:  # This would be too long as green-elasticsearch6-domain-cp-stag
-      schedule: "17 4 * * 1"
```
emmalowe, line 76: Why are you changing this?

stringiest: Just about to update the description, but suddenly thought to check integration too, to see if there was a similar spike in traffic while the restore from staging job is running - and there is! Both of these jobs coincide with the elasticsearch cluster health dropping to yellow. I'm hoping that might fix it 🤞🏻 . Will update the PR description now :)

samsimpson1 (approved):

PR #4395, opened 2026-08-06
Lint app-config chart alongwith all other helm charts

AP-Hunt:
## What
We had an issue where the `app-config` chart was failing because of some bad templating code, but it wasn't detected because that chart wasn't being linted. We now lint it.

Brings the helm linting code into Make to make it easier to run the linter locally, and simplifies the script while the opportunity presented itself.

## How to review

I haven't rebase this branch atop `main` yet on purpose! The place in `main` it branches off from doesn't have the fix for the templating error, which allows us to see that this change will error in the right ways.

Once a reviewer has seen that to be the case, I will rebase the branch and they can test it again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,31 +29,21 @@ jobs:
     runs-on: ubuntu-latest
     needs: render-charts
     steps:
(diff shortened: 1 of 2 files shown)


jfharden (approved):
Looks good ❤️

PR #4400, opened 2026-08-10
[PNP-10105] Remove 410 intercept in prod/staging

KludgeKML:
This has previously been removed in integration, but we couldn't remove it in production until router and email-alert-frontend were returning 460s. Now we clear the way for custom 410s. If frontend returns a 410, it will be passed through unchanged (allowing custom HTML), and router/email-alert-frontend will return bare 460s (which are already being intercepted in all environments and altered to a 410 and decorated with pre-rendered HTML from S3)

Once this is deployed, frontend PR https://github.com/alphagov/frontend/pull/5697 can be merged (returning 410 for custom gone routes rather than 200).

https://gov-uk.atlassian.net/browse/PNP-10105

### DIFF
--- a/charts/app-config/templates/router-nginx-config.tpl
+++ b/charts/app-config/templates/router-nginx-config.tpl
@@ -249,14 +249,8 @@ http {
     # Error pages
     charset utf-8;
 
-    {{- if eq $.Values.govukEnvironment "integration" }}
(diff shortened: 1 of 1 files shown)


nimalank7 (approved):
LGTM!

PR #4403, opened 2026-08-13
Remove unused OS_MAPS_API_KEY secret handling

KludgeKML:
- This was used for the PfC map, so no longer required.

(Once this is deployed we'll also have to remove the secrets from AWS secrets manager in all three environments)


AgaDufrat (approved):
🧹

AP-Hunt (approved):

PR #4408, opened 2026-08-14
Create govuk-operators ArgoCD project

AP-Hunt:
## What

The operator, and any future operators, need to be able to deploy a more privileged set of resource types than the applications in the govuk project. For that, we create a new ArgoCD project with a greater set of permissions.

The permissions granted are based on the errors here 
image


jfharden (approved):
I think this looks good 👍

PR #4414, opened 2026-08-18
Deploy the job request operator as a Helm chart

AP-Hunt:
## What

Deploys the job request operator using the helm chart it builds and publishes. This is a little different from other things we've done before, because the chart is published as an OCI image to a registry, instead of being hosted on a Helm chart repository.

## How to review

n.b. check that `cluster-services` has been run in all environments before test and merging; there were some additional PRs that added some supporting pieces around the edges of this

1. Deploy the branch to integration
2. Check the operator is deployed
3. Check it works

### DIFF
--- a/charts/app-config/helm-versions/integration
+++ b/charts/app-config/helm-versions/integration
@@ -1,3 +1,4 @@
 # $repo_url $chart_name: "$chart_version"
 https://kubernetes.github.io/autoscaler cluster-autoscaler: "9.59.0"
 https://stakater.github.io/stakater-charts reloader: "2.2.16"
+oci://ghcr.io/alphagov/govuk/charts/govuk-job-request-operator govuk-job-request-operator: "v0.0.14"
--- a/charts/app-config/templates/job-request-operator.yaml
+++ b/charts/app-config/templates/job-request-operator.yaml
@@ -1,0 +1,37 @@
+{{- $helmReleaseInfo := include "app-config.helm-release"
+        ( merge (deepCopy .) (dict "repoURL" "oci://ghcr.io/alphagov/govuk/charts/govuk-job-request-operator" "chart" "govuk-job-request-operator") )
+}}
+{{- $imageTag := get ($helmReleaseInfo|fromYaml) "targetRevision" }}
+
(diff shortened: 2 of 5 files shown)


charts/app-config/helm-versions/integration
```diff
@@ -1,3 +1,4 @@
 # $repo_url $chart_name: "$chart_version"
 https://kubernetes.github.io/autoscaler cluster-autoscaler: "9.58.0"
 https://stakater.github.io/stakater-charts reloader: "2.2.14"
+oci://ghcr.io/alphagov/govuk/charts/govuk-job-request-operator govuk-job-request-operator: "v0.0.13"
```
jfharden:
```suggestion
oci://ghcr.io/alphagov/govuk/charts/govuk-job-request-operator govuk-job-request-operator: "v0.0.14"
```

(Similar for the other envs)

charts/app-config/templates/job-request-operator.yaml
```diff
@@ -0,0 +1,40 @@
+{{- $imageTagConfig := $.Files.Get (printf "image-tags/%s/%s" $.Values.govukEnvironment "job-request-operator") | fromYaml }}
```
jfharden: The image-tags files these refer to are not present in the repo (like this example for account-api) https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/image-tags/integration/account-api

I think because this is a helm-release versioned thing, rather than an imageTags one, you can just leave out these top 4 lines completely like cluster-autoscaler does https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/templates/cluster-autoscaler.yaml#L1

jfharden (changes requested):

charts/app-config/templates/job-request-operator.yaml
jfharden: You will have to adapt some of the annotations too I realise now

AP-Hunt: Ooh good catch. They're there because I originally went down the path of using image tag files, then rowed it back.

jfharden (approved):

PR #4416, opened 2026-08-18
Set `dbMigrationEnabled: true` for fact-check-manager in production

young-rob:
The fact-check-manager database is set up now so we can enable database migrations.


nimalank7 (approved):
LGTM!

PR #4440, opened 2026-08-27
URL-encode database username/password when constructing URL

ChrisBAshton:
One of the database passwords I created recently has a '?' char in it, which has a special meaning in URL construction, breaking the db connection URL. URL-encoding the username and password that is interpolated should solve the issue.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3578


samsimpson1 (changes requested):
https://gds.slack.com/archives/C09ALJB14GZ/p1787822910681089?thread_ts=1787733391.739329&cid=C09ALJB14GZ

PR #4448, opened 2026-09-03
Suspend cronjobs syncing data to Green cluster on integration

koetsier:
The Green Elasticsearch cluster on Integration is going to be removed, so these cronjobs will fail if they are not suspended.

The Blue cluster is not compatible with the Green cluster, so we cannot use snapshots to sync data between them. These cronjobs can only be reinstated after all clusters have been upgraded in staging and production.


stringiest (approved):
🎉

emmalowe (approved):

jaskaransarkaria (approved):

PR #4460, opened 2026-09-08
Run Elasticdump on production at 2:30 UTC on Wednesday 9 September

koetsier:
To copy data across from the green to blue cluster

### DIFF
--- a/charts/app-config/values-production.yaml
+++ b/charts/app-config/values-production.yaml
@@ -2905,8 +2905,8 @@ govukApplications:
       cronTasks:
         - name: elasticdump
           task: "elasticdump"
-          schedule: "0 0 31 2 *" # never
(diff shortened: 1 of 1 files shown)


charts/app-config/values-production.yaml
```diff
@@ -2905,8 +2905,8 @@ govukApplications:
       cronTasks:
         - name: elasticdump
           task: "elasticdump"
-          schedule: "0 0 31 2 *" # never
-          suspend: true # Only run this task manually
+          schedule: "30 02 9 9 3" # Wednesday 9 September 11:15 UTC
```
samsimpson1: This states 02:30 AM, not 11:15

koetsier: fixed

samsimpson1 (approved):

PR #4462, opened 2026-09-09
Switch back from the Opensearch Blue cluster to Green in production 

koetsier:
If the Blue cluster fails we can switch back to Green


koetsier:
we upgraded so we can close this now

PR #4484, opened 2026-09-18
Add 'reboot' db-backup operation

samsimpson1:
This operation checks to see if the RDS instance requires a reboot to apply pending PG changes

If a reboot is required, it blocks until the reboot is complete

We're also deploying this reboot operation to all RDS instances in staging for now. It will be rolled out to all environments once its ran a few times and we understand it is working as intended.

## Testing

This has been tested in staging on:
- A PostgreSQL instance that does not require a reboot
- A PostgreSQL instance that does require a reboot
- A MySQL instance that does not require a reboot
- A MySQL instance that does require a reboot

https://github.com/alphagov/govuk-infrastructure/issues/4116


charts/db-backup/scripts/reboot
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+set -euo pipefail
+
+: "${DB_HOST:?required}"
+
+# Resolve the DBMS hostname to the AWS endpoint it is a CNAME for
+rds_endpoint () {
+  local endpoint
+  endpoint=$(dig +short +search "$1" | sed 's/\.$//' | grep -E '\.rds\.amazonaws\.com$' | tail -1) || true
```
jaskaransarkaria, line 9: if $endpoint = || true

then what's the point of:

  if [ -z "$endpoint" ]; then

charts/db-backup/scripts/reboot
```diff
@@ -0,0 +1,87 @@
...
+# Wait until the instance has left the "available" state and come back to it
+wait_for_reboot () {
+  local identifier="$1"
+  local status
+
+  # The instance stays "available" for a while after the reboot requested, 
+  # so wait for it to start rebooting before waiting for it to
+  # finish. Give up on seeing the transition after a minute, in case we
+  # missed it entirely.
+  for _ in $(seq 1 6); do
```
jaskaransarkaria, line 57: is 60 secs also enough?

charts/db-backup/scripts/reboot
jaskaransarkaria: ah ok I double checked the `|| true` only changes the exit status so the if will read $endpoint as empty still

charts/db-backup/scripts/reboot
samsimpson1: The 60s is for the instance to enter the rebooting state, not for the reboot to complete. The `aws rds wait db-instance-available` does the actual waiting

jaskaransarkaria (approved):

PR #4488, opened 2026-09-22
Add reboot step to all RDS instances in integration

samsimpson1:
This was enabled in staging last week, and it seems to be working as expected.

https://github.com/alphagov/govuk-infrastructure/issues/4116


nimalank7 (approved):
LGTM!

